PR #1, opened 2020-07-24
Initial Definition of JS API

hossenlopp:
Defines the functions and types of the inputs and outputs of running calculations.


test/index.test.ts
```diff
@@ -1,3 +1,9 @@
-test('test tests placeholder', () => {
-  expect(true).toBeTruthy();
+import * as index from '../src';
+
+test('index.ts to have Calculator', () => {
+  expect(index).toHaveProperty('Calculator');
+});
+
+test('index.ts to have Calculator.calculate', () => {
```
mgramigna, line 7: Are we also exposing `calculateMeasureReports`? Should we have a test for that too?

PR #6, opened 2020-09-02
Fix some typedef issues and comments. Move enums out of d.ts

hossenlopp:
- A few comments have added clarification.
- Added a missing item from an ELM structure definition.
- Moved Enums to their own file that is not a d.ts file. They need to be in a .ts to be compiled and used.


okeefm (approved):
Changes make sense to me, and good comment updates.

PR #7, opened 2020-09-16
Add basic CLI to call calculate functions with command line args

mgramigna:
Usage: `ts-node src/cli.ts -m path/to/measure/bundle -p path/to/patient/bundle`

Output is just the return result of the `calculate` function stringified to the console. To write it to a file, add `> output.json` or something to the end of the above command.


hossenlopp (approved):
This will do the trick nicely for our initial development.

PR #8, opened 2020-10-02
TACOSTRAT-453 Basic CQL Measure Execution Endpoint

okeefm:
This PR is for the "raw results" calculation work done by myself, Tom, Matt, and Chris. 

NOTE: To make this work locally, switch the `cql-exec-fhir` and `cql-execution` dependencies to local deps (and download the branches indicated by package.json locally). The dependencies should look like below:

```
"cql-exec-fhir": "file:../cql-exec-fhir",
"cql-execution": "file:../cql-execution"
```

TODO: add tests


mgramigna:
DoD says we should also add an endpoint in https://github.com/projecttacoma/fqm-execution-service for the calculateRaw functionality, so I think a companion PR there is warranted. It should be pretty simple, as the app is all set up. Just need to pass measure and patient into the calculate call [here](https://github.com/projecttacoma/fqm-execution-service/blob/master/src/app.ts). For now perhaps we should also rename this endpoint to calculateRaw

script.ts
```diff
@@ -0,0 +1,12 @@
+import { R4 } from '@ahryman40k/ts-fhir-types';
```
mgramigna, line 1: do we still need this now that we have the cli?

src/Execution.ts
```diff
@@ -0,0 +1,110 @@
...
+  measureBundle: R4.IBundle,
+  patientBundles: R4.IBundle[],
+  options: CalculationOptions
+): {
+  rawResults?: cql.Results;
+  errorMessage?: string;
+  elmLibraries?: ELM[];
+  mainLibraryName?: string;
+} {
+  // TODO: return type^
```
mgramigna, line 21: Obsolete TODO?

src/Execution.ts
```diff
@@ -0,0 +1,110 @@
...
+      libraries.push(library);
+      const elmsEncoded = library.content?.filter(a => a.contentType === 'application/elm+json');
+      elmsEncoded?.forEach(elmEncoded => {
+        if (elmEncoded.data) {
+          const decoded = Buffer.from(elmEncoded.data, 'base64').toString('binary');
+          const elm = JSON.parse(decoded) as ELM;
+          if (library.id === rootLibId) {
+            rootLibIdentifer = elm.library.identifier;
+          }
+          // This line is a hack to
```
mgramigna, line 51: ...

src/Execution.ts
okeefm: Yeah, looks that way

src/Execution.ts
okeefm: 😕

script.ts
okeefm: Probably not

src/Execution.ts
hossenlopp: I think I meant it as a question along the lines as "should we externalize the definition to an `interface`".

okeefm: Ah. I thought it was a comment we'd made when we were returning an `any` here before

mgramigna: Ah, that makes sense. i think we can move this return type to one of our type declaration files, yeah

okeefm: Yeah, I'd be fine with that change too

src/DebugHelper.ts
```diff
@@ -0,0 +1,56 @@
+import { ELM } from './types/ELMTypes';
+import { ValueSetMap } from 'cql-execution';
+import fs = require('fs');
```
mgramigna, line 3:
```suggestion
import fs from 'fs';
```

okeefm: Updated. You may have to clean away your branch and re-pull, since I had to amend the commit before this one.

mgramigna (dismissed):
ez

mgramigna (approved):

PR #9, opened 2020-10-09
Use beta versions of CQL packages, and bump version

okeefm:
Use beta versions of cql-execution and cql-exec-fhir with is, as, and overloading support, and bump the library version to 0.1.0.


hossenlopp:
Version bumps are good and function as expected.

hossenlopp (approved):

PR #10, opened 2020-10-09
Update types to properly be exposed to other TS libraries

mgramigna:
Exposing our custom types defined in this library to be used in other typescript libraries.

Important thing I learned: type declaration files that you intend to be used by other typescript projects potentially should end in `.ts` not `.d.ts`. This PR renames those files accordingly.

I also added an index.ts to the types directory to export them all, and added an export to the main index.ts that also exposes the types.

lastly, i added the declaration flag to tsconfig.json which will autogenerate typescript declaration files when compiling the typescript. This allows for proper usage of fqm-execution in other TS libraries as they will contain the proper declarations for all of our code.

This does not change any behavior, only the exposed source code.


okeefm (approved):
Looks good! Super stoked to be able to use these types in `fqm-execution-service`

PR #12, opened 2020-10-12
Measure report cp

lmd59:
# Summary
Uses the existing calculate results to construct a report according to standard https://www.hl7.org/fhir/measurereport.html
## New behavior
calculateMeasureReports creates (and dumps to debug) a full MeasureReport. This includes all simple fields and references as well as population counts and measure score. In the future, we may need to still add stratifiers, SDEs, more robust start/end date handling, other kind of measure calculation, and more evaluated resource details
## Code changes
Only in calculateMeasureReports(...) and supporting function
# Testing guidance
Use cli with arguments: "args": ["-o", "reports","-m", "${workspaceFolder}/test/fixtures/EXM130-7.3.000-bundle.json", "-p", "${workspaceFolder}/test/fixtures/EXM130-8.0.000/EXM130-8.0.000-patients/numerator/Alejandro916_Becker968_6a50325f-89c6-4d09-88e7-1eede4a49df5.json"]


src/Calculator.ts
```diff
@@ -116,20 +118,109 @@ export function calculate(
...
-      resourceType: 'MeasureReport',
-      measure: 'Measure/ImplementMe',
-      period: {}
-    }
-  ];
+  // options should be updated by this call if measurementPeriod wasn't initially passed in
+  const results: ExecutionResult[] = calculate(measureBundle, patientBundles, options);
+  const reports: R4.IMeasureReport[] = [];
+  results.forEach(function (result) {
+    const report: R4.IMeasureReport = <any>{};
```
mgramigna, line 132:
```suggestion
    const report = <R4.IMeasureReport>{};
```

I'm assuming the reasoning for the `any` cast here is that this interface requires specific information and it won't let you initialize it to empty object. We then later fill out the proper information in the report in subsequent lines. The "proper" way to do this in typescript is to omit a type from the left side and use a type assertion for the desired type on the right side. This essentially tells the compiler: "dang trust me, this is a measure report"

There are a couple more instances of this that I'll add suggestions for as well

src/Calculator.ts
```diff
@@ -116,20 +118,109 @@ export function calculate(
...
+    report.period = {
+      start: options.measurementPeriodStart, // double check format of start and end that we're passing in https://www.hl7.org/fhir/datatypes.html#dateTime... we don't seem to be passing anything in from CLI
+      end: options.measurementPeriodEnd
+    };
+    report.status = R4.MeasureReportStatusKind._complete; // are there cases where this should be _pending or _error?
+    report.type = R4.MeasureReportTypeKind._individual; // are there cases where this should be _subjectList or _summary or _dataCollection
+
+    // measure url from measure bundle
+    const measureEntry = measureBundle.entry?.find(e => e.resource?.resourceType === 'Measure') as R4.IBundle_Entry;
+    const measure = measureEntry.resource as R4.IMeasure;
```
mgramigna, line 145:
```suggestion
    const measureEntry = measureBundle.entry?.find(e => e.resource?.resourceType === 'Measure');
    const measure = measureEntry?.resource as R4.IMeasure;
```

The type of the `find` will be implied as a bundle entry. It also might be undefined, so the "proper" way around this is to add the questionMark on the subsequent line for the measureEntry

src/Calculator.ts
```diff
@@ -116,20 +118,109 @@ export function calculate(
...
+    report.measure = measure.url || 'UnknownMeasure'; // or some other default?
+
+    // create group population counts from result's detailedResults (yes/no->1/0)
+    report.group = [];
+    result?.detailedResults?.forEach(function (dr) {
+      // TODO: check the measure definition for stratification to determine whether to add group.stratiier
+      // if yes, add stratifier with population copied into. Set counts to 0 if the result for the stratifier is false
+      const group: R4.IMeasureReport_Group = <any>{};
```
mgramigna, line 153:
```suggestion
      const group = <R4.IMeasureReport_Group>{};
```

src/Calculator.ts
```diff
@@ -116,20 +118,109 @@ export function calculate(
...
+      const detail: DetailedPopulationGroupResult = dr;
```
mgramigna, line 154:
```suggestion
      const detail = dr;
```

Type of this variable is implied

src/Calculator.ts
```diff
@@ -116,20 +118,109 @@ export function calculate(
...
+      group.id = detail.groupId;
+      group.population = [];
+      let numeratorCount = 0.0;
+      let denominatorCount = 0.0;
+      // TODO: handle EXM111 (doesn't identify itself as a episode of care measure). if it's an episode of care, you need to iterate over
+      // stratifications : may need to clone results for one population group and adjust (in this case, just a straight clone)
+      detail?.populationResults?.forEach(function (pr) {
```
mgramigna, line 161:
```suggestion
      detail?.populationResults?.forEach((pr) => {
```

src/Calculator.ts
```diff
@@ -116,20 +118,109 @@ export function calculate(
...
-    {
-      resourceType: 'MeasureReport',
-      measure: 'Measure/ImplementMe',
-      period: {}
-    }
-  ];
+  // options should be updated by this call if measurementPeriod wasn't initially passed in
+  const results: ExecutionResult[] = calculate(measureBundle, patientBundles, options);
+  const reports: R4.IMeasureReport[] = [];
+  results.forEach(function (result) {
```
mgramigna, line 131:
```suggestion
  results.forEach((result) => {
```

src/Calculator.ts
```diff
@@ -116,20 +118,109 @@ export function calculate(
...
+    report.type = R4.MeasureReportTypeKind._individual; // are there cases where this should be _subjectList or _summary or _dataCollection
+
+    // measure url from measure bundle
+    const measureEntry = measureBundle.entry?.find(e => e.resource?.resourceType === 'Measure') as R4.IBundle_Entry;
+    const measure = measureEntry.resource as R4.IMeasure;
+    report.measure = measure.url || 'UnknownMeasure'; // or some other default?
+
+    // create group population counts from result's detailedResults (yes/no->1/0)
+    report.group = [];
+    result?.detailedResults?.forEach(function (dr) {
```
mgramigna, line 150:
```suggestion
    result?.detailedResults?.forEach((dr) => {
```

yay arrow functions :)

src/Calculator.ts
```diff
@@ -116,20 +118,109 @@ export function calculate(
...
+      const group: R4.IMeasureReport_Group = <any>{};
+      const detail: DetailedPopulationGroupResult = dr;
+      group.id = detail.groupId;
+      group.population = [];
+      let numeratorCount = 0.0;
+      let denominatorCount = 0.0;
+      // TODO: handle EXM111 (doesn't identify itself as a episode of care measure). if it's an episode of care, you need to iterate over
+      // stratifications : may need to clone results for one population group and adjust (in this case, just a straight clone)
+      detail?.populationResults?.forEach(function (pr) {
+        const pop: IMeasureReport_Population = <any>{};
```
mgramigna, line 162:
```suggestion
        const pop = <R4.IMeasureReport_Population>{};
```

src/Calculator.ts
```diff
@@ -116,20 +118,109 @@ export function calculate(
...
+              system: 'http://terminology.hl7.org/CodeSystem/measure-population',
+              code: pr.populationType,
+              display: populationDisplay(pr.populationType)
+            }
+          ]
+        };
+        pop.count = pr.result ? 1 : 0;
+        if (pr.populationType == PopulationType.NUMER) numeratorCount += pop.count;
+        if (pr.populationType == PopulationType.DENOM) denominatorCount += pop.count;
+        group.population?.push?.(pop);
```
mgramigna, line 176:
```suggestion
        group.population?.push(pop);
```

src/Calculator.ts
```diff
@@ -116,20 +118,109 @@ export function calculate(
...
+    const evalId = uuidv4();
+    const contained: R4.IBundle = {
+      resourceType: 'Bundle',
+      id: evalId,
+      type: R4.BundleTypeKind._collection,
+      entry: []
+    };
+    const patient = patientBundles[0].entry?.filter(function (e) {
+      return e.resource?.resourceType == 'Patient';
+    })[0].resource;
```
mgramigna, line 197:
```suggestion
    const patient = patientBundles[0].entry?.find((e) => {
      return e.resource?.resourceType === 'Patient';
    }).resource;
```

If you use `find` here instead, you shouldn't need to index using `[0]` since `find` will always grab the first occurrence of something. The condition in the function can stay the same.

If we think we might need other resources, you can ignore this suggestion

mgramigna (changes requested):
Looks good! many small comments

1) Inline comments made

2) Was the bundle for EXM130 7.3.000 inadvertently committed? We want to avoid committing raw bundles since they have valueset resources

lmd59:
> Looks good! many small comments
> 
> 1. Inline comments made
> 2. Was the bundle for EXM130 7.3.000 inadvertently committed? We want to avoid committing raw bundles since they have valueset resources

@2, Ach- I had that in my testing recommendations, so I included it. I thought that was just new because we were updating the measure version, but I see that it's in a different format now... I'll get rid of it. :/

src/Calculator.ts
```diff
@@ -144,3 +235,10 @@ export function calculateRaw(
     return results.errorMessage ?? 'something happened with no error message';
   }
 }
+
+function populationDisplay(type: PopulationType): string {
+  // Code->Display https://terminology.hl7.org/1.0.0/CodeSystem-measure-population.html
+  // Always same pattern, example: 'initial-population' -> 'Initial Population'
+  const words: string[] = type.split('-');
+  return words.map(word => word.charAt(0).toUpperCase() + word.slice(1)).join(' ');
+}
```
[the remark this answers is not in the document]
lmd59: This does always work. It would be more flexible than mapping if any population types were added... but since I don't think that's ever going to happen, I think I like yours better.

src/Calculator.ts
mgramigna: bump

src/Calculator.ts
mgramigna: bump

src/Calculator.ts
mgramigna: bump

mgramigna (changes requested):

mgramigna (approved):

PR #13, opened 2020-10-19
Results Processing

hossenlopp:
# Summary
Processes raw results from `cql-execution` into the the results structures defined in `types/Calculator.ts`. This provides a load of information that can be used for creating a MeasureReport or Highlighting. Most of this code was ported over from `cqm-execution` and converted to work with FHIR objects.

## New Behavior
`calculate` now returns detailed ExecutionResults instead of a placeholder sample.

# Testing Guidance
Similar to the MeasureReport PR #12. User cli with arguments: `"args": ["-o", "reports","-m", "${workspaceFolder}/test/fixtures/EXM130-7.3.000-bundle.json", "-p", "${workspaceFolder}/test/fixtures/EXM130-8.0.000/EXM130-8.0.000-patients/numerator/Alejandro916_Becker968_6a50325f-89c6-4d09-88e7-1eede4a49df5.json"]`


src/Calculator.ts
```diff
@@ -72,20 +118,109 @@ export function calculate(
...
+    const evalId = uuidv4();
+    const contained: R4.IBundle = {
+      resourceType: 'Bundle',
+      id: evalId,
+      type: R4.BundleTypeKind._collection,
+      entry: []
+    };
+    const patient = patientBundles[0].entry?.find(e => {
+      return e.resource?.resourceType === 'Patient';
+    })?.resource;
```
mgramigna, line 197: Whoops, I think I missed this when reviewing @lmd59's PR, so I know this isn't actually a snippet you wrote. But shouldn't we not be accessing the first entry of patientBundles directly? We should have the ID of the patient in the detailed results and then do a find in the bundle based on that?

src/ResultsHelpers.ts
```diff
@@ -0,0 +1,839 @@
...
+  } else if (result instanceof Object) {
+    // if the object has it's own custom toString method, use that instead
+    if (typeof result.toString === 'function' && result.toString !== Object.prototype.toString) {
+      return result.toString();
+    }
+    prettyResultReturn = '{\n';
+    const baseIndentation = Array(3).join(' ');
+    const sortedKeys = Object.keys(result)
+      .sort()
+      .filter(key => key !== '_type' && key !== 'qdmVersion');
```
mgramigna, line 462: Will qdmVersion ever be a thing?

src/ResultsHelpers.ts
```diff
@@ -0,0 +1,839 @@
...
+  // Create initial results starting with all true to create the basis for relevance.
+  const relevantResults: PopulationResult[] = results.map(result => {
+    return {
+      populationType: result.populationType,
+      result: true
+    };
+  });
+
+  /** TODO: Determine how Straifiers play here
+  // If STRAT is 0 then everything else is not calculated
```
mgramigna, line 641: Again asking about the TODO: is this something we need to address?

mgramigna (changes requested):

src/Calculator.ts
okeefm: We probably want to actually add all of the patients that get passed in to the contained resources list, right? So in theory we should just loop over `patientBundles` here and add them all (or loop over the list of patients coming back from the detailed results. Either way in theory this method can handle >1 patient, so we should handle that case here).

src/CalculatorHelpers.ts
```diff
@@ -0,0 +1,533 @@
...
+ * Create the code service valueset database that the cql-execution engine needs.
+ *
+ * NOTE: This uses the `compose` attribue of the ValueSet to get code. This is incorrect and
+ * should be using the `expansion`. But current example measures have ValueSets with compose
+ * only.
+ *
+ * @param valueSetResources FHIR ValueSets.
+ * @returns The value set DB structure needed for the cql-execution CodeService.
+ */
+export function valueSetsForCodeService(valueSetResources: R4.IValueSet[]): any {
```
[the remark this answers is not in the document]
okeefm: Gonna guess this is my fault from grabbing stuff from Hoss's branch. Probably worth deleting one version or the other though.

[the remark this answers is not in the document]
mgramigna: Yeah, I think this one should be deleted since the other one is more type-specific (returns the value set map instead of any), and lives in a standalone file, etc.

src/ResultsHelpers.ts
```diff
@@ -0,0 +1,839 @@
...
+ * @param {(Array|object|boolean|???)} result - The result from the calculation engine.
+ * @param {number|undefined} indentLevel - For nested objects, the indentLevel indicates how far to indent.
+ *                                Note that 1 is the base because Array(1).join ' ' returns ''.
+ * @param {number|undefined} keyIndent - Indent count used for key indentation.
+ * @returns {String} a pretty version of the given result
+ */
+export function prettyResult(result: any | null, indentLevel?: number, keyIndent?: number): string {
+  // TODO: Sort out a better way to have a friendly system display for FHIR codes.
+  //   This will need to be replaced with a canonical URL to code system name map.
+  const nameOidHash = {
```
okeefm, line 349: According to my editor, this variable isn't used anywhere. Is this here for future-proofing, or should we remove it?

hossenlopp: The TODO there explains the deal. We need to replace it with something more FHIR friendly. I think this would be good further work if we want to make use of the the pretty results from statements. Is it okay to leave this for now with the TODO?

src/Calculator.ts
```diff
@@ -19,51 +24,92 @@ export function calculate(
...
-            },
-            {
-              populationType: PopulationType.DENOM,
-              result: false
-            }
-          ]
-        }
-      ]
+  const measureEntry = measureBundle.entry?.find(e => e.resource?.resourceType === 'Measure') as R4.IBundle_Entry;
+  const measure = measureEntry.resource as R4.IMeasure;
```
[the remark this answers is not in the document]
hossenlopp: Addressed in commit. https://github.com/projecttacoma/fqm-execution/pull/13/commits/fae1d0bbc58ad4ab7a2b4d8d548d2384cfe2d5c6

src/Calculator.ts
```diff
@@ -19,51 +24,92 @@ export function calculate(
...
+  if (!results.elmLibraries || !results.mainLibraryName) {
+    throw new Error('no libraries were found');
+  }
+  const elmLibraries = results.elmLibraries;
+  const mainLibraryName = results.mainLibraryName;
+
+  // Iterate over patient bundles and make results for each of them.
+  patientBundles.forEach(patientBundle => {
+    const patientEntry = patientBundle.entry?.find(e => e.resource?.resourceType === 'Patient') as R4.IBundle_Entry;
+    const patient = patientEntry.resource as R4.IPatient;
```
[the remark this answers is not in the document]
hossenlopp: Addressed in https://github.com/projecttacoma/fqm-execution/pull/13/commits/5680d4e3defecafafd0d8d12ba4b3824768c580f

src/Calculator.ts
hossenlopp: This definitely needs to be replaced. We are inside a iteration over the list of `ExecutionResults`. Each of these results indicates the patientId so we should use that to find the bundle for this patient.  As of right now it would use the data from the first patient to build the MeasureReport, so this would produce bad reports for any patient except the first one. Should I address this in my PR now or should this be addressed in later work?

src/CalculatorHelpers.ts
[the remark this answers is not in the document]
hossenlopp: I'll delete this. It's my bad, I told @okeefm he needed the function so I would remove my version of it.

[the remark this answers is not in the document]
hossenlopp: Deleted in bd92736

src/ResultsHelpers.ts
hossenlopp: Nope. This code is to print info about a result that contains some element of data originally from QDM. We wanted to filter out the internal/not relevant pieces. We may have to do similar work with the FHIR data models we use. This function isn't important for the initial release of Measure Highlighting. Should I add a TODO to address this later for FHIR types?

src/ResultsHelpers.ts
hossenlopp: Stratifiers are handled at a different level now. This has been removed in 2284c38.

src/ResultsHelpers.ts
mgramigna: Yeah that makes sense to me. TODO for FHIR types is good

src/Calculator.ts
mgramigna: I think this should be addressed here. Otherwise the measure report logic will only work for one patient, right?

src/ResultsHelpers.ts
okeefm: Yeah, that makes sense. As long as there's a reason for it to be here, it's g2g.

src/Calculator.ts
hossenlopp: Addressed in c09ce66. Can now run cli with list of paths to patient bundles. Existing calls with just one work as well.

src/ResultsHelpers.ts
hossenlopp: TODO added in 3878c44.

src/cli.ts
```diff
@@ -9,7 +9,7 @@ import { calculate, calculateMeasureReports, calculateRaw } from './Calculator';
 program
   .option('-o, --output-type <type>', 'type of output, "raw", "detailed", "reports"', 'detailed')
   .requiredOption('-m, --measure-bundle <measure-bundle>', 'path to measure bundle')
-  .requiredOption('-p, --patient-bundle <patient-bundle>', 'path to  patient bundle')
+  .requiredOption('-p, --patient-bundles <patient-bundles...>', 'paths to patient bundle')
```
okeefm, line 12: This may be a silly question, but what does this look like in calling the CLI? Is it -p path/to/patient1.json,path/to/patient2.json? -p path/to/patient1.json -p path/to/patient2.json? Something else?

hossenlopp: `-p path/to/patient1.json path/to/patient2.json`. So it's backwards compatible with how we have been calling it for the one.

hossenlopp: It makes use of `commander`'s variadic options. https://www.npmjs.com/package/commander#variadic-option

mgramigna (approved):
Looks great to me :)

src/cli.ts
okeefm: Ah, cool. Once we have anything at all in the README, we can document it there.

okeefm (approved):
lgtm ✅

PR #14, opened 2020-10-20
HTML Generation from ELM Annotations

mgramigna:
**DRAFT**: Review/merge #13 first, then the base branch of this will be updated to `master`

# Summary

Uses the statementResults processed in #13 to assemble the HTML tree structure that mirrors the ELM annotations for a given statement. Uses the [handlebars](https://handlebarsjs.com/) library to make a recursive template for the annotation. The resulting HTML is put onto the text narrative of the `MeasureReport` resource(s) generated when the `calculateMeasureReports` function is called.

The HTML string is also added to the detailed results for each group and can be calculated using the `calculateHTML` calculation option.

## HTML Example

An annotation on an ELM statement can have 3 relevant properties for this: `r` (reference id), `s` (sub-clause(?), can be recursive), and `value` (array of strings that compose the CQL statement). The algorithm goes roughly as follows:

1) If the object has `value`, then it is a leaf node, and concatenate the contents of `value` into a `<span>`
2) Else, the object should have `r` and `s`. Create a `<span>` with `data-ref-id` equal to the value of `r`
3) Iterate through every child node in `s` and recurse back to 1)

Take the following CQL snippet for example:

``` cql
define "SDE Ethnicity":
  SDE."SDE Ethnicity"
```

The following is the ELM annotation for this statement;

``` JSON
{
  "type": "Annotation",
  "s": {
    "r": "26",
    "s": [
      {
        "value": [
          "define ",
          "\"SDE Ethnicity\"",
          ":\r\n  "
        ]
      },
      {
        "r": "25",
        "s": [
          {
            "r": "24",
            "s": [
              {
                "value": [
                  "SDE"
                ]
              }
            ]
          },
          {
            "value": [
              "."
            ]
          },
          {
            "r": "25",
            "s": [
              {
                "value": [
                  "\"SDE Ethnicity\""
                ]
              }
            ]
          }
        ]
      }
    ]
  }
}
```

Following the above algorithm, this is the desired result:

``` HTML
<span data-ref-id="26">
	<span>define "SDE Ethnicity":
  </span>
	<span data-ref-id="25">
		<span data-ref-id="24">
			<span>SDE</span>
		</span>
		<span>.</span>
		<span data-ref-id="25">
			<span>"SDE Ethnicity"</span>
		</span>
	</span>
</span>
```

## New behavior

* For detailed results, each result can now optionally have an `html` property which is the HTML string of the clause results for that group
* For measure reports, it will optionally attach the html string for the results to the text.div property
* HTML files that are generated are written to the debug folder

## Code changes

* New dependencies and scripts in package.json. Added the `copy-templates` script which moves the `.hbs` templates into the build directory since they won't be picked up by `tsc`
* `HTMLGenerator.ts`: New helper file that generates html structure for a given group/set of statement results
* `calculate` function calls html generator if the calculateHTML calculationOption is provided
* HTML is added to measure report narrative in `calculateMeasureReports` if the option is provided
* `templates` directory containing handlebars templates for the HTML structure
* Added more granular ELM types for annotations 
* Created a testHelpers file for re-usable functionality across all tests
* Added PR template cause why not

# Testing guidance

Run the cli with the output type "reports". This will generate HTML on the MeasureReport narrative as well as dump the html file to the `debug` inspect the contents of both. (I recommend opening the HTML in the actual browser and using the chrome console to look at the structure).

Run the tests with `npm test`

**NOTE**: If you use the VS Code launch debugger to run the CLI, the configuration needs to be updated, since we need the `npm run copy-templates` script to be run. Update the following in your `launch.json`:

```
            "preLaunchTask": "npm: build",
```

The build script will both run`tsc` and copy the template files necessary at runtime


mgramigna:
closing for #16

PR #16, opened 2020-10-29
Html annotation

mgramigna:
# Summary

Uses the statementResults processed in #13 to assemble the HTML tree structure that mirrors the ELM annotations for a given statement. Uses the [handlebars](https://handlebarsjs.com/) library to make a recursive template for the annotation. The resulting HTML is put onto the text narrative of the `MeasureReport` resource(s) generated when the `calculateMeasureReports` function is called.

The HTML string is also added to the detailed results for each group and can be calculated using the `calculateHTML` calculation option.

## HTML Example

An annotation on an ELM statement can have 3 relevant properties for this: `r` (reference id), `s` (sub-clause(?), can be recursive), and `value` (array of strings that compose the CQL statement). The algorithm goes roughly as follows:

1) If the object has `value`, then it is a leaf node, and concatenate the contents of `value` into a `<span>`
2) Else, the object should have `r` and `s`. Create a `<span>` with `data-ref-id` equal to the value of `r`
3) Iterate through every child node in `s` and recurse back to 1)

Take the following CQL snippet for example:

``` cql
define "SDE Ethnicity":
  SDE."SDE Ethnicity"
```

The following is the ELM annotation for this statement;

``` JSON
{
  "type": "Annotation",
  "s": {
    "r": "26",
    "s": [
      {
        "value": [
          "define ",
          "\"SDE Ethnicity\"",
          ":\r\n  "
        ]
      },
      {
        "r": "25",
        "s": [
          {
            "r": "24",
            "s": [
              {
                "value": [
                  "SDE"
                ]
              }
            ]
          },
          {
            "value": [
              "."
            ]
          },
          {
            "r": "25",
            "s": [
              {
                "value": [
                  "\"SDE Ethnicity\""
                ]
              }
            ]
          }
        ]
      }
    ]
  }
}
```

Following the above algorithm, this is the desired result:

``` HTML
<span data-ref-id="26">
	<span>define "SDE Ethnicity":
  </span>
	<span data-ref-id="25">
		<span data-ref-id="24">
			<span>SDE</span>
		</span>
		<span>.</span>
		<span data-ref-id="25">
			<span>"SDE Ethnicity"</span>
		</span>
	</span>
</span>
```

## New behavior

* For detailed results, each result can now optionally have an `html` property which is the HTML string of the clause results for that group
* For measure reports, it will optionally attach the html string for the results to the text.div property
* HTML files that are generated are written to the debug folder

## Code changes

* New dependencies and scripts in package.json. Added the `copy-templates` script which moves the `.hbs` templates into the build directory since they won't be picked up by `tsc`
* `HTMLGenerator.ts`: New helper file that generates html structure for a given group/set of statement results
* `calculate` function calls html generator if the calculateHTML calculationOption is provided
* HTML is added to measure report narrative in `calculateMeasureReports` if the option is provided
* `templates` directory containing handlebars templates for the HTML structure
* Added more granular ELM types for annotations 
* Created a testHelpers file for re-usable functionality across all tests
* Added PR template cause why not

# Testing guidance

Run the cli with the output type "reports". This will generate HTML on the MeasureReport narrative as well as dump the html file to the `debug` inspect the contents of both. (I recommend opening the HTML in the actual browser and using the chrome console to look at the structure).

Run the tests with `npm test`

**NOTE**: If you use the VS Code launch debugger to run the CLI, the configuration needs to be updated, since we need the `npm run copy-templates` script to be run. Update the following in your `launch.json`:

```
            "preLaunchTask": "npm: build",
```

The build script will both run`tsc` and copy the template files necessary at runtime

### DIFF
--- a/src/Calculator.ts
+++ b/src/Calculator.ts
@@ -162,6 +162,7 @@ export function calculateMeasureReports(
       // add narrative for relevant clauses
       if (detail.html) {
         report.text = {
+          status: R4.NarrativeStatusKind._generated,
           div: detail.html
         };
       }
--- a/src/Calculator.ts
+++ b/src/Calculator.ts
@@ -2,14 +2,12 @@ import { R4 } from '@ahryman40k/ts-fhir-types';
 import { ExecutionResult, CalculationOptions } from './types/Calculator';
 import { PopulationType } from './types/Enums';
 import { v4 as uuidv4 } from 'uuid';
-
-// import { PatientSource } from 'cql-exec-fhir';
 import * as cql from './types/CQLTypes';
 import * as Execution from './Execution';
-import { dumpObject } from './DebugHelper';
-
+import { dumpHTML, dumpObject } from './DebugHelper';
 import * as CalculatorHelpers from './CalculatorHelpers';
 import * as ResultsHelpers from './ResultsHelpers';
+import { generateHTML } from './HTMLGenerator';
 
 /**
  * Calculate measure against a set of patients. Returning detailed results for each patient and population group.
@@ -104,6 +102,12 @@ export function calculate(
         true
       );
 
+      if (options.calculateHTML) {
+        const html = generateHTML(elmLibraries, detailedGroupResult.statementResults, detailedGroupResult.groupId);
+        detailedGroupResult.html = html;
+        dumpHTML(html, `clauses-${detailedGroupResult.groupId}.html`);
+      }
+
       // add this group result to the patient results
       patientExecutionResult.detailedResults?.push(detailedGroupResult);
     });
@@ -155,6 +159,13 @@ export function calculateMeasureReports(
     // create group population counts from result's detailedResults (yes/no->1/0)
     report.group = [];
     result?.detailedResults?.forEach(detail => {
+      // add narrative for relevant clauses
+      if (detail.html) {
+        report.text = {
+          div: detail.html
+        };
+      }
+
       // TODO: check the measure definition for stratification to determine whether to add group.stratiier
       // if yes, add stratifier with population copied into. Set counts to 0 if the result for the stratifier is false
       const group = <R4.IMeasureReport_Group>{};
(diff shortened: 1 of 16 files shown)


src/Calculator.ts
```diff
@@ -155,6 +159,14 @@ export function calculateMeasureReports(
     // create group population counts from result's detailedResults (yes/no->1/0)
     report.group = [];
     result?.detailedResults?.forEach(dr => {
+      // add narrative for relevant clauses
+      if (dr.html) {
+        report.text = {
```
hossenlopp, line 164: For measures with multiple population groups this would only have the HTML for the last population group on the report. They should be appended together.

src/HTMLGenerator.ts
```diff
@@ -0,0 +1,55 @@
...
+      throw new Error(`Could not find library ${s.libraryName} for statement ${s.statementName}`);
+    }
+
+    const matchingExpression = matchingLibrary.library.statements.def.find(e => e.name === s.statementName);
+    if (!matchingExpression) {
+      throw new Error(`No statement ${s.statementName} found in library ${s.libraryName}`);
+    }
+
+    if (matchingExpression.annotation) {
+      annotations = annotations.concat(matchingExpression.annotation);
```
hossenlopp, line 41: Why is `concat` being used instead of `push` to fill the array? Seems like an unnecessary overhead in my opinion.

hossenlopp (changes requested):
Looks good. But it won't show data for each population group and only keeps the last one.

src/HTMLGenerator.ts
mgramigna: matchingExpression.annotation itself is an array. Just using push would create a 2d array instead of just `Annotation[]` like i defined above.

I can probably change it to be like this but might require some flattening logic later on? Not sure.

src/Calculator.ts
mgramigna: Good catch. I just pushed up code that appends rather than overwrites.

To test this, I actually modified the exm130 bundle I was using. I just added a second group in the measure resource within that bundle which is identical to group-1. The resulting HTML narrative had both groups together.

hossenlopp (dismissed):
Looks good! Nice approach that we can build upon as well.

mgramigna:
@hossenlopp just rebased this since Lauren's was merged, Could you give it one more check?

hossenlopp (approved):
Post rebase looks good.

PR #19, opened 2020-11-17
HTML Logic highlighting

mgramigna:
Initial work on this done by @hossenlopp as well.

# Summary

Adds highlighting colors to the HTML output of the ELM annotations

## New behavior

Autogenerated HTML now has style attributes that contain the CSS for highlighting truthy/falsy clasues

## Code changes

* Added a handlebars helper for applying CSS based on final result of the clause
  * Argument to the template now includes library name and clause results in the context to look up the result based on the localId. Used to get the final result of the actual clause
* HTMLGenerator.ts exports the consts used for styles, as well as helper for making the CSS string from the obj
* Test now searches HTML string for proper styling, including cases for false and true

# Testing guidance

See testing instructions for #16, they're basically the same. Colors should be there now.


src/HTMLGenerator.ts
```diff
@@ -2,30 +2,76 @@ import { Annotation, ELM } from './types/ELMTypes';
...
+
+export const cqlLogicClauseFalseStyle = {
+  'background-color': '#edd8d0',
+  color: '#a63b12',
+  'border-bottom-color': '#a63b12',
+  'border-bottom-style': 'double'
+};
+
+/**
+ * Conver JS object to CSS Style string
```
mgramigna: :monkaS:

mgramigna: resolved

mgramigna: thanks

src/HTMLGenerator.ts
```diff
@@ -2,30 +2,76 @@ import { Annotation, ELM } from './types/ELMTypes';
...
+  'background-color': '#edd8d0',
+  color: '#a63b12',
+  'border-bottom-color': '#a63b12',
+  'border-bottom-style': 'double'
+};
+
+/**
+ * Convert JS object to CSS Style string
+ *
+ * @param obj JS object representing CSS styles
```
lmd59, line 28: Could add @return for what the returned string represents

src/HTMLGenerator.ts
```diff
@@ -2,30 +2,76 @@ import { Annotation, ELM } from './types/ELMTypes';
...
+  }
+  return '';
+});
+
 /**
  * Generate HTML structure based on ELM annotations in relevant statements
  *
  * @param elmLibraries main ELM and dependencies to lookup statements
  * @param statementResults StatementResult array from calculation
  * @param groupId ID of population group
```
lmd59, line 63: Similar, could add @return in comments

lmd59 (approved):

PR #18, opened 2020-11-12
episode of care and CV support (and untested ratio/cohort support)

lmd59:
# Summary
Creates appropriate scoring and counting in the measure report for measures with episode of care and CV calculations; contains preliminary support for ratio and cohort measures, but there are not existing fixtures for these, and they remain untested
## New behavior
measure-report reflects appropriate population counts and scores (and evaluated resources as appropriate) for episode of care and CV measures
## Code changes
Adds new Enums, changes episode logic in Calculator.ts' calculateMeasureReport, and adds createMeasureScore function (and other helpers) to handle different measure types
# Testing guidance
Using vs code arguments
 "args": ["-o", "reports","-m", "${workspaceFolder}/test/fixtures/EXM111-9.1.000-bundle-rent.json", "-p", "${workspaceFolder}/test/fixtures/EXM111-9.1.000/Denis399_Gibson10_43d28899-6b06-4442-b93c-7bf8296ec5d5.json"]
The rent bundle has been adjusted to contain the appropriate aggregatemethod extension - I can send to reviewer. 
This patient has 2 relevant episodes with the same 525599 minute value. Fixture patient Armando can also be used for testing, but has no relevant episodes.

PR #20, opened 2020-11-18
working stratifier and stratification score

lmd59:
# Summary
Adds stratification data to the measure report. This includes stratification population counts and stratification measure score (the count and score for all episodes in that stratification).
## New behavior
measure-report reflects appropriate stratification data.
## Code changes
All code changes are in Calculator.ts. This includes building the stratification pieces and a refactor of the measure score logic so that it can be used for both stratification scores and total measure scores.
# Testing guidance
Using vs code arguments
"args": ["-o", "reports","-m", "${workspaceFolder}/test/fixtures/EXM111-9.1.000-bundle-rent.json", "-p", "${workspaceFolder}/test/fixtures/EXM111-9.1.000/Denis399_Gibson10_43d28899-6b06-4442-b93c-7bf8296ec5d5.json"]
Similar to https://github.com/projecttacoma/fqm-execution/pull/18, testing uses the rent bundle (which has been adjusted to contain the appropriate aggregatemethod extension) and patient Denis, which has 2 relevant episodes. Episodes for Denis fall into a single stratification, so stratification counts and scores reflect similar values to those for the group as a whole or all 0/nulls in the case of the other stratification.
Fixture patient Armando can also be used for testing, but has no relevant episodes.

### DIFF
--- a/src/Calculator.ts
+++ b/src/Calculator.ts
@@ -184,8 +184,6 @@ export function calculateMeasureReports(
         report.text.div += detail.html;
       }
 
-      // TODO: check the measure definition for stratification to determine whether to add group.stratiier
-      // if yes, add stratifier with population copied into. Set counts to 0 if the result for the stratifier is false
       const group = <R4.IMeasureReport_Group>{};
       group.id = detail.groupId;
       group.population = [];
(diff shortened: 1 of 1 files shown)


src/Calculator.ts
```diff
@@ -213,7 +211,63 @@ export function calculateMeasureReports(
         });
       }
 
-      createMeasureScore(measure, group, detail, report);
+      // if the measure definition has stratification, add group.stratifier
+      const stratifier = measure.group?.find(g => g.id == detail.groupId)?.stratifier;
+      if (stratifier) {
+        group.stratifier = [];
+        stratifier.forEach(s => {
```
mgramigna, line 218: From the [structure def for MeasureReport](https://www.hl7.org/fhir/measurereport.html#invs):

> Stratifiers SHALL be either a single criteria or a set of criteria components (group.stratifier.stratum.all(value.exists() xor component.exists()))

This is easy to miss since it's an additional constraint that isn't in the structure definition. Since the TS interfaces are also based on the structure def, you won't get a TypeError either. I noticed this by trying to use the FHIR validator on the MeasureReport to ensure that the structure of the stratifier is in line with the spec

src/Calculator.ts
```diff
@@ -213,7 +211,63 @@ export function calculateMeasureReports(
...
+          const reportStratifier = <R4.IMeasureReport_Stratifier>{};
+          reportStratifier.code = s.code ? [s.code] : [];
+          const strat = <R4.IMeasureReport_Stratum>{};
+          // use existing populations, but reduce count as appropriate
+          strat.population = JSON.parse(JSON.stringify(group.population));
```
mgramigna, line 223: Q: what's the need for the parse/stringify combo here?

src/Calculator.ts
```diff
@@ -213,7 +211,63 @@ export function calculateMeasureReports(
...
+            });
+          }
+
+          const scoringCode =
+            measure.scoring?.coding?.find(c => c.system === 'http://hl7.org/fhir/measure-scoring')?.code || '';
+          if (scoringCode === MeasureScoreType.CV) {
+            // TODO: should we add score observation for stratification?
+            strat.measureScore = calcMeasureScoreCV(measure, detail, group.id || '', s);
+          } else {
+            strat.measureScore = calcMeasureScore(scoringCode, JSON.parse(JSON.stringify(strat.population)));
```
mgramigna, line 254: Q: what's the need for the parse/stringify combo here?

mgramigna (changes requested):

src/Calculator.ts
mgramigna: So we need to add a value to these, I'm not entirely sure what this should be. It can probably just match the "text" that you put in the code property for now?

src/Calculator.ts
lmd59: Deep copy + matching type structures that aren't actually the same type

src/Calculator.ts
lmd59: Deep copy + matching type structures that aren't actually the same type

mgramigna (approved):

PR #25, opened 2020-12-15
Integration test fixtures

MegRiley:
# Summary
## New behavior
With these changes when there is a new pull request submitted a process similar the CI  process that is run on the connectathon repo should now be run. The process that should happen is as follows:

1.  Build the library (this is not new)
2.  Pull the data from the connectathon repo
3.  create the measure bundles and patient bundles 
4. run these bundles through fqm execution and compare the results against reference reports

## Code changes
Most of the changes for this code are in new files that were added to the newly created "_tests_" folder which is under the .github folder that was added with the move to github actions.  With a few of these files it may be easier to review them against the connectathon repo. The bulk of the code changes are in: 

-  **setup.h**  - this file handles pulling the needed data down (a case could be made to roll this into the testMeasures.sh)
- **test_measures.sh** - this file has had it's name carried over from connectathon repo, and it calls the typscript test files and passes the paths to test files
- **run_tests.ts** - This is the file that invokes the  logic to execute the measures and run the comparisons 
- **testDataHelpers.js** - I didn't make many changes in this file except to change cqf references to fqm, it is largely unchanged from the connectathon repo
- **fhirInteractions.ts**- I changed this file from js to ts, and removed the logic that was needed to interact with a fhir server, and could potentially remove this file completely, I left it in because I wasn't completely sure if in the future the fqm-service and fqm execution repo might join together. 
- **rebuild_bundles.sh** - this file is also largely unchanged in the move from the connectathon repo, just updating cqf ruler references and pathing

# Testing guidance
Currently there are no tests to run because I'm not sure where the stock reports are supposed to exist/come from. A user should be able to run setup.sh and then launch test_measures.sh.


.github/workflows/ci.yml
```diff
@@ -21,5 +21,5 @@ jobs:
     - run: npm run check
     - env:
         VALUESET_KEY: ${{ secrets.VALUESET_KEY }}
-      run: unzip -P $VALUESET_KEY test/fixtures/EXM130-8.0.000/EXM130-8.0.000-bundle.zip
-    - run: rm EXM130-8.0.000-bundle.json
+    - run: ./setup.sh
```
hossenlopp, line 24: The EXM130 package still needs to be extracted for the unit tests to run AFAIK.

_tests_/setup.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+echo "> Cloning FHIR Generated Patients..."
+git clone --single-branch --branch master https://github.com/projecttacoma/fhir-patient-generator.git
+
+echo "> Fetching CQF-tooling JAR"
+./_tests_/download_cqf_tooling.sh
```
hossenlopp, line 7: This is not needed.

_tests_/setup.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+echo "> Cloning FHIR Generated Patients..."
+git clone --single-branch --branch master https://github.com/projecttacoma/fhir-patient-generator.git
```
hossenlopp, line 4: The connectathon repo `https://github.com/DBCG/connectathon.git` also needs to be cloned.

.github/workflows/ci.yml
hossenlopp: These two script references are likely incorrect as they are not in the current folder.

hossenlopp (changes requested):
This fails to run in github actions and the Connectathon repo is not cloned. cqf-tooling and measure bundle rebuilding does not need to be done. That setup can be removed.

There is a lot of work in progress code that appears to be for TACOSTRAT-466. This code is breaking the CI runs because of typescript issues. This code should be moved to another branch so it can be incorporated and reviewed separately.

Additionally, this branch needs to be rebased on master.

_tests_/fhirInteractions.ts
```diff
@@ -0,0 +1,34 @@
+
```
hossenlopp, line 1: This file should be moved to a different branch for the follow on work to run against the fixture data.

hossenlopp, line 1: This file should be moved to a different branch for the follow on work to run against the fixture data.

_tests_/measureReportCompare.ts
```diff
@@ -0,0 +1,132 @@
+/**
```
hossenlopp, line 1: This file should be moved to a different branch for the follow on work to run against the fixture data.

_tests_/run_tests.ts
```diff
@@ -0,0 +1,135 @@
+/**
```
hossenlopp, line 1: This file should be moved to a different branch for the follow on work to run against the fixture data.

_tests_/setup.sh
```diff
@@ -0,0 +1,10 @@
...
+
+echo "> Cloning connectathon repo"
+git clone --single-branch --branch master https://github.com/projecttacoma/connectathon.git  
+
+echo "> running the tests"
+node ./run_tests.ts
```
hossenlopp, line 9: Running the tests should happen in a follow on PR.

_tests_/testDataHelpers.ts
```diff
@@ -0,0 +1,107 @@
+import cql from 'cql-execution';
```
hossenlopp, line 1: This file should be moved to a different branch for the follow on work to run against the fixture data.

hossenlopp (changes requested):
This still fails to run in GithubCI. The TypeScript code for running the tests should be moved to a separate PR.

.github/workflows/ci.yml
```diff
@@ -23,3 +23,5 @@ jobs:
         VALUESET_KEY: ${{ secrets.VALUESET_KEY }}
       run: unzip -P $VALUESET_KEY test/fixtures/EXM130-8.0.000/EXM130-8.0.000-bundle.zip
     - run: rm EXM130-8.0.000-bundle.json
+    - run:  sudo chmod +x ./_tests_/setup.sh
```
hossenlopp, line 26:
```suggestion
    - run: sudo chmod +x ./_tests_/setup.sh
    - run: ./_tests_/setup.sh
```
The `setup.sh` needs to also be run.

_tests_/setup.sh
hossenlopp: Remove this. since the file does not exist. Just the cloning of the repos is needed. Anyways it would be good to separate the running of the tests from the setup work.

hossenlopp (changes requested):

hossenlopp (approved):
This appears to pull down the the needed fixtures in Github actions. When merging, a squash merge will be done so the history can be kept clean.

PR #26, opened 2020-12-15
Add example detected issue fixture

mgramigna:
This PR showcases how we plan to model gaps in care using the [DEQM DetectedIssue Profile](http://build.fhir.org/ig/HL7/davinci-deqm/StructureDefinition-gaps-detectedissue-deqm.html). This requires 2 extensions. Neither of which actually exist yet, but can be added to the ballot. The URLs i've included here are just for demonstration.

1) `http://hl7.org/fhir/us/davinci-deqm/StructureDefinition/extension-gap-data`: this extension will contain the dataRequirement for this given query for which data type and code/valueset it is looking for
2) `http://hl7.org/fhir/us/davinci-deqm/StructureDefinition/extension-gaps-status`: this extension will indicate if the gap is due to missing data (open-gap) or present data (closed-gap)

The only review that's required here is visual inspection

### DIFF
--- a/test/fixtures/gaps/example-detected-issue.json
+++ b/test/fixtures/gaps/example-detected-issue.json
@@ -1,5 +1,6 @@
 {
   "resourceType": "DetectedIssue",
+  "status": "final",
   "code": {
     "coding": [
       {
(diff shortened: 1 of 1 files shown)


okeefm (approved):
Looks good to me.

PR #27, opened 2020-12-17
Debug Refactor

mgramigna:
# Summary

Refactor debug output behavior to be optional, and to not use `fs` in any core pieces of the code

## New behavior

Debug output is now a property on the return result of `calculate`, `calculateRaw`, and `calculateMeasureReports`. the CLI will dump this to a `debug` folder as before when the debug option is specified at runtime of the CLI.

## Code changes

* Modify core functions to return `{ results: CurrentReturnType, debugOutput: DebugOutput }` type motif
* Move debug output writing functionality to purely CLI
* Add `-d/--debug` option to CLI and `enableDebugOutput` calculation option
* update docs

# Testing guidance

* Remove your existing debug folder to start from clean slate
* Run the CLI first without the `-d` option (there should be no debug folder created)
* Run the CLI with the `-d` option. `debugOutput` should be present on the results and all the same debug info will be dumped to the `debug` folder as before

PR #28, opened 2020-12-17
Browser friendly

mgramigna:
# Summary

DRAFT: Review #27 first.

This PR removes non-browser-friendly code from the library. Namely, the usage of `fs` in the HTMLGenerator. Everything else is fair game. Webpack/other bundlers were not needed.

## New behavior

Nothing really. Build script no longer copies over handlebars files as there are none now.

## Code changes

* Add the fhir types library to regular dependencies: we require it directly into the app.
* Change handlebars templates to be strings exported by ts files

# Testing guidance

* Clone [this repo](https://github.com/mgramigna/fqm-example-browser-usage)
* npm install
* npm start

Note that this example app that I made uses fqm-execution as a dependency. An app will load in the browser with a button that will call fqm-execution's `calculate` function directly in the browser. Calculation results will be rendered on the screen.

Also worth testing that html generation still works by running with output type `reports` and confirming that html output is as expected.


okeefm (approved):
Tested, still works locally, and works in the demo browser app

PR #29, opened 2020-12-18
Update gaps in care to generate FHIR gaps in care Bundle

mgramigna:
# Summary

Gaps in care functionality used to return just a list of queries. Now will return a [DEQM Gaps in Care Bundle](http://build.fhir.org/ig/HL7/davinci-deqm/StructureDefinition-gaps-bundle-deqm.html)

The Bundle contains the following:

* A composition resource that references the measure report and a FHIR DetectedIssue resource generated during gaps in care calculation
* The patient, measurereport, and detectedissue resources so that all references can resolve nicely

## New behavior

Gaps in care operation returns bundle adhering to above format

## Code changes

* Add new functions for generating both detectedIssue and Bundle resources
* Added unit tests for above functions
* Updated debug output for gaps to include both bundle and queries
* Update fixtures to adhere to unit tests

# Testing guidance

Run cli with output type gaps, EXM130, and a denominator patient. Turn on debugging with `-d` to inspect the query/bundle output in the debug folder. run tests.


hossenlopp (dismissed):
Looks good. Tested with both patients in numerator and not. The one that made it into the numerator had an empty object result, which seems correct.

One suggestion for future refactoring is to move the gaps in care related functions into their own file to separate the gaps code from the core calculation code.

mgramigna:
Rebased and moved gaps helpers to own file

hossenlopp (approved):
Refactor looks good and works. 👍

PR #32, opened 2021-01-22
Refactored calculation code to allow better reuse of calculation artifacts

hossenlopp:
# Summary
Moves MeasureReport building to its own file and adds an option for `calculate` to return the elm libraries and main library elm so they can be used for further processing. 

https://jira.mitre.org/browse/TACOSTRAT-552

## New behavior
None.

## Code changes
MeasureReportBuilder.ts is simply MeasureReport building code moved out of Calculator.ts. Gaps in care now calls the MeasureReportBuilder code instead of kicking off a re-calculate. `calculate` call now returns the ELM JSON and main library name if requested.

# Testing guidance
Check that detailed, reports and gaps still function.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3835,7 +3835,8 @@
       "version": "1.3.0",
       "resolved": "https://registry.npmjs.org/growly/-/growly-1.3.0.tgz",
       "integrity": "sha1-8QdIy+dq+WS3yWyTxrzCivEgwIE=",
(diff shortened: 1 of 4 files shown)


mgramigna (approved):
Nice approach. Looks good to me.

We can have a separate housekeeping task for unit-testing the measure report builder now that that's stubbed out the way it is. Not your responsibility here.

PR #33, opened 2021-01-27
Measure hl tests

MegRiley:
# Summary
These changes expand upon a prior changes to add integration test fixtures. 
## New behavior
in addition to downloading the needed setup files, this code will now execute integration tests using the files that were downloaded. Hopefully testing the base functionality of fqm execution and measure highlighting
## Code changes
New files were added: 
**test_measures.sh** - this file has had it's name carried over from connectathon repo, and it calls the typscript test files and passes the paths to test files
**run_tests.ts** - This is the file that invokes the logic to execute the measures and run the comparisons
**testDataHelpers.ts** - I didn't make many changes in this file except to change cqf references to fqm, it is largely unchanged from the connectathon repo
**fhirInteractions.ts**- I changed this file from js to ts, and removed the logic that was needed to interact with a fhir server, 
**testhighlighting.ts**- this file was added to support the testing of measure highlighting 
**ci.yml**- Modified this file to call test_measures.sh and pass in the list of measures to execute. 
**package.json** - added a "test:integration" script to no longer directly call the ts files


.github/workflows/ci.yml
```diff
@@ -23,6 +24,6 @@ jobs:
         VALUESET_KEY: ${{ secrets.VALUESET_KEY }}
       run: unzip -P $VALUESET_KEY test/fixtures/EXM130-8.0.000/EXM130-8.0.000-bundle.zip
     - run: rm EXM130-8.0.000-bundle.json
-    - run: sudo chmod +x ./_tests_/setup.sh
-    - run: ./_tests_/setup.sh
-  
+    - run: sudo chmod +x ./src/_tests_/setup.sh
```
mgramigna, line 27: Looks like you only chmod and never run the setup script.

We're still failing GH actions now it looks like the error is bc it can't find any test data (i.e. fpg doesn't exist)

Just run the setup script before running test_measure

src/_tests_/testDataHelpers.ts
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * Grabs the list of measures that fhir-patient-generator has test data for.
+ *
+ * @returns {Promise<TestMeasureInfo[]>} List of information about each available test measure data.
+ */
+export function getTestMeasureList() {
+  // Find applicable measure test data directories in fhir-patient-generator
+
+  const fpgPath = './src/_tests_/fhir-patient-generator';
+  const prefixPath = './src/_tests_/connectathon/fhir401/bundles/measure/';
```
mgramigna, line 35:
```suggestion
  const fpgPath = './fhir-patient-generator';
  const prefixPath = './connectathon/fhir401/bundles/measure/';
```

src/_tests_/testDataHelpers.ts
```diff
@@ -0,0 +1,136 @@
...
+  });
+
+  // Pull out applicable measure information on them
+  /** @type {TestMeasureInfo[]} */
+  const measureDirInfo = applicableMeasuresDirs.map(measureDir => {
+    /** @type {TestMeasureInfo} */
+    const testDirInfo = {
+      // format exmId into simple, non-versioned id
+      exmId: measureDir.includes('-') ? measureDir.split('-')[0] : measureDir,
+      path: `./src/_tests_/fhir-patient-generator/${measureDir}/patients-r4`,
```
mgramigna, line 48:
```suggestion
      path: `./fhir-patient-generator/${measureDir}/patients-r4`,
```

mgramigna (changes requested):

mgramigna (changes requested):
Can you remove the checked in copies of connectathon and fhir-patient-generator from `src/_tests_` and add them to the .gitignore instead?

src/_tests_/measureReportCompare.ts
```diff
@@ -0,0 +1,112 @@
...
+          if (!(population?.count == referencePopulation.count)) {
+            if (!(population?.count == 0 && referencePopulation.count == 0)) {
+              console.log('        MISSING  ' + patientName);
+
+              addBadPatientEntry(badPatientsList, patientName, true, `Missing from ${popName}`);
+            } else if (!(population?.count > 0 && referencePopulation.count > 0)) {
+              addBadPatientEntry(badPatientsList, patientName, true, `Missing from ${popName}`);
+            }
+          }
+        }
```
mgramigna, line 106: The boolean logic here is slightly incorrect. This will still make it such that any inequality between the counts will result in a missing patient. 

If the 2 counts are not equal, then it will check if they are both non zero, which will always be true.

You can restructure this in a way such that 

```
if (they're equal)
  return;
else if they're both greater than 0
  return;
else 
  they're bad and add them to missing
```

mgramigna (changes requested):

.gitignore
```diff
@@ -5,3 +5,7 @@ debug/
 **/.DS_Store
 test/fixtures/elm/queries/translation-output
 test/fixtures/elm/queries/.start-translator
+_tests_/fhir-patient-generator/
+src/_tests_/connectathon
+src/_tests_/connectathon
+_tests_/connectathon/
```
mgramigna, line 11: I think this only should be root directory ones since that's where they're cloned now

```suggestion
fhir-patient-generator
connectathon
```

src/_tests_/measureReportCompare.ts
```diff
@@ -0,0 +1,114 @@
...
+        // find corresponding population
+
+        const population = findCorrespondingPopulation(referencePopulation, group);
+
+        if (population?.count && referencePopulation?.count) {
+          if (population?.count == referencePopulation.count) {
+            return;
+          } else if (
+            (population?.count == 0 && referencePopulation.count == 0) ||
```
mgramigna, line 100:
```suggestion
```

This line is never going to be reached and is unnecessary

mgramigna (changes requested):
Small comments.

Plus, could you also fix the typos of `connectahton -> connectathon` in the variable names to avoid confusion? Global search and replace in VS Code should make this easy

mgramigna (approved):

mgramigna (changes requested):
Can we separate the measure integration tests to their own job? So right now we have one job, `build`, that does everything. We should have one job called `checks` or something that does the `npm ci` `npm run build --if-present` `npm run check` commands

then another job `measure-tests` that does the integration tests per each measure

I think the former should be done for both node 12.x and 14.x, but the measure integration tests only need to run on node 12.x.

Also... can you get rid of the unzipping and VALUESET_KEY stuff? we don't need that anymore now that we have these integration tests

mgramigna (approved):

PR #35, opened 2021-01-28
Add makefile for translating cql test fixtures in gaps

mgramigna:
Add Makefile for translating all cql files in the fixtures/elm/queries folder.

User will have to manually copy over the outputted elm json for now because I'm not in the business of parsing out each piece of the multipart form response especially not via bash/makefile.

`make` to translate everything
`make clean` to clean

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ build/
 debug/
 .vscode
 **/.DS_Store
+test/fixtures/elm/queries/translation-output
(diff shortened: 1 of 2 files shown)


hossenlopp (approved):
Definitely helpful and worth getting in now. Perhaps later we can expand on it to pull out the multipart pieces and show when there are CQL issues.

PR #36, opened 2021-02-03
Initial backbone of improvement notation logic

mgramigna:
# Summary

First pass at setting the code up to process negative improvement measures. Basically, all I think we can do right now is change the boolean logic to switch on the improvement notation to determine when it needs to traverse the numerator queries.

Worth noting: I decide to move away from the "satisfied" nomenclature, as it will inevitably get confusing when we start dealing more with the negative improvement logic. I renamed it to `___HasResult` instead, indicating whether or not the retrieve or query came back with a result or not. This way, the notion of what being "satisfied" means is up to the improvement notation of the measure

## New behavior

The core functionality is the same, only now the evidence is filtered according to both the improvement notation of the measure and the status of whether or not the query had a truthy result or not.

* For positive improvement (most) measures, queries that have results are **not** considered gaps
* For negative improvement measures, queries that have results are considered to be gaps since we don't want them to be there if the numerator is doing those queries

## Code changes

* Made the nomenclature change mentioned above
* Added filter on query result to exclude non-gaps from evidence based on improvement notation of the measure
* Update gaps logic to traverse queries based on improvement notation
  * If negative improvement, check only for numerator patients
  * If positive improvement, check for patients that are in the denominator but not the numerator
* Add test cases for the above filter in detected issue generation

# Testing guidance

Testing recommended with EXM130 since we don't have episode of care support yet (EXM506 is an episode of care negative improvement measure, but we don't have a FHIR-based regular negative improvement measure yet)

* run usual checks
* run gaps for a normal measure normal denominator patient and ensure results are as expected.
* Update the EXM130 measure resource in the bundle you're using to have negative improvement instead by changing the code:

```
      "improvementNotation": {
        "coding": [ {
          "system": "http://terminology.hl7.org/CodeSystem/measure-improvement-notation",
          "code": "decrease"
        } ]
      },
```
  * After doing this update, a denominator patient should have no gaps output
  * A numerator patient should have the same gaps output now, but the satisfied queries are treated as gaps since we don't want them on the patient's record if this were a negative improvement measure

thoughts?


okeefm (dismissed):
Tested locally by swapping the improvement notation in a measure bundle. Looks good to me!

okeefm (approved):
Tested with the rebased modeling changes, still looks good!

PR #37, opened 2021-02-05
Add helper for extracting measurement period from Measure, and set sane defaults in Calculator.ts

okeefm:
# Summary
This PR adds "sane" defaults for:

* `calculateHTML`: `true`
*  `calculateSDEs`: `true`
* `measurementPeriodStart`/`measurementPeriodEnd`: Whatever the measure sets it to, or `2019-01-01`/`2019-12-31` respectively, if not set in the measure `effectivePeriod` object.
 
 to `Calculator.ts`. It also adds the ability to pass in measurement period strings on the command line, to override the measure.

## New behavior
See above, regarding measurement period default changes.

## Code changes
Added `extractMeasurementPeriod` to `MeasureHelpers.ts` to pull the `effectivePeriod.start`/`.end` out of the `Measure` object, and use that to set defaults in `Calculator.ts`

# Testing guidance
Use the command line `-s` and `-e` flags to test changing the measurement period when calling `fqm-execution`. Using a VS Code debug session, drop a breakpoint in `Calculator.ts` to check the measurement period in `options`.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,7 +1,7 @@
 module.exports = {
   globals: {
     'ts-jest': {
-      tsConfig: 'tsconfig.json'
+      tsconfig: 'tsconfig.json'
     }
   },
   moduleFileExtensions: ['ts', 'js', 'd.ts'],
--- a/README.md
+++ b/README.md
@@ -84,6 +84,8 @@ Options:
   -o, --output-type <type>                    type of output, "raw", "detailed", "reports", "gaps" (default: "detailed")
   -m, --measure-bundle <measure-bundle>       path to measure bundle
   -p, --patient-bundles <patient-bundles...>  paths to patient bundle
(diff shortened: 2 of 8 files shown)


jest.config.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   globals: {
     'ts-jest': {
-      tsConfig: 'tsconfig.json'
+      tsconfig: 'tsconfig.json'
```
mgramigna, line 4: Nice. been meaning to do this lol

mgramigna (changes requested):
Looks good. Can you also update package-lock.json to have the proper version number? This is best done using the `npm version` command which auto updates package, package-lock and creates a git tag.

So I'd undo your change in package.json, bringing the version back to 0.1.0, then run `npm version minor` which will increment both and create a tag. Then push those up

mgramigna (approved):

PR #39, opened 2021-02-15
Add stack of expressions traversed during gaps in care to output

mgramigna:
# Summary

Adds a property to the external data model for GiC that tracks each expression traversed during the calculation. Each entry on the stack indicate the expression type, and the local ID/library for lookup later.

This is not used right now, but will be useful when we eventually start to handle more logic like `Or`s and `Not`s

## New behavior

N/A. Just another property that will exist on the debug output for now. Will be used later.

## Code changes

* Add new property to `DataTypeQuery` interface for the stack of expressions
* Push to stack during recursion of gaps in care traversal
* Add unit test cases

# Testing guidance

* Run usual checks
* Inspect the debug output when running gaps to confirm the presence/format of the expression stack (`debug/gaps.json`)


okeefm (approved):
Looks good to me

PR #40, opened 2021-03-02
Do not calculate gaps for denominator exclusion or exception patients

lmd59:
# Summary
Updates gaps calculation so that gaps are not calculated for patients that fall into the denominator exception or denominator exclusion.
## New behavior
Adds a check of detailed results population relevance. If the numerator is false, then this patient was never even considered for the numerator, so there are no gaps that could have gotten it into the numerator. (Or for negative improvement notation, it can't be in the negatively-associated numerator). Now gaps are not generated for this case of patients for which there are no gaps.
## Code changes
Small changes in Calculator.ts
# Testing guidance
This was tested with connectathon's [EXM124-9.0.000](https://github.com/DBCG/connectathon/blob/master/fhir401/bundles/measure/EXM124-9.0.000/EXM124-9.0.000-bundle.json). [Denex patient](https://github.com/DBCG/connectathon/blob/master/fhir401/bundles/measure/EXM124-9.0.000/EXM124-9.0.000-files/tests-denomexcl-EXM124-bundle.json) now does not generate gaps, but [Denom patient](https://github.com/DBCG/connectathon/blob/master/fhir401/bundles/measure/EXM124-9.0.000/EXM124-9.0.000-files/tests-denom-EXM124-bundle.json) still does generate gaps.

### DIFF
--- a/src/Calculator.ts
+++ b/src/Calculator.ts
@@ -241,6 +241,7 @@ export function calculateGapsInCare(
 
       const denomResult = dr.populationResults?.find(pr => pr.populationType === PopulationType.DENOM)?.result;
       const numerResult = dr.populationResults?.find(pr => pr.populationType === PopulationType.NUMER)?.result;
+      const numerRelevance = dr.populationRelevance?.find(pr => pr.populationType === PopulationType.NUMER)?.result;
 
       if (!measureResource.improvementNotation?.coding) {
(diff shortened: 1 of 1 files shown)


okeefm (approved):
Tested with EXM124 and EXM74. Works with both. Looks good!

PR #42, opened 2021-03-10
Initial near miss

lmd59:
# Summary
Establishes an initial approach for finding certain near miss gaps. This identifies gaps where the data criteria required by the retrieve was found, but the query it belongs to was not satisfied.
## New behavior
For positive improvement, identifies all data queries where the retrieve is true, but the query is false and marks these as near misses such that they may eventually be identified in the gaps report, dependent on desired formatting.
## Code changes
Adds calculateNearMisses function. Calls `GapsInCareHelpers.calculateNearMisses` on all retrieves found in the calculator. 
# Testing guidance
This was tested with connectathon's [EXM124-9.0.000](https://github.com/DBCG/connectathon/blob/master/fhir401/bundles/measure/EXM124-9.0.000/EXM124-9.0.000-bundle.json) and adjusted [numerator patient](https://github.com/DBCG/connectathon/blob/master/fhir401/bundles/measure/EXM124-9.0.000/EXM124-9.0.000-files/tests-numer-EXM124-bundle.json) with observation time shifted several years earlier. Additionally, a test has been added which can be run with `npm test`

### DIFF
--- a/test/GapsInCareHelpers.test.ts
+++ b/test/GapsInCareHelpers.test.ts
@@ -1,5 +1,10 @@
 import { R4 } from '@ahryman40k/ts-fhir-types';
-import { findRetrieves, generateDetectedIssueResource, generateGapsInCareBundle } from '../src/GapsInCareHelpers';
+import {
+  findRetrieves,
+  generateDetectedIssueResource,
+  generateGapsInCareBundle,
+  calculateNearMisses
+} from '../src/GapsInCareHelpers';
 import { DataTypeQuery, DetailedPopulationGroupResult } from '../src/types/Calculator';
 import { FinalResult, ImprovementNotation } from '../src/types/Enums';
 import { getELMFixture, getJSONFixture } from './helpers/testHelpers';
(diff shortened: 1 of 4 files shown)


test/GapsInCareHelpers.test.ts
```diff
@@ -307,6 +312,17 @@ describe('Generate DetectedIssue Resource', () => {
   });
 });
 
+describe('Find Near misses', () => {
+  test('simple query/retrieve discrepancy near misses', () => {
+    const retrieves = calculateNearMisses(EXPECTED_CODE_RESULTS, ImprovementNotation.POSITIVE);
+    console.log('retrieves:', retrieves);
```
mgramigna, line 318:
```suggestion
```

mgramigna (changes requested):
1) Small removal of console log

2) Is there a way we can add a test for the case where we expect isNearMiss to be falsy? i.e. we should test when query and retrieve are both false for positive improvement. the variable should not be truthy in that case.

mgramigna (approved):
LGTM

PR #44, opened 2021-03-15
Improve CQL translation in fixtures

okeefm:
# Summary
Adds a Typescript script for running CQL through the translation service, and splitting out the resulting ELM JSON into separate files in `test/fixtures/elm/queries/output`. Makes use of https://github.com/mcode/cql-translation-service-client to do the heavy lifting.

## New behavior
* `Makefile` now calls `cql-translator.ts` to do the CQL translation using `ts-node`
* ELM JSON is now written to `test/fixtures/elm/queries/output` as individual files
* `make clean` no longer deletes the output JSON
* `make` _will_ delete old output JSON, prior to re-translating.

## Code changes
* Added `cql-translator.ts`, which uses the `cql-translation-service-client` to translate every `.cql` file in `test/fixtures/elm/queries/output`

# Testing guidance

## Happy path
* `cd` to `test/fixtures/elm/queries`
* run `make`
* Ensure the two `.cql` files that are already in that dir properly translate, and the translated ELM for both shows up in the `output/` folder

## Sad path
* `cd` to `test/fixtures/elm/queries`
* Break one of the `.cql` files (easy way to do this: delete a `:` after any of the `define` statements)
* run `make`
* Ensure the broken file is _not_ output to the `output/` directory, and that an error is spit out on the command line

## Extra credit
Run some of your own CQL files through the `make` process.

### DIFF
--- a/test/fixtures/elm/queries/cql-translator.ts
+++ b/test/fixtures/elm/queries/cql-translator.ts
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/ban-types */
+
 import fs from 'fs';
 import path from 'path';
 import * as translationService from 'cql-translation-service-client';
--- a/test/fixtures/elm/queries/cql-translator.ts
+++ b/test/fixtures/elm/queries/cql-translator.ts
@@ -1,0 +1,108 @@
+import fs from 'fs';
+import path from 'path';
+import * as translationService from 'cql-translation-service-client';
+
+const TRANSLATION_SERVICE_URL = 'http://localhost:8080/cql/translator';
+
+const client = new translationService.Client(TRANSLATION_SERVICE_URL);
+
+// This interface is extended here to add the "annotation" property
+// because the client type sets its value type to "object" rather than "object[]"
+interface ElmLibrary extends translationService.ElmLibrary {
+  library: {
+    identifier: {
+      id: string;
+      version: string;
+    };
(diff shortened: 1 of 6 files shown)


test/fixtures/elm/queries/cql-translator.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+translateCQL()
+  .then(libraries => {
+    const buildPath = path.join(__dirname, './output');
+    Object.entries(libraries).forEach(([libName, elm]) => {
+      const errors = processErrors(elm as ElmLibrary);
+      if (errors.length === 0) {
+        const elmPath = path.join(buildPath, `${libName}.json`);
+        fs.writeFileSync(elmPath, JSON.stringify(elm), 'utf8');
```
hossenlopp, line 97:
```suggestion
        fs.writeFileSync(elmPath, JSON.stringify(elm, undefined, 2), 'utf8');
```
Pretty much always need to inspect the elm to work on expected results for unit tests. Doing pretty output would be ever useful.

hossenlopp (changes requested):
Works well. Just would like some pretty print output.

test/fixtures/elm/queries/cql-translator.ts
okeefm: Good call

hossenlopp (approved):

PR #46, opened 2021-03-16
Added guidance response fixtures for modeling enhancements

mgramigna:
No changes, just having our example guidances responses controlled somewhere so we can use for unit tests later once we implement more gaps parsing.


p9g:
suggest generating a new uuid for the .id for each example

okeefm (approved):
Examples look good!

PR #47, opened 2021-03-18
Initial revision

MegRiley:
First pass of the architecture diagram  following a discussion at the dev hour

# Summary
I've upload the source file and the html version of the diagram we discussed on Thursday. There are 2 files 
- FQM.drawio - source file that can be edited 
- FQM.html - rendered version of diagram
 The diagram consists of 2 pages:
- Page 1: Walks through the overall flow depending upon the output type that is selected by the user
- Page 2: This is a more in depth breakdown of calculate() function because that is where the bulk of the ecqm specific work occurs and this is common regardless of the output type that has been selected


mgramigna (changes requested):
1) This is insanely nitpicky, but would you mind straightening all the arrows that are misaligned? e.g.: 
image. I think there's only 2

2) I think it would make sense to have these files live in a separate `docs`, but still outside of `src`

3) Maybe we can add some screenshots of these and an updated section of the README that indicates this flow?

mgramigna:
Looks good, just needs a rebase

mgramigna (approved):

PR #50, opened 2021-04-07
Library changes

MegRiley:
Changes to support MAT generated measures that have URLs for the root…

# Summary
This change will now allow  for fqm execution to eventually support MAT generated measures that have urls as their root libraries instead of a fhir resource
## New behavior
This change should be transparent and the behavior shouldn't change
## Code changes
-execution.ts - changed the execute function to now check if a library is valid if it is either a fhir resource or a url
-measureHelpers.ts - added  a new method that accepts a string checks that this string is indeed a valid url (It may not be a valid library but that it is a valid url)
# Testing guidance
Using the debugger I ran this against EXM130 with the same patient  (the mat generated and the original version of the measrure in the 7.3.000 folder) I checked the that the elmLibraries = results.elmLibraries; was similar for both cases. (they were not identical, the libraries appeared to be in a different order and the MAT generated one included a mat specific library)


src/MeasureHelpers.ts
```diff
@@ -403,3 +403,14 @@ export function extractMeasurementPeriod(measureBundle: R4.IBundle): Calculation
 function __guard__(value: any, transform: any) {
   return typeof value !== 'undefined' && value !== null ? transform(value) : undefined;
 }
+/**
+ * Figure out if a string is a  valid URL  or if it's just a string
+ * @public
+ * @param {string} libraryName - The name of the library depending on the origin of the measure this may a url
+ * @return {boolean} If the statement is a function or not.
+ */
+export function isValidLibraryURL(libraryName: string) {
```
okeefm, line 412: We may want to add some positive/negative tests for this, if possible; something like calling `isValidLibraryURL` on `https://example.com/Library-url` and `Library/example` or the like, and ensuring the results are what we expect.

okeefm (changes requested):
One place we may want to add some tests, but other than that looks good!

mgramigna:
## jest coverage report 🧪

### Total coverage

|   🔴   | Statements | 28.34%     |     352/1242    |
|   🔴   | Branches   | 13.61%     |     215/1580    |
|   🔴   | Functions  | 24.68%     |      57/231     |
|   🔴   | Lines      | 28.37%     |     326/1149    |

<details><summary>Show files with reduced coverage 🔻</summary>


### Reduced coverage

| Status | Filename          | Statements         | Branches | Functions          | Lines              |
| :----: | :---------------- | :----------------- | :------- | :----------------- | :----------------- |
|   🔴   | src               | 25.08%             | 12.84%   | 21.27%             | 24.84%             |
|   🔴   | Execution.ts      | 7.87%              | 0%       | 0%                 | 8.43%              |
|   🟡   | MeasureHelpers.ts | 61.38% (-1.03% 🔻) | 68.53%   | 55.56% (-3.26% 🔻) | 61.42% (-1.18% 🔻) |

</details>

okeefm (approved):
Looks good!

PR #51, opened 2021-04-09
support for pulling all dependent valusets from a measure bundle

lmd59:
# Summary
Creates functionality for identifying all dependent valuesets in a measure bundle.
## New behavior
Adds a function that takes a measure bundle and returns an array of dependent valuesets.
## Code changes
Adds getAllDependentValuesets function to CalculatorHelper.ts. Adds unit test file CalculatorHelper.test.ts that uses a fixture EXM130 measure bundle that has all non-demographic codes replaces with fake placeholder codes.
# Testing guidance
A test has been added which can be run with `npm test`


mgramigna:
## jest coverage report 🧪

### Total coverage

| Status | Category   | Percentage         | Covered / Total |
| :----: | :--------- | :----------------- | :-------------: |
|   🔴   | Statements | 29.71% (+1.27% 🔼) |     374/1259    |
|   🔴   | Branches   | 15.08% (+1.44% 🔼) |     243/1611    |
|   🔴   | Functions  | 27.62% (+2.84% 🔼) |      66/239     |
|   🔴   | Lines      | 29.76% (+1.3% 🔼)  |     347/1166    |

mgramigna (changes requested):
I think the approach here should be slightly modified. I don't think that the `getAllDependentValueSets` function needs to return the actual ValueSet resources. We can just make it return the list of URLs or IDs or whatever that the measure bundle needs.

After this, we can have a separate function that just takes in the measure bundle and returns a boolean indicating whether or not that measure bundle does indeed contain all of the required valuesets. Basically, I think just changing the last few statements of this function and moving them out to a new function that does this shouldn't be too hard.

Lastly, in the `execute` function, we should make a call to this boolean check and throw an Error if the measure does not contain all the required valuesets

mgramigna (approved):

PR #54, opened 2021-04-21
Or gap grouping

okeefm:
# Summary
This Pull Requests adds (very basic) support for "or gaps" in the gaps-in-care calculator. 

An Or gap is when any of several gaps in care would, by itself, put the patient into the desired population (usually the numerator) if addressed. The converse is an And gap (which we define as any non-Or gap, at the moment), where you need each of several gaps to be addressed to put the patient into the desired population.

At the moment, our Or gap checking is based on the `ExpressionStack` on each `DataTypeQuery`. We consider gaps to be "or-d" if they share a top-level `ExpressionStackEntry` (aka same library and local ID) and that shared Entry is of type `Or`.

We also re-arranged the Gaps output to match our proposed Grouped Gaps structure, seen in the image below: 
Screen Shot 2021-04-21 at 11 45 10 AM

"or" gaps are represented by a single `DetectedIssue` resource, with multiple `GuidanceResponse`s, indicating that any of the nested `GuidanceResponse`s would satisfy the `DetectedIssue`. "And" gaps are represented by multiple `DetectedIssue` resources, each with a single `GuidanceResponse`. 

This PR will, in theory, support multiple-"or" gaps and "and" gaps at the top level. If there are (for example) two sets of "or-d" gaps, each with two ways to satisfy the Or, you would end up with two `DetectedIssue` resources, each with two `GuidanceResponse` resources nested below them.

We do not yet support nested "and" and "or" gaps.

## New behavior
* Function added to "Group" gaps as "or" gaps where appropriate. Currently very basic, could get more complex in the future.
* Generation of the `ExpressionStack` element on `DataTypeQuery` fixed to not truncate stack entries from parent queries.
* Output of gaps-in-care calculation updated to properly follow the `Representing Grouped Gaps` diagram above. NOTE: This is a breaking change for most measures not named EXM130.

## Code changes
* Added `groupGapQueries` function, which generates an array of arrays of `DataTpeQuery` objects, which will then get turned into the new gap structure
* Updated gap tests to follow the new gaps output structure
* Added `isNearMiss` to the `DataTypeQuery` interface definition.

# Testing guidance
* Run EXM130 using a denominator patient (the "test-denom" patient out of `connectathon` works well) and the measurement period start and end dates of `2019-01-01` and `2019-12-31`. 
* Examine the output Bundle. There should be only one `DetectedIssue` contained in it (because all the missing elements in EXM130 are "or'd"), with several `GuidanceResposne` elements nested as `evidence.detail`
* Run some other measure that doesn't necessarily have OR'd gaps (EXM104 for example). You should see multiple `DetectedIssue` resources in the output `Bundle`.


mgramigna:
## jest coverage report 🧪

### Total coverage

|   🟡   | Statements | 73.85%     |     768/1040    |
|   🔴   | Branches   | 56.36%     |     709/1258    |
|   🟡   | Functions  | 74.15%     |     152/205     |
|   🟡   | Lines      | 73.99%     |     717/969     |

test/fixtures/elm/queries/SimpleQueries.cql
```diff
@@ -1,35 +0,0 @@
-library SimpleQueries version '0.0.1'
```
mgramigna, line 1: was this an intentional nuking of this file?

test/fixtures/elm/queries/SimpleQueriesDependency.cql
```diff
@@ -1,11 +0,0 @@
-library SimpleDep version '0.0.1'
```
mgramigna, line 1: was this an intentional nuking of this file?

tsconfig.json
```diff
@@ -4,6 +4,7 @@
   "compilerOptions": {
     "resolveJsonModule": true,
     "target": "es5",
+    "lib": ["es2019"],
```
mgramigna, line 7: Pardon the ignorance but I guess if we're specifying ES version used in our code, why not ES2020 ;)?

test/fixtures/elm/queries/SimpleQueries.cql
okeefm: nope. Brought it back.

test/fixtures/elm/queries/SimpleQueriesDependency.cql
okeefm: nope. Brought it back.

tsconfig.json
okeefm: I moved to using `ESNext`

mgramigna (dismissed):
My parting gift before heading out.

Merge at your leisure

hossenlopp (approved):
Rebase looks good, coverage of changed code seems to be decent.

PR #58, opened 2021-05-05
Reason code parsing

mgramigna:
# Summary

This PR adds additional parsing of `queryInfo` for Gaps to try to identify the proper `reasonCode`(s) to use in the `GuidanceResponse` resources.

The only supported codes currently are `IncorrectAttribute` (e.g. status was supposed to be "final" but isn't), and `DateOutOfRange` (e.g. procedure needed to be performed sometime in 2020 but wasn't).

These reasonCodes only apply for near misses, as any other gaps will result in code "Missing" or "Present" since the retrieve portion of the whole query will be false in that case.

## New behavior

`GuidanceResponse` output for gaps could now have `IncorrectAttribute` or `DateOutOfRange` reasonCodes

## Code changes

* Update near miss structure to include reason code if applicable
* Update near miss function to handle dates and attributes by checking interval contains and clause results respectively
* Update during filter parsing to return the constructed cql interval back to be able to utilize `contains`
* Add tests for new functionality
* Update some query filter tests to properly handle assertions with the newly returned `interval` from during filter parsing
* ....
* ...
* ...
* Update CI to use GITHUB_TOKEN lol

# Testing guidance

To see the new modeling, run gaps for EXM130 with [this patient](https://github.com/DBCG/connectathon/blob/master/fhir401/bundles/measure/EXM130-7.3.000/EXM130-7.3.000-files/tests-denom-EXM130-bundle.json) who has a procedure but the performed period is outside of the desired range.

Ensure that other measures don't break. (E.g. EXM124 should still produce missing for all its gaps if none of them have an attribute/timing mismatch)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: artiomtr/jest-coverage-report-action@v1.1
         with:
-          github_token: ${{ secrets.ACCESS_TOKEN_CI }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
           threshold: 60
 
   integration_test:
--- a/src/GapsInCareHelpers.ts
+++ b/src/GapsInCareHelpers.ts
@@ -397,6 +397,7 @@ export function calculateNearMisses(
 
             if (duringFilter.valuePeriod.interval && resources) {
               const path = duringFilter.attribute.split('.');
+              const interval = duringFilter.valuePeriod.interval;
 
               // Access desired property of FHIRObject
               resources.raw.forEach((r: any) => {
@@ -410,16 +411,17 @@ export function calculateNearMisses(
                 });
 
                 // Use DateOutOfRange code if data point is outside of the desired interval
-                const isAttrContainedInInterval = duringFilter.valuePeriod.interval?.contains(desiredAttr.value);
+                const isAttrContainedInInterval = interval.contains(desiredAttr.value);
 
                 if (isAttrContainedInInterval === false) {
-                  nearMissInfo.reasonCodes?.push(CareGapReasonCode.DATEOUTOFRANGE);
+                  nearMissInfo.reasonCodes.push(CareGapReasonCode.DATEOUTOFRANGE);
                 }
               });
             }
           } else {
             // TODO: This logic is not perfect, and can be corrupted by multiple resources spanning truthy values for all filters
             // For non-during filters, look up clause result by localId
+            // Ideally we can look to modify cql-execution to help us with this flaw
             const clauseResult = detailedResult.clauseResults?.find(
               cr => cr.libraryName === r.libraryName && cr.localId === f.localId
             );
@@ -428,7 +430,7 @@ export function calculateNearMisses(
             if (clauseResult && clauseResult.final === FinalResult.FALSE) {
               const code = getGapReasonCode(f);
               if (code !== null) {
-                nearMissInfo.reasonCodes?.push(code);
+                nearMissInfo.reasonCodes.push(code);
               }
             }
           }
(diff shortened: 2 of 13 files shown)


mgramigna:
## jest coverage report 🧪

### Total coverage

|   🟡   | Statements | 77.44%     |    1040/1343    |
|   🟡   | Branches   | 61.19%     |     968/1582    |
|   🟡   | Functions  | 77.87%     |     190/244     |
|   🟡   | Lines      | 77.67%     |     981/1263    |

<details><summary>Show files with reduced coverage 🔻</summary>


### Reduced coverage

| Status | Filename             | Statements         | Branches | Functions          | Lines              |
| :----: | :------------------- | :----------------- | :------- | :----------------- | :----------------- |
|   🟢   | GapsInCareHelpers.ts | 90.85% (-2.67% 🔻) | 76.5%    | 93.94% (-2.21% 🔻) | 91.18% (-2.02% 🔻) |

</details>

src/GapsInCareHelpers.ts
```diff
@@ -355,16 +368,96 @@ export function generateGapsInCareBundle(
...
+                  }
+
+                  desiredAttr = desiredAttr[key];
+                });
+
+                // Use DateOutOfRange code if data point is outside of the desired interval
+                const isAttrContainedInInterval = duringFilter.valuePeriod.interval?.contains(desiredAttr.value);
+
+                if (isAttrContainedInInterval === false) {
+                  nearMissInfo.reasonCodes?.push(CareGapReasonCode.DATEOUTOFRANGE);
```
okeefm, line 416: Should we initialize the array here if it hasn't been initialized already, rather than silently failing to add the code?

src/GapsInCareHelpers.ts
```diff
@@ -355,16 +368,96 @@ export function generateGapsInCareBundle(
...
+            // For non-during filters, look up clause result by localId
+            const clauseResult = detailedResult.clauseResults?.find(
+              cr => cr.libraryName === r.libraryName && cr.localId === f.localId
+            );
+
+            // False clause means this specific filter was falsy
+            if (clauseResult && clauseResult.final === FinalResult.FALSE) {
+              const code = getGapReasonCode(f);
+              if (code !== null) {
+                nearMissInfo.reasonCodes?.push(code);
```
okeefm, line 431: Same suggestion here as above

.github/workflows/ci.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: artiomtr/jest-coverage-report-action@v1.1
         with:
-          github_token: ${{ secrets.ACCESS_TOKEN_CI }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
okeefm, line 29: Nice

src/GapsInCareHelpers.ts
mgramigna: Details hashed out in slack. But I went ahead and made reasonCodes a required prop for near misses, which is fine. Is empty array in the bad case.

src/GapsInCareHelpers.ts
mgramigna: Details hashed out in slack. But I went ahead and made reasonCodes a required prop for near misses, which is fine. Is empty array in the bad case.

src/GapsInCareHelpers.ts
```diff
@@ -355,16 +368,96 @@ export function generateGapsInCareBundle(
...
+                path.forEach(key => {
+                  if (desiredAttr.value?.isDateTime) {
+                    return;
+                  }
+
+                  desiredAttr = desiredAttr[key];
+                });
+
+                // Use DateOutOfRange code if data point is outside of the desired interval
+                const isAttrContainedInInterval = duringFilter.valuePeriod.interval?.contains(desiredAttr.value);
```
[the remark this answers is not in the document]
mgramigna: Resolved by instantiating a variable in the proper scope

okeefm (approved):
LGTM :shipit:

PR #70, opened 2021-06-15
Initial DataRequirements function

okeefm:
# Summary
This PR implements a basic `calculateDataRequirements` function in fqm-execution. It only requires a measure `Bundle` object with a `Measure` object, and the required `Library` objects, included.

The `calculateDataRequirements` function returns a `Library` object, to conform to the [Measure/$data-requirements](https://www.hl7.org/fhir/measure-operation-data-requirements.html) operation. 

Currently, we only support providing `resourceType` and `valueSet` filter options for `DataRequirement` objects.

## New behavior
* Added a `calculateDataRequirements` function to `src/Calculator.ts`
* Added a `dataRequirements` output type to `src/cli`

## Code changes
* Moved `MeasureHelpers.ts` to `src/helpers`
* Moved `Library` extraction method out of `calculateGapsInCare` method and into `MeasureHelpers`, for re-use in Data Requirements code
* Created an `extractMeasureFromBundle` function, which returns the `Measure` object from a Measure `Bundle`, with a guaranteed library

# Testing guidance
* Run the tests
* Pick a measure bundle from `connectathon`, and run it using the CLI. Compare the list of `dataRequirements` in the exported `Library` to the one in the `Measure` object, and make sure they're somewhat similar

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ const rawResults = await Calculator.calculateRaw(measureBundle, patientBundles, 
 const detailedResults = await Calculator.calculate(measureBundle, patientBundles, options); // Get detailed population results for each patient
 const measureReports = await Calculator.calculateMeasureReports(measureBundle, patientBundles, options); // Get individual FHIR MeasureReports for each patient
 const gapsInCare = await Calculator.calculateGapsInCare(measureBundle, patientBundles, options); // Get gaps in care for each patient, if present
+const dataRequirements = Calculator.calculateGapsInCare(measureBundle); // Get data requirements for a given measure (in a bundle)
 ```
 
 #### Require
@@ -55,6 +56,7 @@ const rawResults = await Calculator.calculateRaw(measureBundle, patientBundles, 
 const detailedResults = await Calculator.calculate(measureBundle, patientBundles, options); // Get detailed population results for each patient
 const measureReports = await Calculator.calculateMeasureReports(measureBundle, patientBundles, options); // Get individual FHIR MeasureReports for each patient
 const gapsInCare = await Calculator.calculateGapsInCare(measureBundle, patientBundles, options); // Get gaps in care for each patient, if present
+const dataRequirements = Calculator.calculateGapsInCare(measureBundle); // Get data requirements for a given measure (in a bundle)
 ```
 
 #### Calculation Options
@@ -83,7 +85,7 @@ Usage: fqm-execution [options]
 
 Options:
   -d, --debug                                 enable debug output (default: false)
-  -o, --output-type <type>                    type of output, "raw", "detailed", "reports", "gaps" (default: "detailed")
+  -o, --output-type <type>                    type of output, "raw", "detailed", "reports", "gaps", "dataRequirements" (default: "detailed")
   -r, --report-type <type>                    type of MeasureReport (only for output type "reports"): "summary" or "individual" (default: "individual")
   -m, --measure-bundle <measure-bundle>       path to measure bundle
   -p, --patient-bundles <patient-bundles...>  paths to patient bundle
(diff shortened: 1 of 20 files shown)


src/Calculator.ts
```diff
@@ -433,3 +427,54 @@ export async function calculateGapsInCare(
...
+  // get the retrieves for every statement in the root library
+  const allRetrieves = rootLib.library.statements.def.flatMap(statement => {
+    if (statement.expression && statement.name != 'Patient') {
+      const retrieves = RetrievesHelper.findRetrieves(rootLib, elmJSONs, statement.expression);
+      return retrieves;
+    } else {
+      return [];
+    }
+  });
+  const uniqueRetrieves = uniqWith(allRetrieves, isEqual);
```
hossenlopp, line 452: This doesn't seem to properly filter and create unique data requirements. I tested with EXM124 `dbcg-connectathon/fhir401/bundles/measure/EXM124-9.0.000/EXM124-9.0.000-bundle.json` and found this entry twice: 
```
{
  "type": "Observation",
  "codeFilter": [
    {
      "valueSet": "http://cts.nlm.nih.gov/fhir/ValueSet/2.16.840.1.113883.3.464.1003.108.12.1017"
    }
  ]
}
```

hossenlopp (changes requested):
This doesn't check the libraries that are used by the main library. Is this going to be follow on work?

The dataRequirements output do not have the path in which the codeFilters should apply to.

This doesn't properly handle retrieves that use direct reference codes. This can be seen in EXM124 for the retrieve `[Condition : "Congenital absence of cervix (disorder)"]`. The current result has a blank code filter for this.

README.md
```diff
@@ -45,6 +45,7 @@ const rawResults = await Calculator.calculateRaw(measureBundle, patientBundles,
 const detailedResults = await Calculator.calculate(measureBundle, patientBundles, options); // Get detailed population results for each patient
 const measureReports = await Calculator.calculateMeasureReports(measureBundle, patientBundles, options); // Get individual FHIR MeasureReports for each patient
 const gapsInCare = await Calculator.calculateGapsInCare(measureBundle, patientBundles, options); // Get gaps in care for each patient, if present
+const dataRequirements = Calculator.calculateGapsInCare(measureBundle); // Get data requirements for a given measure (in a bundle)
```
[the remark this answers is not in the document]
okeefm: Good catch. Copy-paste fail!

okeefm:
> This doesn't check the libraries that are used by the main library

It will check expressions in other libraries that are referenced by expressions in the main library. For example, if you run dataRequirements against EXM130 with `debug` mode on, you'll see `gaps.json`'s list of retrieves (which is used to create the data requirements array) has retrieves from `EXM130`, `Hospice`, and `AdultOutpatientEncounters`.

I think this behavior is where we landed (getting retrieves for every expression in the root library, including referenced retrieves in the dependent libraries), but if we want something different, I can make those changes too.

okeefm:
> The dataRequirements output do not have the path in which the codeFilters should apply to.

Oops. That's a good thing to have...

okeefm:
> This doesn't properly handle retrieves that use direct reference codes.

Good call, that should be a relatively simple fix.

src/Calculator.ts
okeefm: Ugh. I've been battling with uniq-ing these objects. Guess the fight is still going. Good catch

hossenlopp: I think a custom function is going to be needed here to check for unique retrieves. Another approach would be to run the unique operation on the list of DataRequirements after it's initially transformed from the list of retrieves.

okeefm: Yeah, I think you're correct here.

hossenlopp (approved):
This is looking excellent! It's great that it fixes the codeSystem references to use the proper system urls now.

I do have one thing we should consider for future investigation... We should see if we need to have the `profile` attribute on the DataRequirements that example measure libraries all have.

It looks like this: `"profile": [ "http://hl7.org/fhir/StructureDefinition/Encounter" ]`.

hossenlopp:
It looks like prettier in CI is upset about one of the test files. But when running locally it is not finding the same issue.

hossenlopp (approved):
Prettier fix has been made. This PR looks good now.

PR #72, opened 2021-06-21
Value filter for not null check

lmd59:
# Summary
Creates a value filter for Data requirements that are associated with a not null check in the cql.
## New behavior
The produced gaps report for a patient in the denominator (for a positive improvement measure) will include a value filter (extension) for any corresponding not null checks that are a part of the numerator cql. This data requirement will use the ValueMissing CareGapReasonCode (needs to be added to the gaps in care profile/IG that Sam has been working on).
## Code changes
- Small changes in the GapsInCareHelpers to handle the notnull filter case and use the ValueMissing code. 
- Change to add the ValueMissing code in Enums.
- `generateDetailedValueFilter` added to DataRequirementHelpers to generate the value filter extension. Future value filters can be added to this for other filter types, but for now, it just uses a console warn to warn the user if a different filter type gets there (shouldn't happen yet).
# Testing guidance
A test has been added which can be run with `npm test`. More generally, testing can be done with EXM124. Launch,json configuration would look something like this: 
`"args": ["-d","-o", "gaps","-m", "${workspaceFolder}/test/fixtures/EXM124-9.0.000-bundle.json", "-p", "${workspaceFolder}/test/fixtures/EXM124-9.0.000/tests-denom-EXM124-bundle.json"]`

### DIFF
--- a/src/GapsInCareHelpers.ts
+++ b/src/GapsInCareHelpers.ts
@@ -357,12 +357,12 @@ export function calculateReasonDetail(
             const resources = detailedResult.clauseResults?.find(
               cr => cr.libraryName === r.libraryName && cr.localId === r.retrieveLocalId
             );
-            const attr_path = notNullFilter.attribute.split('.');
+            const attrPath = notNullFilter.attribute.split('.');
             if (resources) {
               // Access desired property of FHIRObject
               resources.raw.forEach((r: any) => {
                 let desiredAttr = r;
-                attr_path.forEach(key => {
+                attrPath.forEach(key => {
                   desiredAttr = desiredAttr[key];
                 });
 
--- a/src/GapsInCareHelpers.ts
+++ b/src/GapsInCareHelpers.ts
@@ -213,12 +213,14 @@ export function generateGuidanceResponses(
           } else {
             dataRequirement.dateFilter = [dateFilter];
           }
-        } else if (df.type === 'notnull') {
+        } else {
           const valueFilter = generateDetailedValueFilter(df);
-          if (dataRequirement.extension) {
-            dataRequirement.extension.push(valueFilter);
-          } else {
-            dataRequirement.extension = [valueFilter];
+          if (valueFilter) {
+            if (dataRequirement.extension) {
+              dataRequirement.extension.push(valueFilter);
+            } else {
+              dataRequirement.extension = [valueFilter];
+            }
           }
         }
       });
--- a/src/GapsInCareHelpers.ts
+++ b/src/GapsInCareHelpers.ts
@@ -14,7 +14,7 @@ import {
   generateDetailedDateFilter,
   generateDetailedValueFilter
 } from './helpers/DataRequirementHelpers';
-import { EqualsFilter, InFilter, DuringFilter, AnyFilter } from './types/QueryFilterTypes';
+import { EqualsFilter, InFilter, DuringFilter, AnyFilter, NotNullFilter } from './types/QueryFilterTypes';
 
 /**
  * Iterate through base queries and add clause results for parent query and retrieve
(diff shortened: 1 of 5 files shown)


src/helpers/DataRequirementHelpers.ts
```diff
@@ -69,6 +77,28 @@ export function generateDetailedDateFilter(filter: DuringFilter): R4.IDataRequir
...
+ * Map a filter into a FHIR DataRequirement valueFilter extension
+ *
+ * @param filter the filter to map
+ * @returns extension for the valueFilter list of dataRequirement
+ */
+export function generateDetailedValueFilter(filter: Filter): R4.IExtension {
+  if (filter.type === 'notnull') {
+    const notnullFilter = filter as NotNullFilter;
+    return {
+      url: 'http://example.com/dr-value',
```
hossenlopp, line 90: Just double checking, these are place holder URL's until we have a defined one in our Profile, right?

src/GapsInCareHelpers.ts
```diff
@@ -344,6 +352,8 @@ export function calculateReasonDetail(
                 }
               });
             }
+          } else if (f.type === 'notnull') {
+            reasonDetail.reasonCodes.push(CareGapReasonCode.VALUEMISSING);
```
hossenlopp, line 356: This doesn't check if the resource is missing a value or not for the specified attribute. I know this may fall into realm of the Reason Detail work that I am starting. The above code on the `detail` filter does deal with checking if the value is in the interval so it may be prudent to do the same type of check here.

hossenlopp (commented):
Just a few comments for now for discussion.

src/GapsInCareHelpers.ts
```diff
@@ -344,6 +352,26 @@ export function calculateReasonDetail(
...
+            const notNullFilter = f as NotNullFilter;
+            const resources = detailedResult.clauseResults?.find(
+              cr => cr.libraryName === r.libraryName && cr.localId === r.retrieveLocalId
+            );
+            const attr_path = notNullFilter.attribute.split('.');
+            if (resources) {
+              // Access desired property of FHIRObject
+              resources.raw.forEach((r: any) => {
+                let desiredAttr = r;
+                attr_path.forEach(key => {
```
mgramigna, line 365:
```suggestion
            const attrPath = notNullFilter.attribute.split('.');
            if (resources) {
              // Access desired property of FHIRObject
              resources.raw.forEach((r: any) => {
                let desiredAttr = r;
                attrPath.forEach(key => {
```

sorry

lmd59: ugh- this was a quick change because vscode was doing something *dumb* with debugging, and I forgot to change it back. Sorry my rubyisms are trying to infect the js.

src/GapsInCareHelpers.ts
```diff
@@ -212,6 +213,13 @@ export function generateGuidanceResponses(
           } else {
             dataRequirement.dateFilter = [dateFilter];
           }
+        } else if (df.type === 'notnull') {
+          const valueFilter = generateDetailedValueFilter(df);
+          if (dataRequirement.extension) {
+            dataRequirement.extension.push(valueFilter);
+          } else {
+            dataRequirement.extension = [valueFilter];
+          }
```
hossenlopp, line 222: This whole chunk of code isn't covered by unit tests. Perhaps we could create a set of tests real quick.

src/helpers/DataRequirementHelpers.ts
```diff
@@ -69,6 +77,28 @@ export function generateDetailedDateFilter(filter: DuringFilter): R4.IDataRequir
...
+      extension: [
+        { url: 'dr-value-attribute', valueString: notnullFilter.attribute },
+        { url: 'dr-value-filter', valueString: 'not null' }
+      ]
+    };
+  } else {
+    console.warn(`Detailed value filter is not yet supported for filter type ${filter.type}`);
```
hossenlopp, line 97: A quick test could be added to throw something unexpected at this function to increase coverage and make sure it has an appropriate "fail" outcome.

src/helpers/DataRequirementHelpers.ts
```diff
@@ -69,6 +77,28 @@ export function generateDetailedDateFilter(filter: DuringFilter): R4.IDataRequir
...
+    return {};
```
hossenlopp, line 98: 🤔 Would it be better to return `null` and check the output of this function to make sure a result came back before putting it in the report? This would make sure we don't end up putting an empty object where there should be a complete extension.

hossenlopp (changes requested):
Just a few more tests to get more code covered and one last actual change idea. Sorry I didn't notice this in my first review.

hossenlopp (approved):
Pair reviewed and cleaned up on call. Good to go!

PR #74, opened 2021-07-07
Additional MAT export support

mgramigna:
# Summary

This PR does 2 main things in order to further support MAT bundles:

1. Allow an alternate code system for searching for the measure scoring code (both `http://hl7.org/fhir/measure-scoring` and 
`http://terminology.hl7.org/CodeSystem/measure-scoring` are now supported)
2. Add a fallback in the `MeasureReportBuilder` where a groupId not being present on the report falls back to accessing the groups by index. Some of the MAT Measures don't include groupIds, so it's impossible to persist those over to the MeasureReport. This was causing the builder to break for MAT bundle. It doesn't now :) 

## New behavior

Main behavior change is measure report output for MAT bundles no longer erroring

## Code changes

* Add additional code system to measure scoring lookup (thanks @hossenlopp)
* Add logic for falling back to measure report indexed group when groupId doesn't exist.

# Testing guidance

I recommend testing this with a MAT bundle that has valuesets jammed into it just to isolate the changed behavior rather than rely on VSAC, esp because that behavior may change with caching. Let me know and I can provide such a bundle for testing.

Test all output types for the bundle and ensure that they all calculate without errors.

### DIFF
--- a/src/Calculator.ts
+++ b/src/Calculator.ts
@@ -312,8 +312,11 @@ export async function calculateGapsInCare(
       const measureResource = MeasureHelpers.extractMeasureFromBundle(measureBundle);
 
       // Gaps only supported for proportion/ratio measures
-      const scoringCode = measureResource.scoring?.coding?.find(c => c.system === 'http://hl7.org/fhir/measure-scoring')
-        ?.code;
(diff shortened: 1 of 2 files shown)


okeefm (approved):
Tested, all output types work. lgtm

PR #75, opened 2021-07-09
ValueSet Caching

mgramigna:
# Summary

Add ability to cache ValueSets obtained via API calls. When the option is enabled, the CLI will write ValueSets to a gitignored directory after calculation succeeds.

## New behavior

* If the calculation engine encounters a scenario where valuesets are passed in to the calculation functions, it will include those valuesets in the code service used for cql execution
* If `useValueSetCaching` is specified as a calculation option, any new ValueSets retrieved from an API will be cached in memory until the results are returned to the CLI, where it will write them to disk in a `cache/terminology` directory (I went with a subdirectory in case we ever want to store something other than ValueSets in a cache

## Code changes

* New CLI option `-c/--cache-valuesets` which, if enabled, will update a gitignored directory `cache/terminology` on the users' machine with any valuesets retrieved via API requests (valuesets present in the provided measure bundle are not cached)
* All Calculator functions now have a new argument `valueSetCache` which can either be directly provided if calling the functions (e.g. in a service or other application), or in this case, is auto-generated by reading the contents of `cache/terminology`. This also makes the task to include an optional valueSet argument in calculate functions OBE
* ValueSet resolver code
  * Normalize all VSAC URLs to use the r4 base (we were getting back some stu3 valuesets somehow)
  * Include the provided valueset cache to check for missing valuesets before querying the API
* Add helper for VSAC specific resolutions and unit tests
* My editor formatted the README using prettier so instead of undoing that I just committed it idk

# Testing guidance

Testing might be annoying, but I'd recommend the following steps:

1. Run the CLI with a measure bundle where a valueset is missing (e.g. `test/fixtures/EXM130-7.3.000-bundle-nocodes-missingVS.json`), and make sure you provide an api key and use caching (`-a` and `-c` respectively)
2. The code will notice that there is a valueset missing, and will ping VSAC for it
3. After the results are output, there should be a new directory on your machine `cache/terminology` that contains that missing valueset
4. Run the CLI again with the exact same inputs, and it should be a little faster this time (that is because it is using the cached valueset rather than reaching out to VSAC). If you don't believe me, add a console log before GET requests are sent to VSAC to ensure that they aren't happening in this case

I recommend repeating this behavior for all output types except data requirements just to ensure that the argument passing is correct

### DIFF
--- a/src/Calculator.ts
+++ b/src/Calculator.ts
@@ -22,18 +22,21 @@ import { generateDataRequirement } from './helpers/DataRequirementHelpers';
  * @param measureBundle Bundle with a MeasureResource and all necessary data for execution.
  * @param patientBundles List of bundles of patients to be executed.
  * @param options Options for calculation.
+ * @param valueSetCache Cache of existing valuesets
  * @returns Detailed execution results. One for each patient.
  */
 export async function calculate(
   measureBundle: R4.IBundle,
   patientBundles: R4.IBundle[],
-  options: CalculationOptions
+  options: CalculationOptions,
+  valueSetCache: R4.IValueSet[] = []
 ): Promise<{
   results: ExecutionResult[];
   debugOutput?: DebugOutput;
   elmLibraries?: ELM[];
   mainLibraryName?: string;
   parameters?: { [key: string]: any };
+  valueSetCache?: R4.IValueSet[];
 }> {
   const debugObject: DebugOutput | undefined = options.enableDebugOutput ? <DebugOutput>{} : undefined;
 
@@ -49,7 +52,7 @@ export async function calculate(
   const measure = MeasureHelpers.extractMeasureFromBundle(measureBundle);
   const executionResults: ExecutionResult[] = [];
 
-  const results = await Execution.execute(measureBundle, patientBundles, options, debugObject);
+  const results = await Execution.execute(measureBundle, patientBundles, options, valueSetCache, debugObject);
   if (!results.rawResults) {
     throw new Error(results.errorMessage ?? 'something happened with no error message');
   }
@@ -167,10 +170,15 @@ export async function calculate(
       debugOutput: debugObject,
       elmLibraries: results.elmLibraries,
       mainLibraryName: results.mainLibraryName,
-      parameters: results.parameters
+      parameters: results.parameters,
+      ...(options.useValueSetCaching && results.valueSetCache && { valueSetCache: results.valueSetCache })
     };
   } else {
-    return { results: executionResults, debugOutput: debugObject };
+    return {
+      results: executionResults,
+      debugOutput: debugObject,
+      ...(options.useValueSetCaching && results.valueSetCache && { valueSetCache: results.valueSetCache })
+    };
   }
 }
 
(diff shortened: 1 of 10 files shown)


src/helpers/ValueSetHelper.ts
```diff
@@ -132,15 +133,15 @@ export function getMissingDependentValuesets(measureBundle: R4.IBundle): string[
   // unique-ify
   const uniqueVS = vsURLs.filter((value, index, self) => self.indexOf(value) === index);
 
+  // full array of all valueset URLs present across measureBundle and cache
+  const existingValueSets = measureBundle.entry
+    .filter(e => e.resource?.resourceType === 'ValueSet')
+    .map(e => e.resource as R4.IValueSet)
+    .concat(valueSetCache)
+    .map(vs => vs.url as string);
```
mgramigna, line 141: what does this do

mgramigna: filters stuff

mgramigna: thanks

src/Calculator.ts
```diff
@@ -268,35 +288,54 @@ export async function calculateAggregateMeasureReport(
...
+ * @returns pass through of raw calculation results from the engine
+ */
 export async function calculateRaw(
   measureBundle: R4.IBundle,
   patientBundles: R4.IBundle[],
-  options: CalculationOptions
-): Promise<{ results: cql.Results | string; debugOutput?: DebugOutput }> {
+  options: CalculationOptions,
+  valueSetCache: R4.IValueSet[] = []
+): Promise<{ results: cql.Results | string; debugOutput?: DebugOutput; valueSetCache?: R4.IValueSet[] }> {
```
sarahmcdougall, line 308: Still trying to learn the code, could you explain what this does?

src/helpers/VSACHelper.ts
```diff
@@ -0,0 +1,28 @@
+const VSAC_REXEG = /http:\/\/cts\.nlm\.nih\.gov.*ValueSet/;
```
natjoe4, line 1: is REXEG instead of REGEX intentional? What is REXEG?

mgramigna: incredible

hossenlopp (commented):
Does the job it promises to do. Doesn't break in the fqm-execution-demo.

src/cli.ts
```diff
@@ -42,28 +42,38 @@ program
...
+  return [];
+}
+
 async function calc(
   measureBundle: R4.IBundle,
   patientBundles: R4.IBundle[],
-  calcOptions: CalculationOptions
+  calcOptions: CalculationOptions,
+  valueSetCache: R4.IValueSet[] = []
 ): Promise<any> {
```
natjoe4, line 66: What's the point of defining the return object to be of type <any>? Is it because the different report types differ too greatly for it to make sense to define an interface that encompasses all of them?

mgramigna: Short answer: this is laziness on our part.

Usually `any` is really only supposed to be used when either the result of something could really be anything, or the type is not really strictly defined by whatever you're using to get the results. Really, this type should be a concatenation of all the possible return types defined by the functions exported in calculator. As you've pointed out before, a lot of those return types are defined inline in the function declaration. 

I feel like it makes sense to abstract everything to interfaces, and then we could really easily make stricter typing here. Sounds like a good task for people trying to get more familiar with Typescript 👀

src/Calculator.ts
hossenlopp: This function is the main entry point for simply running the measure logic against a list of patients and returning the raw results from the `cql-execution` engine. These raw results don't have any processing on them to relate the results to the populations defined in the measure.

Pretty much all of the work that this function returns is in done in the `execute` function in `Execution.ts`. This function just formats it correctly. In this PR it passes back the valueSetCache if one was created and filled with any valuesets pulled from VSAC.

natjoe4 (approved):
Passed all my checks. Thanks for the explanations!

PR #76, opened 2021-07-13
Codesystem lookup

natjoe4:
# Summary
Added a codeLookup function inside the GenerateDetailedCodeFilters function which takes as arguments a string representing a FHIR dataType and an attribute of that dataType and returns the url of the codesystem that outlines the valid settings of said attribute when appropriate (returns undefined otherwise). Updated GenerateDetailedCodeFilters to call codeLookup when passed an appropriate dataType, attribute pair and add the codesystem url to the output code object.

## New behavior
adds the url of the codesystem to appropriate code objects in JSON output

## Code changes
GenerateDetailedCodeFilters now takes as input and additional string representing a FHIR dataType 

# Testing guidance
Tests will need to be updated to verify the new functionality is working properly and old tests will need to be altered to call the updated GenerateDetailedCodeFilters function which now takes an additional dataType: string as the second argument.

### DIFF
--- a/src/helpers/DataRequirementHelpers.ts
+++ b/src/helpers/DataRequirementHelpers.ts
@@ -10,8 +10,6 @@ import {
   NotNullFilter
 } from '../types/QueryFilterTypes';
 
-
-
 /**
  * Take any nesting of base filters and AND filters and flatten into one list
  *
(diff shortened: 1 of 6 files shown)


mgramigna:
Some unit testing guidance:

1) Move `codeLookup` to its own exported function within `DataRequirementsHelpers` so it can be unit tested

2) Write tests that provide known good (or bad) inputs to the codeLookup function and assert that the results are what you expect. Try to cover all the branches that exist (i.e. all the if/else/switch cases)

3) There are some broken tests rn for the `genereateDetailedCodeFilter` function because it now requires a second argument. I'd add that argument to the jest tests and make sure you can get the existing tests passing in addition to the new ones

src/helpers/DataRequirementHelpers.ts
```diff
@@ -35,22 +37,82 @@ export function flattenFilters(filter: AnyFilter): AnyFilter[] {
  * @param filter the filter to translate
  * @returns codeFilter to be put on the DataRequirement
  */
-export function generateDetailedCodeFilter(filter: EqualsFilter | InFilter): R4.IDataRequirement_CodeFilter | null {
+export function generateDetailedCodeFilter(filter: EqualsFilter | InFilter, dataType: string): R4.IDataRequirement_CodeFilter | null {
```
mgramigna, line 40: Suggest making the `dataType` argument optional here, that way some of the existing tests should hopefully pass as is when this is omitted.

This will require you to change the logic a bit when you actually call the `codeLookup` function, as you only want to do this when `dataType` is defined. 

Best approach for that is probably to initialize the `system` variable to be nothing first, and then only set its value inside an `if` for `if(dataType)`

src/helpers/DataRequirementHelpers.ts
```diff
@@ -35,22 +37,82 @@ export function flattenFilters(filter: AnyFilter): AnyFilter[] {
...
+          return "http://hl7.org/fhir/request-priority";
+          
+        default:
+          return undefined;
+      }
+    }
+    return undefined;
+    
+  }
+  let system : string | undefined = codeLookup(dataType, filter.attribute);
```
mgramigna: This thing is a small nitpick, so I might defer to whatever you prefer. BUT

In my opinion, it's better to use `null` over `undefined` to indicate when something has "No value" versus the concept of it "not existing." Obviously these two things are very similar, but I recommend using `null` instead of `undefined` here. This would require switching up the return values and types, but it shouldn't affect much else

mgramigna (changes requested):
Some inline comments along with the tests that we already discussed

test/DataRequirementHelpers.test.ts
```diff
@@ -9,6 +9,7 @@ import {
 } from '../src/types/QueryFilterTypes';
 import { R4 } from '@ahryman40k/ts-fhir-types';
 import { DataTypeQuery } from '../src/types/Calculator';
+import exp from 'constants';
```
mgramigna, line 12:
```suggestion
```

mgramigna (approved):
yay

PR #78, opened 2021-07-16
Created interfaces for Calculator function return types

sarahmcdougall:
# Summary
Add interfaces to the Calculator function return types so that we have stricter typing across the entire application.
 
## New behavior
Behavior should remain unchanged, but the application will have stricter typing due to these changes for the Calculator functions.

## Code changes
- New interfaces added to `src/types/Calculator.ts`, including a parent interface `CalculatorFunctionOutput` for the 'results' and 'debugOutput`. 
- The interfaces `MRCalculationOutput`, `RCalculationOutput`, `GICCalculationOutput`, and `DRCalculationOutput` extend from the parent interface.
- The interfaces `AMRCalculationOutput` and `IMRCalculationOutput` extend from `MRCalculationOutput`, as they pertain to measure reports.
- Function outputs were adjusted in `src/cli.ts` from `any` type to stricter typing.

# Testing guidance
Functionality should be unchanged. Run standard tests and look over for typescript errors. Review that the interface names make sense (are they too long? confusing? not allowed to have acronyms?)

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -12,7 +12,7 @@ import {
   calculateDataRequirements
 } from './Calculator';
 import { clearDebugFolder, dumpCQLs, dumpELMJSONs, dumpHTMLs, dumpObject, dumpVSMap } from './DebugHelper';
-import { CalculationOptions } from './types/Calculator';
+import { CalculationOptions, CalculatorFunctionOutput } from './types/Calculator';
 
 program
   .option('-d, --debug', 'enable debug output', false)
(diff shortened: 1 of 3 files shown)


src/cli.ts
```diff
@@ -78,6 +78,9 @@ async function calc(
     // CalculateDataRequirements doesn't make use of the calcOptions object at this point
     result = calculateDataRequirements(measureBundle);
   }
+  if (!result) {
+    throw new Error('Result is not defined');
```
mgramigna, line 82:
```suggestion
    throw new Error(`Could not obtain result based on outputType ${program.outputType}`);
```

Suggest rephrasing this slightly to more clearly indicate what's happening.  Technically this should only happen when none of these other `if`s get processed, which would occur if we are provided an output type that we can't understand (this really can't even happen but the error message is still valuable just incase).

mgramigna (approved):
Nice!

PR #80, opened 2021-07-22
Nested query in library

natjoe4:
# Summary
Added support for nested queries which reference a query in another library. Added libraryName attribute to QueryInfo type and filters.

## New behavior
Queries which reference another query in a different library can now be handled by fqm-execution. Previously, nested queries were only supported if defined in the same library. 

Objects in the retrieves output array should now have attributes retrieveLibraryName and queryLibraryName instead of attribute libraryName. 

QueryInfo and Filter objects which have a localId attribute should now all have a libraryName attribute to accompany it.

## Code changes
parseQueryInfo and findRetrieves now take an additional param allElm: ELM[]
Main changes can be found in QueryFilterHelpers.ts and RetrievesHelpers.ts

# Testing guidance
Added and edited tests to ensure support for nested query referencing another query which can be run using npm run test
Run EXM-130 or any other Measure + Patient bundle combo which is appropriate for gaps in care testing and ensure that the output is unchanged. As of right now there is not a Measure bundle which causes fqm-execution to handle a nested query in another library, so the best method is to ensure all other functionality remains unchanged and that the added unit tests are correct and passing. New tests are in RetrievesHelper.test.ts ('query is further filtered by another query from another library') and parseQueryInfo.test.ts ('simple - query references query in another library, combines filters'). Also ensure that all output of parseQueryInfo is unchanged besides adding correct libraryNames where localIds are found. I accomplished this with some console.logs to print the output of parseQueryInfo while running EXM130 bundles for gaps in care report. There may also be a way to do this with the -d flag.

### DIFF
--- a/src/GapsInCareHelpers.ts
+++ b/src/GapsInCareHelpers.ts
@@ -30,7 +30,7 @@ export function processQueriesForGaps(
   return queries.map(q => {
     // Determine satisfaction of parent query and leaf node retrieve
     const parentQueryResult = detailedResult.clauseResults?.find(
-      cr => cr.libraryName === q.retrieveLibraryName && cr.localId === q.queryLocalId
+      cr => cr.libraryName === q.queryLibraryName && cr.localId === q.queryLocalId
     );
 
     const retrieveResult = detailedResult.clauseResults?.find(
--- a/src/GapsInCareHelpers.ts
+++ b/src/GapsInCareHelpers.ts
@@ -431,10 +431,8 @@ export function calculateReasonDetail(
                       path: duringFilter.attribute,
                       reference: `${resource._json.resourceType}/${resource.id.value}`
                     });
-
                   }
                 }
-
               } else if (f.type === 'notnull') {
                 const notNullFilter = f as NotNullFilter;
                 const attrPath = notNullFilter.attribute.split('.');
@@ -460,7 +458,7 @@ export function calculateReasonDetail(
                 // For non-during filters, look up clause result by localId
                 // Ideally we can look to modify cql-execution to help us with this flaw
                 const clauseResult = detailedResult.clauseResults?.find(
-                  cr => cr.libraryName === r.libraryName && cr.localId === f.localId
+                  cr => cr.libraryName === r.retrieveLibraryName && cr.localId === f.localId
                 );
 
                 // False clause means this specific filter was falsy
--- a/src/GapsInCareHelpers.ts
+++ b/src/GapsInCareHelpers.ts
@@ -30,11 +30,11 @@ export function processQueriesForGaps(
   return queries.map(q => {
     // Determine satisfaction of parent query and leaf node retrieve
     const parentQueryResult = detailedResult.clauseResults?.find(
-      cr => cr.libraryName === q.libraryName && cr.localId === q.queryLocalId
+      cr => cr.libraryName === q.retrieveLibraryName && cr.localId === q.queryLocalId
     );
 
     const retrieveResult = detailedResult.clauseResults?.find(
-      cr => cr.libraryName === q.libraryName && cr.localId === q.retrieveLocalId
+      cr => cr.libraryName === q.retrieveLibraryName && cr.localId === q.retrieveLocalId
     );
 
     const parentQueryHasResult = parentQueryResult?.final === FinalResult.TRUE;
@@ -348,7 +348,7 @@ export function calculateReasonDetail(
           if (f.type === 'during') {
             const duringFilter = f as DuringFilter;
             const resources = detailedResult.clauseResults?.find(
-              cr => cr.libraryName === r.libraryName && cr.localId === r.retrieveLocalId
+              cr => cr.libraryName === r.retrieveLibraryName && cr.localId === r.retrieveLocalId
             );
 
             if (duringFilter.valuePeriod.interval && resources) {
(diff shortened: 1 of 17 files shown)


src/QueryFilterHelpers.ts
```diff
@@ -60,6 +61,7 @@ import { findClauseInLibrary } from './helpers/ELMHelpers';
  */
 export function parseQueryInfo(
   library: ELM,
+  allELM: ELM[],
```
sarahmcdougall, line 64: Minor comment here - could you update the `parseQueryInfo()` docstring so that it includes a `@param` tag for `allElm`? This would make it easier to understand what `allElm` represents.

natjoe4: Good catch!

sarahmcdougall (approved):

src/QueryFilterHelpers.ts
```diff
@@ -60,10 +61,12 @@ import { findClauseInLibrary } from './helpers/ELMHelpers';
...
   queryLocalId: string | undefined,
   parameters: { [key: string]: any } = {},
   patient: R4.IPatient
 ): QueryInfo {
+  console.log('reached');
```
mgramigna, line 69:
```suggestion
```

The perils of not doing `git add -p` ;)

src/QueryFilterHelpers.ts
```diff
@@ -207,38 +218,58 @@ export function interpretExpression(
...
+      break;
     default:
       console.error(`Don't know how to parse ${expression.type} expression.`);
       // Look for a property (source attribute) usage in the expression tree. This can denote an
       // attribute on a resource was checked but we don't know what it was checked for.
       const propUsage = findPropertyUsage(expression, expression.localId);
+
       if (propUsage) {
         return propUsage;
       }
```
mgramigna, line 258: Open question: I see you've changed the above cases to set the value of a variable, rather than return it.

Do we want this behavior here to? Or is the default a special case here where we'll always want to return immediately rather than do the logic at the bottom of this function. @hossenlopp might know.

src/helpers/RetrievesHelper.ts
```diff
@@ -46,26 +46,38 @@ export function findRetrieves(
...
         bottomExprs[0].type === 'Query' &&
         bottomExprs[1].type === 'ExpressionRef' &&
         bottomExprs[2].type === 'Query'
       ) {
         // check if the outer query is indeed referencing the inner one in the source.
-        const outerQuery = findClauseInLibrary(elm, bottomExprs[0].localId) as ELMQuery;
+        /**
+         * seems like here we will need to replace elm with the ELM object of
+         * the referenced library when necessary
+         */
```
mgramigna, line 68: Is this comment still necessary? If so, could it be condensed to a one-liner?

test/ELMDependencyHelper.test.ts
```diff
@@ -130,6 +130,9 @@ describe('ELMDependencyHelper', () => {
         accessLevel: 'Public'
       };
       const vs = ELMDependencyHelper.findValueSetReference(elm, [elm, dependency], valueSetRef);
+      if (vs) {
+        vs.annotation = undefined;
+      }
```
mgramigna, line 135: Is the purpose here just to get rid of the annotation property on the valueset? If so, `delete vs.annotation` is a way of doing that that makes the goal a little clearer

mgramigna (dismissed):
See inline comments. Looks good overall

src/GapsInCareHelpers.ts
```diff
@@ -30,11 +30,11 @@ export function processQueriesForGaps(
   return queries.map(q => {
     // Determine satisfaction of parent query and leaf node retrieve
     const parentQueryResult = detailedResult.clauseResults?.find(
-      cr => cr.libraryName === q.libraryName && cr.localId === q.queryLocalId
+      cr => cr.libraryName === q.retrieveLibraryName && cr.localId === q.queryLocalId
```
hossenlopp, line 33:
```suggestion
      cr => cr.libraryName === q.queryLibraryName && cr.localId === q.queryLocalId
```
This one wants the query results.

src/QueryFilterHelpers.ts
hossenlopp: I don't understand why the changes above were made. It makes sense to return immediately. My worry is that that malformed filter with no `type` could be returned if it isn't replaced with a valid filter.

src/QueryFilterHelpers.ts
```diff
@@ -207,38 +217,55 @@ export function interpretExpression(
   parameters: any,
   patient: R4.IPatient
 ): AnyFilter {
+  let returnFilter: AnyFilter = {
+    type: ''
```
hossenlopp, line 221: I highly suggest making this default to be an UnknownFilter.
```suggestion
    type: 'unknown'
```

src/QueryFilterHelpers.ts
```diff
@@ -207,38 +217,55 @@ export function interpretExpression(
...
       // attribute on a resource was checked but we don't know what it was checked for.
       const propUsage = findPropertyUsage(expression, expression.localId);
+
       if (propUsage) {
         return propUsage;
       }
+      return {
+        type: 'unknown',
+        localId: expression.localId
+      };
```
hossenlopp, line 260: Actually tacking on to the above comment. This could be done above when the `returnFilter` is filled out before the switch statement.

src/QueryFilterHelpers.ts
```diff
@@ -207,38 +217,55 @@ export function interpretExpression(
...
   }
   // If we cannot make sense of this expression or find a parameter usage in it, then we should return
   // an UnknownFilter to denote something is done here that we could not interpret.
-  return {
-    type: 'unknown',
-    localId: expression.localId
-  };
+
+  if (expression.type !== 'And' && expression.type !== 'Or') {
```
hossenlopp, line 265: This if statement isn't needed. We DO want the libraryName for the filters created here.

hossenlopp (approved):
This looks good!

PR #81, opened 2021-07-26
restructured src directory

natjoe4:
# Summary
Restructured the src directory for better organization by function
Most of the functional modules and helpers fit well into subdirectories based on their file references. There are still a few helper modules which did not fit cleanly into one of these subdirectories, so I created the "universal_helpers" subdirectory to house them. I had more difficulty splitting up the types directory into these subdirectories, as I found many different modules reference the type files. Any guidance for how to split up the types and universal helpers file, or changes to the subdirectories I created is greatly appreciated.
## New behavior
No new behavior
## Code changes
Updated all necessary import statements
Restructured directory
# Testing guidance
use npm run test or any measure + patient bundle and report type through the cli to ensure no import path errors. All functionality should remain the same.


sarahmcdougall (commented):
Overall, I think these are good changes and the new restructuring makes it easier for me to find the files I need. I think other team members will have different opinions because they are used to the existing structure, so they may know how to better restructure the directory.

I'm not really sure how to split up the types because I'm still getting used to fqm-execution. As for the universal helpers, would it be okay to not group them into a subdirectory? I think people may forget that they are under the universal helpers subdirectory - it doesn't seem super intuitive to me. It is hard for me to determine similarities between the files, aside from the fact that they all do not fit cleanly somewhere else.

mgramigna:
Reviewing in a bit, but posting the actual file structure here of the initial pass for posterity:

image

mgramigna (changes requested):
This looks good overall and introduces a lot of improvements. I have a few comments:

* I think we should make more use of subdirectories here, specifically with `helpers` It's better to have one base `helpers` directory (which can contain the universal helpers at the root), and then subdirectories within that to identify the kinds of specific helpers for other files, like so:

```
├── helpers
│   ├── DataRequirementHelpers.ts
│   ├── DebugHelper.ts
│   ├── MeasureHelpers.ts
│   ├── elm
│   │   ├── ELMDependencyHelper.ts
│   │   └── ELMHelpers.ts
│   └── gaps
│       ├── GapsInCareHelpers.ts
│       ├── QueryFilterHelpers.ts
│       └── RetrievesHelper.ts
```
This should eliminate the need for `elm_helpers`, `gaps_in_care` and `universal_helpers`

* Using `snake_case` for file/directory naming in Javascript project typically isn't the convention. There are a few other ways you could name them, but I honestly think every directory in this case can pretty much be shortened to one word. The above bullet shortens every helper to one word anyway, this is more so just an observation
* I think it probably makes sense to include MeasureReportBuilder.ts within the calculation function since that is interpreting execution results to assemble some output, which seems to be the purpose of that directory

sarahmcdougall (approved):
Everything looks good to me! The renames and restructuring make more sense now, and all the functionality seems to be the same.

okeefm (changes requested):
In general this looks good to me! I think we just have to update the examples in the `README` to change the import paths.

mgramigna (approved):
Looks good to me! Unsure where in the README would need updates, but @okeefm might be seeing something that I'm not

okeefm (approved):

PR #82, opened 2021-07-27
Reason Detail Extension with Resource References

hossenlopp:
# Summary
Captures the references to the resources that did not meet the requirements listed in the guidance response and includes that information in the extensions on the `reasonCode.coding`.

## New behavior
Gaps in care GuidanceResponses now have extensions to include the reference and path to the problem attribute that was the reason for the guidance.

## Code changes
- Reworked some of the reason detail calculation to not iterate over resources multiple times.
- Fixed gaps tests to use "real" `cql-exec-fhir` objects to match what data the real code will experience
- Added to `GapsDataTypeQuery` interface nested types to have space for collecting reference and path on the ReasonDetail.
- GuidanceResponse resource building now adds the extensions defined in the draft IG to include the resource reference and path in the reasonCode.

# Testing guidance
Run gaps on EXM130 from c-thon with denom test patient. Look for the `reasonCode` with "DateOutOfRange" in the GuidanceResponses. It should reference `Procedure/denom-EXM130-2` with path `performed.end`. Also test other measures with gaps output.

### DIFF
--- a/src/GapsInCareHelpers.ts
+++ b/src/GapsInCareHelpers.ts
@@ -182,11 +182,11 @@ export function generateGuidanceResponses(
     let gapCoding: R4.ICoding[];
 
     // TODO: update system to be full URL once defined
-    if (q.reasonDetail?.hasReasonDetail && q.reasonDetail.reasonCodes) {
-      gapCoding = q.reasonDetail.reasonCodes.map(c => ({
+    if (q.reasonDetail?.hasReasonDetail && q.reasonDetail.reasons.length > 0) {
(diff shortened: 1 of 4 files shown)


lmd59 (changes requested):
Everything looks pretty darn good to me. Only complaints are about some comment weirdness, which should be quick fixes.

lmd59 (approved):

natjoe4:
I was able to produce Hoss's described results for the described bundles in testing guidance. Wasn't able to find any clear logic errors in the code. Good work!

PR #83, opened 2021-08-04
Graceful fail

natjoe4:
# Summary
Added more graceful error handling in a few situations that have arisen with recent bundles. Now throws a custom error if the uploaded patient bundle has no entries. Also converted console.warn statements to instead be returned as GracefulError objects and bubble up the call chain. This will later allow fqm-execution-service to show more detailed and user-friendly error messages, rather than simply printing warn statements to the console.
## New behavior
A custom error should now be thrown when uploading a patient bundle without any entries
detectedIssues should now return inside of an object of the form {
    detectedIssues: detectedIssues, 
    withErrors: GracefulError[]
}
## Code changes
Added new GracefulError type to new file GracefulError.ts
Updated types in calculator and QueryFilterTypes to include a GracefulError array (withErrors) as an optional attribute
Updated calculateGapsInCare() in calculator.ts
- Now returns an errorLog under attribute withErrors when appropriate
Updated execute() in execution.ts
- now checks if patient bundle has no entries before calculation and returns an error if so
Updated generateDetectedIssueResources() in GapsReportBuilder
- now returns an object with a withErrors: GraceulError[] attribute
- reduces all encountered warnings into a single arrray
- updated generateGuidanceResponses similarly
Updated interpretNot in QueryFilterParser.ts
- now returns error info instead of calling console.warn
# Testing guidance
Ensure all previous functionality is unaffected. In order to test the new console.warn handling, add 

"args": ["-d","-o", "gaps","-m", "${workspaceFolder}/test/fixtures/ColorectalCancerScreeningCQM-bundle.json", "-p", "${workspaceFolder}/test/fixtures/EXM130-Bryn/tests-denom-EXM130-bundle.json"]

to your launch.json file and comment out all other args attributes. Either move the needed patient and measure bundles to the correct directories or update their file paths in the command, then run the debugger. If you do not have these bundles please contact me. Ensure that the output object contains a withErrors attribute that contains 5 GracefulError all with the message "could not handle 'isNull' inside expression type not..."

add
"args": ["-d","-o", "reports","-m", "${workspaceFolder}/test/fixtures/connectathon/EXM104-8.2.000-bundle.json", "-p", "${workspaceFolder}/test/fixtures/connectathon/tests-denomexcp-EXM104-bundle.json", "-s", "20190101", "-e", "20191231"]

to the launch.json file to check new error handling for patient bundle with no entries. Ensure this fails with the following error message: "No entries found in passed patient bundles"


test/queryFilters/interpretGreaterOrEqual.test.ts
```diff
@@ -379,7 +382,8 @@ describe('interpretGreaterOrEqual', () => {
     ) as UnknownFilter;
 
     expect(filter).toEqual({
-      type: 'unknown'
+      type: 'unknown',
+      withError: { message: 'Function referenced is not "CalendarAgeInYearsAt"' }
     });
```
natjoe4, line 387: Can someone double check these last 3 error messages to ensure that they are appropriate for the given situation.

src/gaps/QueryFilterParser.ts
```diff
@@ -377,7 +382,7 @@ export function interpretFunctionRef(functionRef: ELMFunctionRef, library: ELM):
           break;
       }
     } else {
-      console.warn(`do not know how to interpret function ref ${functionRef.libraryName}."${functionRef.name}"`);
+      //(`do not know how to interpret function ref ${functionRef.libraryName}."${functionRef.name}"`);
```
mgramigna, line 385: was this commented out line supposed to be another withError aggregation that was missed?

src/types/GracefulError.ts
```diff
@@ -0,0 +1,12 @@
...
+  type?: string;
+  stackTrace?: string;
+}
+
+export const isOfTypeGracefulError = (tbd: any): tbd is GracefulError => {
+  if ((tbd as GracefulError).message) {
+    return true;
+  }
+  return false;
+};
```
mgramigna, line 12: I wonder if this type guard can be more specific. Technically this will return true if any object with a `message` property is passed in. E.g. if I pass in `{ message: 'hello', disclaimer: 'this is not a graceful error but tricks the type guard' }`, it will return true.

I don't think this is a huge deal right now. I'm trying to think of a way to update it that doesn't require too much code, but it would probably require some piece of identification on the interface that we could check. Unfortunately that means everywhere we make a graceful error, we'd have to add a new property. That seems like too much work for now, so I'm fine leaving this. Just worth noting for the future

src/types/QueryFilterTypes.ts
```diff
@@ -121,3 +124,9 @@ export interface UnknownFilter extends Filter {
 export interface TautologyFilter extends Filter {
   type: 'truth';
 }
+
+export interface CQLInterval {
```
mgramigna, line 128: Might suggest renaming this to not have it appear that we are defining a specific CQL type. Technically, this is a construct we made from the results of QueryFilterParsing, so it slightly deviates from cql system types.

What about, `ParsedFilterInterval` or someth. @hossenlopp ?

mgramigna (commented):
Functionality looks good, just a few comments, and nothing glaring

src/types/QueryFilterTypes.ts
hossenlopp: I think `ParsedFilterInterval` is okay. Definitely want to differentiate it from the CQL system types.

mgramigna (approved):
Looks good, great improvements

PR #84, opened 2021-08-04
Export required helpers in index.ts, and port Measure Bundle Helpers to their own file

okeefm:
# Summary
In order to use ELM/Measure helper functions in `cql-module-generator`, we have to export them from this library. The entrypoint into this library is `src/index.ts`, so that's where we have to add them.

Also, we moved helpers out of `ClauseResultsHelper` that only had to do with parsing and extracting info from the Measure Bundle resource into their own helper file, `src/helpers/MeasureBundleHelpers.ts`. This file then gets exported as `MeasureBundleHelpers` in `index.ts`.

## New behavior
Added exports for:
* `ELMHelpers`
*  `ELMDependencyHelpers`
* `MeasureBundleHelpers`

Moved these helpers from `ClauseResultsHelpers` to `src/helpers/MeasureBundleHelpers`:
* `isEpisodeOfCareMeasure`
* `extractMeasurementPeriod`
* `extractLibrariesFromBundle`
* `extractMeasureFromBundle`
* `isValidLibraryURL`
* `codeableConceptToPopulationType`

And updated references and tests as a result of those moves.
## Code changes
See above
# Testing guidance
If you run `npm link` in the `fqm-execution` directory while this PR is checked out, then run `npm link fqm-execution` in another project that uses this one (`fqm-execution-service`, `fqm-execution-demo`, etc.), you should be able to successfully `import` the exported helpers in a typescript file. Should look something like:
`import { ELMHElpers, ELMDependencyHelpers, MeasureBundleHelpers } from 'fqm-execution';`


sarahmcdougall (approved):
Everything looks good to me and works as it should. Thanks for helping me troubleshoot the npm issues on my computer!

PR #88, opened 2021-08-13
Integration changes

MegRiley:
# Summary
I moved the code to execute the integration tests from under the ./src directory to the ./test/integration-test folder.
## New behavior
There should be no behavior changes and everything should function as before
## Code changes
- added tsconfig.int.json - this applies the root tsconfig.json to the test folder and extends it to ./test/integration-test
- updated the paths in multiple files that call the integration tests or are importing from src
# Testing guidance

1. run: npm install
2.  execute ./test/integration-tests/setup.sh (this will drop the files from the connectathon repo/fpg folder into the expected places)
3. Now run: npm run test:integration EXMXXX (XXX is your measure of choice from the connectathon repo 401 subfolder, I personally tested with EXM104) 

The tests should run to completion and report no issues


test/integration-tests/tsconfig.int.json
```diff
@@ -0,0 +1,7 @@
+
```
hossenlopp, line 1: This file desperately needs formatting.

hossenlopp (changes requested):
This move looks decent. The instructions to run locally didn't work off the bat for me. It should be a single `:` in the `npm run` call. Also it just blew up since I didn't run test_setup.sh first. This seems like something to address some other time.

Just fix the formatting on the new tsconfig and this would look good.

hossenlopp (approved):

PR #89, opened 2021-08-13
Switch to @types/fhir for FHIR TypeScript Definitions

hossenlopp:
# Summary
Replaced `@ahryman40k/ts-fhir-types` usage with `@types/fhir` for FHIR JSON type defs.

## New behavior
Nothing except for the Composition.status that we now include in gaps reports. Turns out that was a required field.

## Code changes
- Every file that worked with FHIR JSON was updated to make use of the new `@types/fhir` definitions.
- These definitions do not require explicit inclusion in each file. They provide a global namespace `fhir4` that is used to access them.
- Some complex sub-structures had simplistic names in the old library. The new library is a bit more explicit in their naming by including the entire path.
- Enums for FHIR codes were replaced with simply a list of expected strings, this changed they way in how we set the code fields slightly.
 
# Testing guidance
Run unit tests. Try various run configurations.


sarahmcdougall (approved):
All the unit tests pass and all run configurations work as expected. All the changed code makes sense. Up to you if you want another reviewer, but it looks good to me to merge!

PR #91, opened 2021-08-18
Error types

natjoe4:
# Summary
Added ErrorWithDebugInfo, UnsupportedProperty, UnexpectedProperty, and Unexpected Resource errors and converted thrown errors to these types
## New behavior
- calculateRaw should now throw an error when it fails to generate rawResults. Previously was returning an object with the error string
- Behavior should be otherwise unchanged except that error logs will have 'UnsupportedProperty', 'UnexpectedProperty' or 'UnexpectedResource' listed as the name when appropriate
## Code changes
- Created ErrorWithDebugInfo as a subclass of Error
- Created UnsupportedProperty as a subclass of ErrorWithDebugInfo
-- Used when all resources are valid but are requesting services we do not yet support
-- ex. user is trying to calculate measure reports with a report type other than 'summary' or 'individual'
- Created UnexpectedProperty as a subclass of ErrorWithDebugInfo
-- Used when a resource largely valid/syntactically correct, but has conflicting info, is impossibly formatted, or is missing specific needed fields
-- ex. Measure Bundle does not contain improvement notation 
- Created Unexpected Resource as subclass of ErrorWithDebugInfo
-- Thrown when a resource is improperly formatted or empty
-- ex. patient bundle has no entries, measure bundle has no libraries
- Changed all appropriate errors to one of these new error types
- Left some errors (errors where we do not have enough information on how they occur) as regular errors which will be interpreted by fqm-execution-service as standard 500 errors.
# Testing guidance
- run `npm run check` to make sure everything works as usual.
- Using vs-code search feature (magnifying glass on left-hand bar) search 'throw new' to find all places where fqm-execution throws an error with a custom message. Aside from in the integration-tests and cli, ensure the selection for types of all errors make sense.


src/execution/ValueSetHelper.ts
```diff
@@ -112,8 +113,12 @@ export function getMissingDependentValuesets(
...
   const vsURLs: string[] = libraryEntries.reduce((acc, lib) => {
     const libraryResource = lib.resource as fhir4.Library;
-    if (!libraryResource || !libraryResource.dataRequirement || !(libraryResource.dataRequirement.length > 0)) {
-      throw new Error('Expected library entry to have resource with dataRequirements that have codeFilters');
+    if (!libraryResource) {
+      throw new UnexpectedResource('Library entry not included in measure bundle');
+    } else if (!libraryResource.dataRequirement || !(libraryResource.dataRequirement.length > 0)) {
+      throw new UnexpectedProperty(
+        'Expected library entry to have resource with dataRequirements that have codeFilters'
+      );
```
natjoe4, line 121: Here's what's causing the reduced coverage. Can try to remedy this if people think its important, but my feeling is that it's a very direct translation from the previous lines, and doesn't change the functionality

sarahmcdougall (approved):
Everything works as usual with `npm run check` and `npm run test`. I looked through all the places where the errors are thrown, and they all make sense to me for the given situations. Good work!

PR #100, opened 2022-01-20
Passed-in PatientSource Refactor

hossenlopp:
# Summary

Now able to use a provided PatientSource for all calculation interfaces instead of requiring a list of PatientBundles.

## New behavior

If a PatientSource is passed in (as already defined in the options object) the patientBundles array will be ignored and the PatientSource will be used instead. 

## Code changes

- Creates a PatientSource using cql-exec-fhir for a given list of patient bundles if a PatientSource is not provided.
- The "PatientObject" used during calculation is pulled from results and added to ExecutionResult object so it can be used for later calculation such as gaps in care and MeasureReport building.
- All iterations over `patientBundles` have been removed. Iterations are now done over results.

# Testing guidance

Run unit tests.

Run CLI with the new `-f` option to make it so a PatientSource is passed in to the API instead of an array of bundles. Preferably asking for a summary report or Gaps in Care.

For example a summary report run assuming dbcg connectathon is checked out at `../dbcg-connectathon`:
```
node build/cli.js -d -o reports -r summary -s 2019-01-01 -e 2019-12-31 -m ../dbcg-connectathon/fhir401/bundles/measure/EXM130-7.3.000/EXM130-7.3.000-bundle.json -p test/fixtures/EXM130-8.0.000/EXM130-8.0.000-patients/numerator/Alejandro916_Becker968_6a50325f-89c6-4d09-88e7-1eede4a49df5.json test/fixtures/EXM130-8.0.000/EXM130-8.0.000-patients/denominator/Abdul218_Fahey393_1c14630a-f7f6-49cd-9c38-b8c975b0864f.json -f
```

### DIFF
--- a/src/execution/Execution.ts
+++ b/src/execution/Execution.ts
@@ -1,6 +1,5 @@
 import { CalculationOptions, RawExecutionData, DebugOutput } from '../types/Calculator';
 
-// import { PatientSource } from 'cql-exec-fhir';
 import cql from 'cql-execution';
 import { parseTimeStringAsUTC, valueSetsForCodeService, getMissingDependentValuesets } from './ValueSetHelper';
 import * as MeasureBundleHelpers from '../helpers/MeasureBundleHelpers';
--- a/src/execution/Execution.ts
+++ b/src/execution/Execution.ts
@@ -2,7 +2,6 @@ import { CalculationOptions, RawExecutionData, DebugOutput } from '../types/Calc
 
 // import { PatientSource } from 'cql-exec-fhir';
 import cql from 'cql-execution';
-import { PatientSource } from 'cql-exec-fhir';
 import { parseTimeStringAsUTC, valueSetsForCodeService, getMissingDependentValuesets } from './ValueSetHelper';
 import * as MeasureBundleHelpers from '../helpers/MeasureBundleHelpers';
 import { PopulationType } from '../types/Enums';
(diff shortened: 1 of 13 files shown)


src/execution/Execution.ts
```diff
@@ -2,7 +2,6 @@ import { CalculationOptions, RawExecutionData, DebugOutput } from '../types/Calc
 
 // import { PatientSource } from 'cql-exec-fhir';
```
natjoe4, line 3:
```suggestion
```
not your addition, but looks like this can be removed

src/gaps/QueryFilterParser.ts
```diff
@@ -838,10 +838,11 @@ export function interpretGreaterOrEqual(
         // as the number of years to add to the birthDate.
         if (greaterOrEqualExpr.operand[1].type === 'Literal') {
           const years = (greaterOrEqualExpr.operand[1] as ELMLiteral).value as number;
+          debugger;
```
natjoe4, line 841:
```suggestion
```

src/execution/Execution.ts
hossenlopp: Absolutely can be removed here since this file only speaks generic PatientSource.

src/gaps/QueryFilterParser.ts
hossenlopp: Removed!

natjoe4 (approved):
Code looked good to me and was able to run a gaps calculation with the -f flag as suggested as well as without.

src/types/CQLExecution.d.ts
```diff
@@ -73,4 +73,20 @@ declare module 'cql-execution' {
...
+    nextPatient(): Patient | undefined;
+  }
+
+  export interface IRecord {
+    get(field: string): any;
+    _json: any;
+  }
+
+  export interface IPatient extends IRecord {
+    findRecords(profile: string): IRecord;
```
mgramigna, line 88: Should this be `IRecord[]`?

mgramigna (commented):
Super solid. Works and code looks good/is simple. I just have one question/comment inline

src/types/CQLExecution.d.ts
hossenlopp: Yep! Fixed.

mgramigna (approved):

PR #101, opened 2022-02-02
Add fhirQueryPattern extension to DataRequirements

sarahmcdougall:
# Summary
The `fhirQueryPattern` extension was added to the `dataRequirement` elements in the `dataRequirements` array when we calculate data requirements.

## New behavior
To learn more about the cqfm fhirQueryPattern extension, see the following link: http://build.fhir.org/ig/HL7/cqf-measures/StructureDefinition-cqfm-fhirQueryPattern.html

**Motivation for this extension:** Different servers will support different search parameters. For example, for a Procedure, a given server may not support `code:in` as a search parameter. Therefore, we should add a query `valueString` as an extension for each data requirement.

When calculating for data requirements, the resulting `Library` resource’s `dataRequirement` array will contain `dataRequirement` elements that each contain an `extension` array with the fhirQueryPattern extension(s). If the given resource can reference a Patient resource in more than one way, then there will be a fhirQueryPattern extension for each way that the Patient can be referenced. For example, Procedure can reference the Patient as `subject` and `performer.actor`, so the extension will contain an element for each of these Patient contexts.

## Code changes
New functions were added to `DataRequirementHelpers.ts` to add the extensions to the data requirements. The `addFhirQueryPatternToDataRequirements()` function creates the query string by taking all the elements of the `codeFilter` array and adding them to the endpoint as search params. The patient context is added onto the query string using `PatientReferences.json`. `PatientReferences.json` was created from the patient compartment definition (the script for parsing the patient compartment definition is available in deqm-test-server). Date filters are also incorporated into the query string. It is assumed that we will always be working with `valuePeriod`, which is inclusive (hence the use of `ge` and `le`).

Unit tests were added for construction of the fhirQueryPattern `valueString`s.

# Testing guidance
Run the new unit tests found in `DataRequirementHelpers.ts` and ensure that they pass. 
It is best to test out these changes using the `fqm-execution` CLI with a command like the following:

```
node build/cli.js -o dataRequirements -m {path to EXM130-7.3.000-bundle} -p {path to EXM130-7.3.000 patient bundle} > reports.json
```
For each `dataRequirement` element in the `dataRequirement` array, there should be an `extension` array. This `extension` array should contain an object containing the fhirQueryPattern url and a `valueString` representing the query.

To see more examples of what the `fhirQueryPattern` query should look like, see the unit tests in this PR from the cql engine: https://github.com/DBCG/cql_engine/pull/513/files


src/compartment-definition/PatientReferences.json
```diff
@@ -0,0 +1,234 @@
+{
```
lmd59, line 1: I know that the PR description says that the sources of this is the compartmentdefinition parsing done with scripts in deqm-test-server. Do you think there's somewhere we can explicitly site that in either comments (stupid json not allowing comments) or fqm-execution documentation? If compartment definition stuff changes in the future, I think it's valuable to have as much of a thread as possible for future developers to be able to make the updates needed.

src/types/QueryFilterTypes.ts
```diff
@@ -129,3 +129,19 @@ export interface ParsedFilterInterval {
   end?: string;
   interval?: cql.Interval;
 }
+
+/**
+ * Represents object containing query parameters and their corresponding values.
+ */
+export interface codeFilterPath {
```
lmd59, line 136: This seems to be representing a generic hash/dictionary type thing. Not a typescript expert, but... Do we think it would make sense to use something standard, like it looks we could use something like the [Record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) utility type, or if that's not a good fit, maybe we could define cross project hash interface utility type?

lmd59 (approved):
This code is clutch. As such, I got to have fun being smidge nitpicky at things. Approved, but let me know what you think or if you have any comments on these two suggestions.

src/compartment-definition/PatientReferences.json
mgramigna: we could make this a `.ts` file and do like `export const PatientReferences = {...}`?? Then we could include any comments/citations at the top of the file or something

src/types/QueryFilterTypes.ts
mgramigna: I'm so proud of this community

mgramigna: Lauren understood the assignment of my "I encourage you to look into some of the other types you can use in TypeScript" during my talk 😭

lmd59 (approved):
Amazing. Looks great to me!

PR #102, opened 2022-02-11
Defaulted calc options to {} in data requirements function

mgramigna:
# Summary

Introducing `calculationOptions` as an argument to the data requirements function actually introduced a breaking change to JS users of the library where the omission of the `options` parameter would throw a runtime error when trying to find the measurements period.

This PR just defaults it to `{}` to prevent that breaking change. The argument can still be left blank and the measurement period will be defaulted to 2019 as it always has 

## New behavior

None

## Code changes

Default arg to `{}`

# Testing guidance

run DR calculation and ensure it does stuff


natjoe4 (approved):
Looks great to me!

PR #104, opened 2022-02-17
Gaps Reason Code Update

natjoe4:
# Summary
The gaps in care reason codes, displays, and descriptions were recently updated. This PR updates our gaps in care reason codes to match the current codes and displays

## New behavior
- When running with output type "gaps" the output reason code displays will be different
- The removal of the ValueMissing code from the code list forced us to change all instances of ValueMissing to NotFound. ValueMissing was previously used to describe the situation where a target resource existed, but the value for the attribute of interest on that resource was missing. NotFound is a more general reason code which describes both the above situation, and the situation in which the desired resource does not exist at all

## Code changes
- Updated CareGapReasonCode and CareGapReasonCodeDisplay Enums to reflect the most recent codes and displays
- Updated all instances of ValueMissing to NotFound
- Updated default Missing reason codes to NotFound, as the new NotFound description better fits a default cannot find resource situation
- Updated testing and fixtures

# Testing guidance
- Run all unit tests
- Look through new gaps reason codes at bottom of page [here](https://fshschool.org/FSHOnline/#/share/3oRwnmE) and ensure they match the updated enum in `src/types/Enums.ts`. Note, the display is the first string listed in the .fsh file for each reason code
- Try running some "gaps" operations on various bundles (I used EXM130 and EXM124) and ensure that the output reason codes and displays are correct. Note that the only code changes should be `Missing -> NotFound` and `ValueMissing` -> NotFound. All other codes should remain the same, but all displays should be different.


src/types/Enums.ts
```diff
@@ -85,33 +85,45 @@ export enum ImprovementNotation {
...
+  [CareGapReasonCode.NOTFOUND]: 'Data Element Not Found',
+  [CareGapReasonCode.MISSING]: 'Missing Data Element',
+  [CareGapReasonCode.PRESENT]: 'Data Element is Present',
+  [CareGapReasonCode.INVALIDATTRIBUTE]: 'Attribute is Invalid',
+  [CareGapReasonCode.DATEINRANGE]: 'Date is within specified range',
+  [CareGapReasonCode.DATEOUTOFRANGE]: 'Date is out of specified range',
+  [CareGapReasonCode.VALUEINRANGE]: 'Value is within specified range',
+  [CareGapReasonCode.VALUEOUTOFRANGE]: 'Value is out of specified range',
+  [CareGapReasonCode.COUNTINRANGE]: 'Count is within specified range',
+  [CareGapReasonCode.COUNTOUTOFRANGE]: 'Count of data elements was not in the expected range',
```
sarahmcdougall, line 127: I'm not sure if we need to exactly follow the text from the FSH  CodeSystem, but for `COUNTOUTOFRANGE` the text is "Count is out of specified range" and not "Count of data elements was not in the expected range"

sarahmcdougall (approved):
Aside from one small comment, everything looks good! I tested with EXM130 and EXM124, and ran the unit tests as well. Looks like this needs a rebase, but after that it should be good to merge.

PR #105, opened 2022-02-18
Add Gaps in Care support for multiple patient bundles

sarahmcdougall:
# Summary
This PR makes small changes to `calculateGapsInCare()` to allow results to be returned for multiple patient bundles.

## New behavior
Previously, `calculateGapsInCare()` would allow a patient bundles array as a parameter, but the `result` variable would only store the results for a single patient. Now, when calculating gaps in care with multiple patient bundles, an array will be returned with an entry for each patient. To avoid breaking changes, if only one patient bundle was provided, then the Bundle that is returned will not be wrapped in an array.

## Code changes
Changes were made in `calculateGapsInCare()` regarding the `result` variable. The output from `generateGapsInCareBundle()` is now pushed onto a `result` array of FHIR Bundles. 

# Testing guidance
First, try a basic request with output `"gaps"` for a single patient bundle, such as:

```
node build/cli.js -o gaps -m <PATH TO MEASURE BUNDLE> -p <PATH TO PATIENT BUNDLE>
```
The output should be the same as it appears on the master branch.

The, try sending a request with multiple patient bundles. For example, if testing with EXM 130, the command would be:

```
node build/cli.js -o gaps -m <PATH TO EXM130 MEASURE BUNDLE> -p <PATH TO EXM130 NUMERATOR PATIENT BUNDLE> <PATH TO EXM130 DENOMINATOR PATIENT BUNDLE> -s 2019-01-01 -e 2019-12-31
```
The results should be an array of length two. Note that the numerator patient's array entry will be an empty object.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "fqm-execution",
-  "version": "0.8.1",
+  "version": "0.9.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fqm-execution",
-  "version": "0.8.1",
+  "version": "0.9.0",
   "description": "FHIR Quality Measure Execution",
   "main": "build/index.js",
   "types": "build/index.d.ts",
--- a/src/calculation/Calculator.ts
+++ b/src/calculation/Calculator.ts
@@ -327,7 +327,7 @@ export async function calculateGapsInCare(
   const { results, debugOutput, elmLibraries, mainLibraryName, parameters } = calculationResults;
   const measureReports = MeasureReportBuilder.buildMeasureReports(measureBundle, results, options);
 
-  let result: fhir4.Bundle = <fhir4.Bundle>{};
+  const result: fhir4.Bundle[] = [];
   const errorLog: GracefulError[] = [];
   results.forEach(res => {
     const matchingMeasureReport = measureReports.find(mr => mr.subject?.reference?.split('/')[1] === res.patientId);
@@ -370,7 +370,6 @@ export async function calculateGapsInCare(
       const populationCriteria =
         numerRelevance &&
         (improvementNotation === ImprovementNotation.POSITIVE ? denomResult && !numerResult : numerResult);
-
       if (populationCriteria) {
         const matchingGroup = measureResource.group?.find(g => g.id === dr.groupId) || measureResource.group?.[i];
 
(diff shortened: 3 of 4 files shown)


mgramigna:
Okay, buckle up.

The one thing we didn't consider when thinking about potential breaking changes from this is other TypeScript users of the library (think `fqm-execution-demo` for instance. The `calculateGapsInCare` function can now return `Bundle | Bundle[]`, so TS will not know for someone who calls this function which one it is. This would then require them to do an additional check or casting on their end, and that constitutes a breaking change.

Now you may be thinking to yourself: "okay, that's fine. We'll just compromise on that change, surely there's no way to avoid that in TypeScript?" BUT THERE IS. We can use a combination of [Generics](https://www.typescriptlang.org/docs/handbook/2/generics.html) and [Conditional Types](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html) in order to tell TypeScript whether or not this function returns 1 or many bundles based on the type of the `patientBundles` argument!! This might be the craziest typescript concoction I've ever come up with, and I couldn't be more proud. I'm about 98% sure it'll work and not introduce a breaking change.

I'll summarize the changes here, but we should likely talk about this verbally:

1. Create a type that indicates that the `patientBundles` argument can either be an array of 1 element, or an array of many:

``` TypeScript
type OneOrMultiPatient = [fhir4.Bundle] | fhir4.Bundle[];
```

2. Create a conditional type that essentially maps the above type into the return result of gaps in care:

``` TypeScript
type OneOrManyBundles<T extends OneOrMultiPatient> = T extends [fhir4.Bundle] ? fhir4.Bundle : fhir4.Bundle[];
```

In english, this is saying "If I use the `OneOrManyBundles` type with an array that only has one element, then the type is just one fhir Bundle. Otherwise, the type is an array of multiple fhir Bundles.

3. Change the `GICCalculationOutput` type to use generics for a type `T extends OneOrMultiPatient` and have it return `OneOrManyBundles<T>` as the result. I'll leave the code as an exercise to the reader ;). Dang this is so fun
4. Now, in `Calculator.ts`, we can use generics for a type `T extends OneOrMultiPatient`, which says that type `T` is either `[fhir4.Bundle] | fhir4.Bundle[]`, and we should know change `patientBundles` arg to be of type `T`
5. Now that we've changed `GICCalculationOutput` to a generic, we need to change the return type of `calculateGapsInCare` to `Promise<GICCalculationOutput<T>>`
6. At this point, we now know that our return result is going to be one Bundle or an array of Bundles, so we can cast the result as a `OneOrManyBundles<T>`

--

Boom. There shouldn't be any breaking changes now. When you call `calculateGapsInCare` as follows, the return type will be inferred by TypeScript, and everybody continues with their day:

``` TypeScript
const { results } = await calculateGapsInCare(mb, [pb], {}); // TypeScript knows that `results` is a `fhir4.Bundle`
```
``` TypeScript
const { results } = await calculateGapsInCare(mb, [pb, pb2], {}); // TypeScript knows that `results` is a `fhir4.Bundle[]`
```

God I love this language. Talk to ya Monday!

mgramigna:
functionality looks good btw

mgramigna (approved):
LGTM. Let's rebase this branch and then we can publish fqm-execution 0.9.0

PR #106, opened 2022-02-18
Fixed effective.end ReasonDetail Bug

natjoe4:
# Summary
We discovered a bug in gaps that causes the reasonDetail extension to be populated with an additional, erroneous element in cases where the patient is in the denominator, and one of the attributes necessary for reasonDetail extraction is a dynamic attribute which can have its value stored at varying depth. If that value ultimately abides by the filter, it could still be flagged as a reason for the gap, since our existing code could not handle this dynamic depth and the value would return undefined. 

Confused? Same. 

Here's an example. This situation was discovered when running EXM124 on a denominator patient. The patient falls in the denominator because the .value on their Cervical Cytology observation is null. The reasonDetail generation code then runs through each of the ELM filters to determine why the patient fell into the denominator. It finds the correct reason and logs it, but it also applies the ELM filter that checks if the Cervical Cytology Observation is within the effective interval. The filter expects the Observation to store information about its effectiveness interval in a period object, with the data on the path effective.end.value, but instead, it is just stored in an effective date time object, with the data on the path effective.value. The code which traverses the Observation object to find the desired attribute value blows right past this attribute and determines it to be null, despite the fact that the value is within the desired interval.

This PR breaks the loop that traverses the Observation object for DateTimes so that is stops "drilling down" as soon as we encounter a .value element along the provided path which is of type DateTime

## New behavior
The above-described behavior should no longer occur. When running gaps on the denominator patient for EXM124, the reasonDetail extension on the first detectedIssue should have just one ValueMissing reason for the .value property being null on the patients Cervical Cytology.

## Code changes
Updated the "drill down" loop for DateTimes in `calculateReasonDetail` in `GapsReportBuilder.ts` to break when it reaches a level along the path with a .value attribute that is of type DateTime

# Testing guidance
- Run `npm test`
- Download EXM124 bundle and the denominator patient bundle from the [Connectathon repo](https://github.com/DBCG/connectathon/tree/master/fhir401/bundles/measure/EXM124-9.0.000) if you haven't already
- Run `npm run build` in fqm-execution
- Run a gaps report using `node build/cli.js -o "gaps" -m _"path/to/EXM/124/Measure/Bundle"_ -p _"path/to/EXM/124/denom/patient/bundle"_ > output.json`
- Ensure that the reasonDetail for the first detectedIssue only contains one reason and its that the .value property was not found
- Edit line 336 in `tests-denom-EXM124-bundle.json` so that the effectiveDateTime of Observation with ID `denom-EXM124-3` is `"2010-11-01T00:00:00"`
- Rerun the command to generate the gaps report
- Ensure that on the first detected issue, there are now two reasonDetail objects, one for the initial null value not being found, and one DateOutOfRange code which references the `effective` attribute


MegRiley (approved):
Looks good to me

PR #109, opened 2022-03-01
Added support QICore measures that output CQL System Codes for SDEs

hossenlopp:
# Summary

QICore measures return CQL System Codes instead of FHIR Codings for some SDEs. This expands support for this in the measure report builder. The builder already supported singular CQL Codes but it wasn't supported as an array.

## New behavior

Allows for SDEs that return list of CQL System Codes to be used in SDE compilation in MeasureReportBuilder.

## Code changes

In `MeasureReportBuilder.addSDE` allows for CQL System Codes to be returned in execution results to be valid and converted to SDEs. Adds a new test file for this

# Testing guidance

This one is a doozy. Running unit tests as-is is one way. To test with a real QICore measure we need to use a temporary patch to `cql-exec-fhir` and use patched test data.
- Setup the `cql-exec-fhir` dependency to use `https://github.com/projecttacoma/cql-exec-fhir/tree/profile-fallback`
  - Check out the `profile-fallback` branch in `projecttacoma` fork.
  - Run `npm install`.
  - Use `npm link` or change the dependency in package.lock to `file:/../path/to/cql-exec-fhir` clone.
- Run QICore EXM124 individual MeasureReport calculation with patient bundle from @hossenlopp on slack.
  - https://github.com/cqframework/ecqm-content-qicore-2020/blob/master/bundles/measure/EXM124v7QICore4/EXM124v7QICore4-bundle.json
- Look for the SDE Race and SDE Ethnicity Observation resources contained in the MeasureReport. Ensure these have codes in the `valueCodableConcept` attribute. Without this fix, the SDEs will not be populated with codes in the `valueCodableConcept` of the SDE Observations.

### DIFF
--- a/src/calculation/MeasureReportBuilder.ts
+++ b/src/calculation/MeasureReportBuilder.ts
@@ -323,6 +323,13 @@ export default class MeasureReportBuilder {
               code: rr.code.value,
               display: rr.display.value
             });
+          } else if (rr.isCode) {
+            // if a CQL system code is returned
+            observation.valueCodeableConcept?.coding?.push({
+              system: rr.system,
+              code: rr.code,
+              display: rr.display
+            });
           }
         });
       } else if (sd.rawResult.isCode) {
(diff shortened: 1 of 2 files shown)


mgramigna (approved):
Looks good. Generates proper SDEs on this branch that aren't there on master when running execution

PR #113, opened 2022-03-22
Basic results interface

mgramigna:
# Summary

This PR introduces a new "slim" mode to detailed results calculation that does not include verbose reporting of clause results, statement results, relevance, and highlighted HTML. By default, it will use the verbose output unless the `verboseCalculationResults` calculation option is set to `false`

## New behavior

* For the CLI, when using the `--slim` options, it will strip off all the unnecessary info on a detailed result that doesn't contribute to interpreting the population
* For the `calculate` API function, when provided `verboseCalculationResults: false`, it will strip off the irrelevant information

## Code changes

* `calculate` strips off unneeded information when `verboseCalculationResults` is false
* `calculate` is now a generic on CalculationOptions
  * This allows the function to infer its own return type based on the value of `verboseCalculationResults`
* Update types to support returning of simple or detailed results
* Update MeasureReportBuilder to be generic on type of detailed result
  * Can optionally do `new MeasureReportBuilder<SimplePopulationGroupResult>(...);` or `new MeasureReportBuilder<DetailedPopulationGroupResult>(...);`
* Le unit tests

# Testing guidance

* Run cli with `--slim` flag on reports and detailed output types. For the former, it should not generate an HTML narrative. For the latter, it should only generate `populationResults`
* Try to mimic a call to the calculate function yourself in a different file with custom calculation options, e.g.:

``` Typescript

const b = <fhir4.Bundle>{};
calculate(b, [b], {
  verboseCalculationResults: false
}).then(({ results }) => {
  // results should be of type ExecutionResult<SimplePopulationGroupResult>[]
});

calculate(b, [b], {
  verboseCalculationResults: true
}).then(({ results }) => {
  // results should be of type ExecutionResult<DetailedPopulationGroupResult>[]
});

calculate(b, [b], {}).then(({ results }) => {
  // results should be of type ExecutionResult<DetailedPopulationGroupResult>[]
});
```

This is the magic of generics

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ const queryInfo = Calculator.calculateQueryInfo(measureBundle); // Get detailed 
 The options that we support for calculation are as follows:
 | option | type | optional? | description |
 | :--------------------- | :-----: | :-------: | :--------------------------------------------------------------------------------- |
+| verboseCalculationResults | boolean | yes | Use the detailed results interfaces for calculation. Defaults to true. |
 | enableDebugOutput | boolean | yes | Enable debug output from function calls. Defaults to false. |
 | includeClauseResults | boolean | yes | Option to include clause results. Defaults to false. |
 | includePrettyResults | boolean | yes | Option to include pretty results on statement results. Defaults to false. |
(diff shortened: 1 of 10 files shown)


hossenlopp (changes requested):
We probably will not need the evaluated resources. But keep patient.

hossenlopp (approved):
Good to go now! Will have some caveats if someone ask for reports with simple output, but that's on them and may actually be a good way to have a less busy MR if they don't need everything.

PR #112, opened 2022-03-15
AsyncPatientSource Functionality and CLI Overhaul

natjoe4:
# Summary
fqm-execution now supports calculation using a cql-exec-fhir AsyncPatientSource. Also cleaned up the CLI and changed the -o flag settings to commands

## New behavior
All existing functionality should be unchanged. Added the ability to pass in a fhir server url and a list of patient ids, to replace patient bundles as a calculation option. Uses cql-execution's new async behavior and cql-exec-fhir's new AsyncPatientSource and AsyncPatient to accomplish this.

### New cli options: 
- --fhir-server-url: a url to a fhir server that contains data for calculation
- --patient-ids: a list of ids which signifies which patients a measure will be run on when using an async patient source
### Updated cli options: 
-d, --debug -> --debug
-f, --as-patient-source -> --as-patient-source
-c, --cache-valuesets -> --cache-valuesets
-a, --vs-api-key -> --vs-api-key
-r, --report-type -> --report-type
### Other changes
Updated Execution to properly await cql-execution execute function
Updated gaps code to correctly await expression executions
Updating testing to account for new async functions

# Testing guidance
- Run unit tests
- Make sure you have the most recent versions of projecttacoma/cql-execution, projecttacoma/cql-exec-fhir, and projecttacoma/deqm-test-server
- Reset your test server db using `npm run db:reset`, then `npm run connectathon-upload`
- Link cql-exec-fhir to fqm-execution by running `npm link path/to/projecttacoma/cql-exec-fhir`
- Link cql-execution to fqm-execution by running `npm link path/to/projectacoma/cql-execution`
- Start up deqm-test-server using `npm start`
- Run `npm run build` in fqm-execution
- Ping me for a file of requests to use for testing. The async patient source version of each request should match the same request when run with regular patient bundles
- Run your own test commands! Make sure you run through each output type and report type.
- Test out invalid combinations of options and make sure that a proper error is thrown
- Look over the cli and the readme to make sure everything is documented correctly. Try running `node build/cli.js -h` and make sure the help command is useful and properly formatted.

PR #120, opened 2022-04-19
Added new cli flag to support meta.profile validation 

MegRiley:
# Summary
Added a new  CLI flag that says whether or not to "trust" the content of meta.profile as a source of truth for what profiles the data that cql-exec-fhir grabs validates against.
The flag is  stored in the CalculationOptions, and passed along to construction of a cql-exec-fhir patient source. It  defaults to false and is optional.

## New behavior
The default behavior is unchanged, if the user chooses to pass the flag `--profile-validation` to the cli or via launch.json fqm-execution will use meta.profile as the source of truth for what profiles the data that cql-exec-fhir grabs validates against.
## Code changes
`Readme.md` - added documentation for the flag
`src/types/Calculator.ts` - added  `trustMetaProfile` flag to track if the calculation was supposed to use the meta.profile as the source of truth. 
`src/cli.ts` - added new cli flag to track signal if we should use the meta.profile, this flag defaults to false  and is optional as a result. Modified calls to `PatientSource` constructor to now accept this flag, updated the creation of `calcOptions` object to store this flag as well. 
# Testing guidance
Run all tests, and make sure unit tests still pass. 
 I used the modified bundles that Nat had previously used for testing [this pr](https://github.com/projecttacoma/cql-exec-fhir/pull/4) as well as using the qi-core bundles referenced  [here](https://github.com/projecttacoma/cql-exec-fhir/pull/4)

Testing:
Run a reports calculation on EXM130 using the altered numerator patient bundle with the `--as-patient-source` and the `--profile-validation`  flag. You should see the patient calculate into the numerator
Now, change the meta.profile of Procedure numer-EXM130-1 in any way you like (I just added a 1 to the end)
Rerun the same report and the patient should evaluate into the denominator, as the necessary procedure is no longer evaluated

Testing an async patient source:
 You should upload the qi core measure and patient bundle to a running deqm-test-server
Then use http://localhost:3000/4_0_1 as the --fhir-server-url CLI option, and set the --profile-validation`  flag as well
And the ID of the patient resource in the qi core numerator patient for the --patient-ids CLI option
Thttps://github.com/projecttacoma/fqm-execution/pull/112 are further details about testing with async patient sources in this pr here

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ The options that we support for calculation are as follows:
 | calculateHTML | boolean | yes | Include HTML structure for highlighting. Defaults to false. |
 | vsAPIKey | string | yes | API key, to be used to access a valueset API for downloading any missing valuesets |
 | useValueSetCaching | boolean | yes | Whether to cache valuesets obtained by an API on the filesystem |
+| profileValidation | boolean | yes |  To "trust" the content of meta.profile as a source of truth for what profiles the data that cql-exec-fhir grabs validates against|
 
 ### CLI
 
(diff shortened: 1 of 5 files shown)


mgramigna (changes requested):
Looks good. We do need this option to be passed along to PatientSource construction that happens when the `--as-patient-source` isn't used.

i.e. we need to use the `trustMetaProfile` option in the `resolvePatientSource` function as well since that is the default behavior for constructing a patient source.

PR #121, opened 2022-04-20
Value filters for comparator parsing

natjoe4:
# Summary
Adds functionality for parsing comparator operators (Greater, Less, GreaterOrEqual, etc...) and returning value filters for comparisons performed outside the context of a where clause

## New behavior
In order to run gaps on the HbA1c measure, we need to check that the last observation has a quantity with value greater than 9%. This was previously uncovered since the CQL has this set up as a Greater operation, one side of which is a toQuantity of a Last operation, all of which is outside of a where clause. To work around this, we added a branch inside the `interpretFunctionRef` function that checks if the functionRef is a toQuantity, and if so, passes through to the underlying functionRef. We also added the `interpretComparator` function which takes in any ELMComparator and returns a corresponding ValueFilter. The result is that we now have ValueFilters returned for the HbA1c measure which have a local reference to the top level query, giving us an avenue to calculate a reasonDetail for detected gaps.

## Code changes
- Added new `toQuantity` branch for `interpretFunctionRef`
- Added new functions `interpretComparator`, `interpretGreater`, `interpretLess`, and `interpretLessOrEqual`
- Updated `interpretGreaterOrEqual` to call `interpretComparator` when appropriate
- Added `ValueComparisonLocalId` tracking to `parseQueryInfo`
- Added and updated testing

# Testing guidance
- run `npm test`
- Ping me for a test bundle/commands to use


src/calculation/Calculator.ts
```diff
@@ -518,44 +519,47 @@ export async function calculateDataRequirements(
     }
   });
 
-  // Only use "unique" retrieves
```
hossenlopp, line 521: I was about to throw a huge red flag on this code. But I noticed it was the removed code. 😆

We do want to look at ALL retrieves and how they are filtered. The hard part of this is that we may want to effectively "or" the requirements of the retrieves with matching resourceType and valueSet or code (aka. data element) together. But I think for now its okay if multiple dataRequirements are generated for the same "data element". That does effectively communicate the "or" scenario just fine.

src/gaps/QueryFilterParser.ts
```diff
@@ -65,7 +73,8 @@ import { UnexpectedResource } from '../types/errors/CustomErrors';
 export async function parseQueryInfo(
   library: ELM,
   allELM: ELM[],
-  queryLocalId: string | undefined,
+  queryLocalId?: string,
+  valueComparisonLocalId?: string,
```
hossenlopp, line 77: This new parameter should have info about what it is in the JSDoc.

src/gaps/QueryFilterParser.ts
```diff
@@ -929,13 +965,118 @@ export function interpretGreaterOrEqual(
...
+}
+
+/**
+ * Default code for parsing a miscellaneous comparator expression
+ * @param comparatorELM The LessOrEqual expression to interpret.
+ * @param library The library it belongs in. This is needed to identify parameters.
+ * @param comparatorString a string determining the type of comparator passed into the function
+ * @returns Filter representation of the comparator clause this will be Unknown or Value depending on if it can be parsed or not.
+ */
+export function interpretComparator(
```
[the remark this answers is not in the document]
natjoe4: Decided to tackle this in another task. TODO added

hossenlopp (approved):
After testing and in person reviews, this looks good.

sarahmcdougall (approved):
All my comments have been addressed and functionality is good! Ready to merge :)

PR #122, opened 2022-04-20
Value compare reason detail

mgramigna:
~Putting up the code diff for preliminary review if anyone wants to look.~

~Will stay in draft until #121 is merged, and then I will update the base branch~

---

# Summary

This PR adds the logic for reporting value-related reasonDetail on applicable measures for gaps in care. It does this by adding logic in reason detail calculation that actually checks the values present on the data found by a retrieve and seeing if it adhered to the range specified by a clause in the ELM represented as a `ValueFilter` from the query filter parser.

# New Behavior

For value filters that come back as a result of #121, any associated retrieves will now have the value filter comparison done. The change in behavior will result in `ValueOutOfRange` reason codes being present where applicable on both positive and negative improvement measures.

An additional piece of new behavior is sorting of the guidance response results, preferring ones that have reasonCodes other than `NotFound` and `Present`, mostly just for readability to the end user (it's better to not hide the guidance responses at the bottom if they're the most detailed ones)

# Code Changes

* Update `GapsReportBuilder` to have handling for value filters
  * Refactored function a bit to handle both positive and negative improvement without duplicating too much code (likely more work that can be done here)
* Add property to queryInfo to indicate if it came from an external clause or directly within a `where`
* Add helper for doing actual value comparison based on camparator
* Add and update tests

# Testing Guidance

A good test case for this is the diabetes measure with a numerator patient (ensure the improvement notation is properly set to `decrease` in the measure bundle, let me know if you need help with this). The resultant gaps bundle will now have value out of range as a reasonCode

### DIFF
--- a/src/helpers/ValueComparisonHelpers.ts
+++ b/src/helpers/ValueComparisonHelpers.ts
@@ -1,0 +1,21 @@
+import { ValueFilterComparator } from '../types/QueryFilterTypes';
+
+/**
+ * Computes truthiness of value comparisons obtained from value filter
+ */
+export function compareValues(actualValue: number, desiredValue: number, comparator: ValueFilterComparator) {
+  switch (comparator) {
+    case 'ge':
+      return actualValue >= desiredValue;
+    case 'gt':
+      return actualValue > desiredValue;
+    case 'le':
+      return actualValue <= desiredValue;
+    case 'lt':
+      return actualValue < desiredValue;
+    case 'eq':
+      return actualValue == desiredValue;
+    default:
+      throw new Error(`Unsupported comparator "${comparator}" when comparing values`);
+  }
+}
--- a/src/helpers/ValueComparisonHelpers.ts
+++ b/src/helpers/ValueComparisonHelpers.ts
@@ -14,7 +14,7 @@ export function compareValues(actualValue: number, desiredValue: number, compara
     case 'lt':
       return actualValue < desiredValue;
     case 'eq':
-      return actualValue == desiredValue;
+      return actualValue === desiredValue;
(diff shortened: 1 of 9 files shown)


src/helpers/ValueComparisonHelpers.ts
```diff
@@ -0,0 +1,21 @@
...
+    case 'ge':
+      return actualValue >= desiredValue;
+    case 'gt':
+      return actualValue > desiredValue;
+    case 'le':
+      return actualValue <= desiredValue;
+    case 'lt':
+      return actualValue < desiredValue;
+    case 'eq':
+      return actualValue == desiredValue;
```
mgramigna, line 17:
```suggestion
      return actualValue === desiredValue;
```

Idiot

mgramigna (commented):
Looks bad

src/helpers/ValueComparisonHelpers.ts
mgramigna: Thanks

hossenlopp (approved):
Just one comment to think about for the future.

src/helpers/ValueComparisonHelpers.ts
```diff
@@ -0,0 +1,21 @@
+import { ValueFilterComparator } from '../types/QueryFilterTypes';
+
+/**
+ * Computes truthiness of value comparisons obtained from value filter
+ */
+export function compareValues(actualValue: number, desiredValue: number, comparator: ValueFilterComparator) {
```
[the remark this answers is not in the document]
mgramigna: Good call, will make a task for this

PR #123, opened 2022-04-26
Group support for AsyncPatientSource - CLI updates

sarahmcdougall:
# Summary
Fqm-execution now supports calculation using cql-exec-fhir AsyncPatientSource when a group Id is specified rather than an explicit list of patient IDs.

## New behavior
Existing calculation functionality should remain unchanged. This PR adds the ability to pass in a FHIR server Url and Group Id for a FHIR Group resource that references patients on the server. For more information on the Group resource structure, see the [Resource Group spec](https://build.fhir.org/group.html).

Note: the new --group-id Cli option is mutually exclusive with the --patient-ids flag, as both accomplish the same goal of populating the list of patient Ids to use in the AsyncPatientSource class in cql-exec-fhir.

## Code changes
* README updates
* CLI updates for new --group-id flag
* Reworking of function calls in CLI to account for new `loadGroupId()` function from cql-exec-fhir

# Testing guidance
* Review and run unit tests with `npm run test`
* Review the README/CLI documentation
* Update package.json to point to the [`group-support` branch of `cql-exec-fhir`](https://github.com/projecttacoma/cql-exec-fhir/pull/7) instead of the main branch (or use npm link) - this will give us the new Group support functionality
* Navigate to deqm-test-server and reset the server with `npm run db:reset`, followed by `npm run connectathon-upload`
* Test with the EXM130 measure. POST the following test Group resource to deqm-test-server: 
```json
{ "resourceType": "Group", "id": "EXM130-patients", "type": "person", "actual": "true", "member": [ { "entity": { "reference": "Patient/denom-EXM130" } }, { "entity": { "reference": "Patient/numer-EXM130" } } ] }
```
* With deqm-test-server running, run `npm run build` in fqm-execution
* Test every command with the --group-id flag. For example, try
```
node build/cli.js reports -m "path/to/measure/bundle" --as-patient-source --fhir-server-url "http://localhost:3000/4_0_1/" --group-id EXM130-patients -s "2019-01-01" -e "2019-12-31” --debug > output.json
```
* While testing all the commands with the --group-id flag, also test them separately with the --patient-ids flag and compare the output. When testing, I replaced `--group-id EXM130-patients` with `--patient-ids denom-EXM130 numerator-EXM130` and saved the output in a new JSON file. Then, I compared the two files in VSCode - you should receive the same output when running the `--group-id` flag with `EXM130-patients` as when you run with `--patient-id` and the patient IDs that are referenced in the group.
* Test that other CLI logic works. For example, 
* test that if `--patient-bundles` is provided, the output is correct
* test that if `--patient-bundles` and `--group-id` are both provided,  an error is thrown
* test that an error is thrown from `cql-exec-fhir` if the specified group Id cannot be found on the server
* test that an error is thrown if both `--patient-id` and `--group-id` are provided

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ Options:
   --patient-ids <ids...>                      A list of patient ids an AsyncPatientSource will use to query a fhir server for patient data. Note: cannot be used with --patient-bundles or --group-id; --as-patient-source and --fhir-server-url are required when --patient-ids is provided.
   --group-id <id>                             A group id an AsyncPatientSource will use to query a fhir server for patient data. Note: cannot be used with --patient-bundles or --patient-ids; --as-patient-source and --fhir-server-url are required when --group-id is provided.
   --as-patient-source                         Load bundles by creating cql-exec-fhir PatientSource to pass into library calls.
-  --fhir-server-url <url>                     Loads bundles into an AsyncPatientSource which queries the passed in FHIR server URL for patient data. Note: --as-patient-source and --patient-ids (or --group-id) are required when --fhir-server-url is provided.
+  --fhir-server-url <url>                     Loads bundles into an AsyncPatientSource which queries the passed in FHIR server URL for patient data. Note: --as-patient-source and either --patient-ids or --group-id are required when --fhir-server-url is provided.
   -s, --measurement-period-start <date>       Start date for the measurement period, in YYYY-MM-DD format (defaults to the start date defined in the Measure, or 2019-01-01 if not set.
                                               there)
   -e, --measurement-period-end <date>         End date for the measurement period, in YYYY-MM-DD format (defaults to the end date defined in the Measure, or 2019-12-31 if not set there).
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ The options that we support for calculation are as follows:
 | calculateHTML | boolean | yes | Include HTML structure for highlighting. Defaults to false. |
 | vsAPIKey | string | yes | API key, to be used to access a valueset API for downloading any missing valuesets |
 | useValueSetCaching | boolean | yes | Whether to cache valuesets obtained by an API on the filesystem |
-| profileValidation | boolean | yes |  To "trust" the content of meta.profile as a source of truth for what profiles the data that cql-exec-fhir grabs validates against|
+| profileValidation | boolean | yes | To "trust" the content of meta.profile as a source of truth for what profiles the data that cql-exec-fhir grabs validates against|
 
 ### CLI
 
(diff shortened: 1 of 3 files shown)


src/cli.ts
```diff
@@ -116,25 +120,62 @@ async function calc(
...
+    if (program.asPatientSource) {
+      let patientSource;
+      if (program.fhirServerUrl) {
+        if (!program.patientIds && !program.groupId) {
+          console.error(
+            'Must provide an array of patient ids with --patient-ids flag or group id with --group-id flag for calculation using AsyncPatientSource'
+          );
+        }
+        if (program.patientIds && program.groupId) {
+          console.error('Cannot provide both --patient-ids and --group-id flags.');
```
natjoe4, line 158: I wonder if console.errors here is the best method for error reporting. The message quickly gets swallowed up by the output of the calculation, which can be unpredictable if the input params are wrong. Thoughts on throwing an error instead @mgramigna ? Could be a separate task also since this is not the only place we console.error in this file.

natjoe4 (changes requested):
Great work! A couple small comments

src/cli.ts
mgramigna: Strategy that I've used in this file in the past is, in cases where we encounter an error that should halt execution of the cli entirely, use a combination of `console.error` and `program.help()`. The latter will print the usage of the cli to stdout and exit the process, making the user try again

README.md
```diff
@@ -114,10 +114,11 @@ Options:
...
+  --patient-ids <ids...>                      A list of patient ids an AsyncPatientSource will use to query a fhir server for patient data. Note: cannot be used with --patient-bundles or --group-id; --as-patient-source and --fhir-server-url are required when --patient-ids is provided.
+  --group-id <id>                             A group id an AsyncPatientSource will use to query a fhir server for patient data. Note: cannot be used with --patient-bundles or --patient-ids; --as-patient-source and --fhir-server-url are required when --group-id is provided.
   --as-patient-source                         Load bundles by creating cql-exec-fhir PatientSource to pass into library calls.
-  --fhir-server-url <url>                     Loads bundles into an AsyncPatientSource which queries the passed in FHIR server URL for patient data. Note: --as-patient-source and --patient-ids are required when --fhir-server-url is provided.
+  --fhir-server-url <url>                     Loads bundles into an AsyncPatientSource which queries the passed in FHIR server URL for patient data. Note: --as-patient-source and --patient-ids (or --group-id) are required when --fhir-server-url is provided.
```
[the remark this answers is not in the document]
sarahmcdougall: I agree, I struggled to find a nice way to list out everything haha. This is much better!

natjoe4 (approved):
Looks great!

PR #125, opened 2022-04-29
Add -o/--out-file flag to CLI

sarahmcdougall:
# Summary
Adds new `-o/--out-file` option to the fqm-execution CLI.

## New behavior
Now, when a user specifies `-o/--out-file` with a file path, fqm-execution will write the contents of the calculation results to that file instead of console.logging it to stdout. When this happens, a message will be logged to the user with the name of the file path.

If the user specifies the `-o/--out-file` flag but does not specify a file path, the CLI will default to using `output.json` as the file path.

If the user does not use this flag, the results will instead be console.logged to stdout.

## Code changes
Changes were made to the CLI to add the `-o/--out-file` flag and to add a function `writeToFile()` that takes in the file path and a results string and writes the results to the file. The `writeToFile()` function is called after the results have been calculated. If the flag was not supplied, then the results will be logged to stdout like before.

Changes were made to the README to reflect these changes.

# Testing guidance
Run `npm run build`, then try the following:
* Run a command using `-o/--out-file` with a file path —> check that the file was created and that the appropriate results are in the file
* Run a command using `-o/--out-file` without a file path —> the results should be written to `output.json`
* Run a command without this flag —> the output should be logged to the console
* Run a command using `-o/--out-file` and also add a few console.log statements that will be hit during calculation (can add a few in `calculate()` in `src/Calculator.ts` —> the calculation results should appear in the newly created file, and the console.log statements should be logged to the console and they should not be present in the new file

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -79,8 +79,7 @@ program
   )
   .option(
     '-o, --out-file [file-path]',
-    'Path to a file that fqm-execution will write the calculation results to (default: output.json)',
-    'output.json'
+    'Path to a file that fqm-execution will write the calculation results to (default: output.json)'
   )
   .parse(process.argv);
 
@@ -277,8 +276,14 @@ populatePatientBundles().then(async patientBundles => {
     }
 
     if (program.outFile) {
-      writeToFile(program.outFile, JSON.stringify(result?.results, null, 2));
+      if (program.outFile === true) {
+        // use default output.json since no file path was provided
+        writeToFile('output.json', JSON.stringify(result?.results, null, 2));
+      } else {
+        writeToFile(program.outFile, JSON.stringify(result?.results, null, 2));
+      }
     } else {
+      // log results to stdout instead of file
       console.log(JSON.stringify(result?.results, null, 2));
     }
   } catch (error) {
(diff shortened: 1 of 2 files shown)


natjoe4 (changes requested):
This looks great. However, adding a default argument for the -o option looks like it made it so -o is always defined, which means we'll never get stuff printed to console, even when we omit it. Instead, I would set no default value, then check if the value of program.outFile is truthy later on and set the default value there. Happy to talk this through if that's confusing.

sarahmcdougall:
Good catch @natjoe4 , I did that originally and then changed it to have a default value and I forgot to test whether we could still do the console.logs

src/cli.ts
```diff
@@ -264,7 +275,17 @@ populatePatientBundles().then(async patientBundles => {
       });
     }
 
-    console.log(JSON.stringify(result?.results, null, 2));
+    if (program.outFile) {
```
natjoe4, line 278: Nitpicking, but I would change this nested if to just an 
```
if(program.outFile === true){...}
else if(program.outFile){...}
else{...}
```
with a couple comments to clear up what the conditions represent

natjoe4 (changes requested):
Just another small change. Functionality works great

natjoe4 (approved):

PR #126, opened 2022-05-10
SDE bug fix

sarahmcdougall:
# Summary
Fixes a bug in fqm-execution where SDE calculation breaks if the `rawResult` is undefined as a result of failing the `meta.profile` check when the `--profile-validation` flag is used in calculation.

## New behavior
Previously, when using the `--profile-validation` flag with a patient resource that fails the `meta.profile` check, an error would be thrown because the patient does not get retrieved from `cql-exec-fhir`, and thus the SDE results are undefined.

Now, instead of a TypeError being thrown when trying to work through the undefined SDEs, the MeasureReportBuilder will check if the SDE raw results are defined before proceeding. Thus, only SDEs with defined results will be added to the final measure report.

## Code changes
Changes were made to the `MeasureReportBuilder`, as this is where the error gets thrown (gets thrown from `addSDE` when the SDE results are undefined). Adds a conditional to check that the SDE results are defined - if not, then the observation will not be pushed onto the measure report.

Adds unit test for the case where the SDE results are undefined.

# Testing guidance
Run unit tests.

Replicate this scenario by testing the following command:
```
fqm-execution reports -m <path to CMS122 Diabetes measure bundle> -p <path to CMS122 Diabetes numerator patient> -s 2021-01-01 -e 2021-12-31 —profile-validation
```

Previously, you would run into the following error:
```
Cannot use 'in' operator to search for 'forEach' in undefined
TypeError: Cannot use 'in' operator to search for 'forEach' in undefined
…
```

Now, you should not run into any errors. Look at the resulting measure report. The SDE Payer should be included in the report (since it is an empty array but not undefined). There should be no other SDEs in the report. You can try testing without the `--profile-validation` flag and you will see that all the SDEs appear in the final report, as expected.

### DIFF
--- a/src/calculation/MeasureReportBuilder.ts
+++ b/src/calculation/MeasureReportBuilder.ts
@@ -287,60 +287,62 @@ export default class MeasureReportBuilder<T extends PopulationGroupResult> {
   private addSDE(result: ExecutionResult<T>) {
     // 	Note that supplemental data are reported as observations for each patient and included in the evaluatedResources bundle. See the MeasureReport resource or the Quality Reporting topic for more information.
     result.supplementalData?.forEach(sd => {
-      const observation = <fhir4.Observation>{};
-      observation.resourceType = 'Observation';
-      observation.code = { text: sd.name };
-      observation.id = uuidv4();
-      observation.status = 'final';
-      observation.extension = [
-        {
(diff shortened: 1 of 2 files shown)


mgramigna (approved):
works

PR #130, opened 2022-06-02
Update TSC config to target ES2017

hossenlopp:
# Summary
Updates the TS compiler config to target ES2017. This allows for native async/await to be used.

## New behavior
More speed.

## Code changes
None.

# Testing guidance
Run tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -776,10 +776,11 @@
       }
     },
     "@lhncbc/ucum-lhc": {
-      "version": "4.1.4",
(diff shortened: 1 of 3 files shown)


mgramigna:
I wonder if we should wait to do this until async is merged to cqf/master. The rebase of our fork will be a lot of work for code that will ideally be merged into cqf in the not so distant future

hossenlopp:
This was merged into cqframework/master so we should be fine just getting this with the rebase.

PR #132, opened 2022-07-27
Detailed obs

lmd59:
# Summary
Solves two problems:
1. The PopulationResult observations should be populated (with a number in this case, but could be a different type in other cases). Find where the PopulationResult (attribute of EpisodeResults) is populated, understand why the observation attribute is not populated, and populate with the appropriate number (which should be available from the raw results).
->This was fixed by reusing code to find the observation's associated population when one isn't specified.

2. For this measure, there are actually two observations: one associated with the numerator, and one associated with the denominator. When the detailed results are created, the populationType attribute is used to identify the PopulationResult, but because there are two PopulationResult's that are of type "measure-observation", there is know way to tell the difference between these results. We need to use some other way of identifying populations (probably an id or identifier?).
-> This was addressed using the cql identifier in the criteriaExpression. This identifier information also allowed for observation results to be appropriately associated with the correct measure-observation population.

## New behavior
Detailed results now have appropriately associated observation values in episodeResults

## Code changes

- `criteriaExpression` has been added as a required attribute to the `PopulationResult` type in Calculator.ts. This has resulted in changes to assign the value appropriately in ClauseResultsBuilder, DetailedResultsBuilder, and all of the top level testing files changed.
- Added `findObsMsrPopl` helper function to DetailedResultsHelpers and corresponding unit testing. This function finds a measure population for a given group and if not found, does special handling of ratio measures without specified populations for the observations. Function used in DetailedResultsBuilder and Execution.
- The `criteriaExpression` is also used in the DetailedResultsBuilder to differentiate between different observation populations as the population type is not enough alone to tell what population the result should be assigned to.

# Testing guidance
Run: `npm run test`
You can test this with CMS 871
Bundle directory: https://github.com/cqframework/ecqm-content-r4-2021/tree/master/bundles/measure/HospitalHarmHyperglycemiainHospitalizedPatientsFHIR
The numerator patient (and I think any other patient chosen from this folder) needs to be updated so that all of its datetimes use milliseconds (add `.0` to the end) to avoid an interval comparison problem. Feel free to reach out if you'd like me to just drop you my updated patient file.

VSCode launch: `
        "args": ["detailed","--debug","-m", "${workspaceFolder}/test/fixtures/connectathon/HospitalHarmHyperglycemiainHospitalizedPatientsFHIR-bundle.json", "-p", "${workspaceFolder}/test/fixtures/connectathon/tests-numer-EXM871-bundle.json", "-s", "20190101", "-e", "20191231", "-o", "test-output.json"]`

Result:
The output detailedResults file should have episodeResults with populationResults where a 9 is associated with the denominator observations and a 1 is associated with the numerator observations.

### DIFF
--- a/src/calculation/ClauseResultsBuilder.ts
+++ b/src/calculation/ClauseResultsBuilder.ts
@@ -596,6 +596,7 @@ export function buildPopulationRelevanceForAllEpisodes(
     populationGroup.population?.map(population => {
       return <PopulationResult>{
         populationType: MeasureBundleHelpers.codeableConceptToPopulationType(population.code),
+        criteriaExpression: population.criteria.expression || 'Unknown',
         result: false
       };
     }) || []; // Should not end up becoming an empty list.
@@ -635,6 +636,7 @@ export function buildPopulationRelevanceMap(results: PopulationResult[]): Popula
   const relevantResults: PopulationResult[] = results.map(result => {
     return {
       populationType: result.populationType,
+      criteriaExpression: result.criteriaExpression,
       result: true
     };
   });
@@ -772,6 +774,7 @@ export function setResult(populationType: PopulationType, newResult: boolean, re
 // create a result for the given population type and result or update the existing value to true if newResult is true
 export function createOrSetResult(
   populationType: PopulationType,
+  criteriaExpression: string,
   newResult: boolean,
   results: PopulationResult[]
 ): void {
@@ -783,6 +786,7 @@ export function createOrSetResult(
   } else {
     results.push({
       populationType,
+      criteriaExpression,
       result: newResult
     });
   }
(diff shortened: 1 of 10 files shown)


mgramigna (changes requested):
Looks good and works! I had a tiny comment and then a question:

Question -- given that a Measure group's population criteria expression might be undefined, should we represent that here? I noticed that since you made the type of it only `string`, in cases where it may not exist we set it to `'Unknown'`. Is this better than just allowing criteriaExpression to be `string | undefined` instead? lmk

lmd59:
> Looks good and works! I had a tiny comment and then a question:
> 
> Question -- given that a Measure group's population criteria expression might be undefined, should we represent that here? I noticed that since you made the type of it only `string`, in cases where it may not exist we set it to `'Unknown'`. Is this better than just allowing criteriaExpression to be `string | undefined` instead? lmk

I think that makes sense? For measure's we're working with, this is a cql population identifier, so it really should exist on all of them I'd think, but maybe there are cases that I'm not thinking of. Either way, the criteriaExpression needs to be assigned in some way for that exception case, and undefined might make more sense than 'Unknown'.

mgramigna (approved):
Great job

PR #134, opened 2022-08-15
Fhir query fix

elsaperelli:
# Summary
This PR fixes a bug found in fqm-execution and ensures that our fhir query pattern extensions are generating valid search parameters. Before, we were using the PatientReferences.json lookup which is a map of resoureType -> **attribute path**, but we really want a lookup of resourceType -> **search parameter**. 

## New behavior
ParatientParameters.json is now used for search parameter lookup instead of PatientReferences.json. Tests were modified for this change and added to include some missing cases. All other behavior remains the same.

## Code changes
- `DataRequirementsHelpers.ts`: Changed the JSON lookup file from `PatientReferences.json` to `PatientParameters.json`.
- `PatientParameters.json`: This JSON lookup maps resourceType to the search parameters of each data requirement type. This was the output of the script `parseCompartmentDefinition.js` in the `deqm-test-server` repository. 
- `DataRequirementHelpers.test.ts`: Modified tests that involved the fhir query pattern extension as well as added some use cases that were previously incorrect. 

# Testing guidance
1. Run `npm run test`
2. You can run this with any measure bundle but I recommend one in the connectathon repository: [https://github.com/DBCG/connectathon]
3. Run the `cli.ts` file: `./src/cli.ts -m <path to bundle.json> -o`
4. Inspect the output to make sure all of the fhir query pattern extensions are generating valid search parameters. For example, there should not be any camelCase search params but rather kebab-case. In addition, the parameter for Procedure resources should be performer not performer.actor.


sarahmcdougall (changes requested):
Wooo good work addressing this bug! Aside from the in-line comments, we should make sure that we delete `src/compartment-definition/PatientReferences.ts` since it isn't used in fqm-execution anymore (it was only used for the fhirQueryPattern code that we are changing in this PR)

sarahmcdougall (approved):

natjoe4 (approved):
Nice work :)

PR #136, opened 2022-09-06
Add ratio measure score support

sarahmcdougall:
# Summary
Adds measure score to individual and summary ratio measure reports.

## New behavior
When running calculation on a ratio measure, an individual report will now include the measure score for each patient, where the measure score is calculated as: `numerator measure observation / denominator measure observation`. 

## Code changes
Most changes occur in the `MeasureReportBuilder`. Uses the aggregate method for the numerator and denominator observations when calculating the measure score.

# Testing guidance
Test with the [EXM871 bundle from the ecqm-content-r4-2021 repo](https://github.com/cqframework/ecqm-content-r4-2021/tree/master/bundles/measure/HospitalHarmHyperglycemiainHospitalizedPatientsFHIR). 
Some tests to try out include:
* Test with the denominator patient - the final measure score should be 0 since there is no numerator observation
* Test with the numerator patient - the final measure score should be 0.1111 (or 1/9) since the numerator observation value is 1 and the denominator observation value is 9
* Test with both the denominator and numerator patient, and generate a summary report - the measure score should be 0.0833 (or 1/12) since the total numerator observation is 0+1 = 1, and the total denominator observation is 3+9=12

NOTE: When testing, you may run into CQL-related errors with these patient bundles. To fix these errors, you will need to edit the patient resource files locally so that the date times end in “.0”

### DIFF
--- a/src/calculation/MeasureReportBuilder.ts
+++ b/src/calculation/MeasureReportBuilder.ts
@@ -1,10 +1,4 @@
-import {
-  ExecutionResult,
-  CalculationOptions,
-  PopulationResult,
-  PopulationGroupResult,
-  DetailedPopulationGroupResult
-} from '../types/Calculator';
+import { ExecutionResult, CalculationOptions, PopulationResult, PopulationGroupResult } from '../types/Calculator';
 import { PopulationType, MeasureScoreType, AggregationType } from '../types/Enums';
 import { v4 as uuidv4 } from 'uuid';
 import { extractMeasureFromBundle } from '../helpers/MeasureBundleHelpers';
@@ -179,7 +173,7 @@ export default class MeasureReportBuilder<T extends PopulationGroupResult> {
             group.measureScore = this.calcMeasureScoreCV(this.measure, groupResults, group.id || '');
             this.addScoreObservation(group.measureScore, this.measure, this.report);
           } else if (this.scoringCode === MeasureScoreType.RATIO) {
-            const measureScore = this.calcMeasureScoreRatio(groupResults);
+            const measureScore = this.calcMeasureScoreRatio(this.measure, groupResults, group.id || '');
             if (this.isIndividual) {
               group.measureScore = measureScore;
             }
@@ -206,7 +200,11 @@ export default class MeasureReportBuilder<T extends PopulationGroupResult> {
                       stratResults.strataCode
                     );
                   } else if (this.scoringCode === MeasureScoreType.RATIO) {
-                    stratum.measureScore = this.calcMeasureScoreRatio(groupResults, stratResults.strataCode);
+                    stratum.measureScore = this.calcMeasureScoreRatio(
+                      this.measure,
+                      groupResults,
+                      stratResults.strataCode
+                    );
                   }
                 } else {
                   throw new UnexpectedProperty(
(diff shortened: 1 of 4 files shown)


src/calculation/MeasureReportBuilder.ts
```diff
@@ -393,6 +415,116 @@ export default class MeasureReportBuilder<T extends PopulationGroupResult> {
...
+
+    // Calculate as (NUMER OBSERVATION) / (DENOM OBSERVATION)
+    const numeratorCount = observations
+      .filter(obs => obs.criteriaExpression === 'Numerator Observations')
+      ?.map(o => o.observation)
+      .reduce((accumulator, currentValue) => accumulator.observation + currentValue.observation);
+    const denominatorCount = observations
+      .filter(obs => obs.criteriaExpression === 'Denominator Observations')
+      ?.map(o => o.observation)
+      .reduce((accumulator, currentValue) => accumulator.observation + currentValue.observation);
```
mgramigna, line 461: Discussed on a call, but should be respecting the `cqfm-aggregateMethod` extension defined on the populations

src/calculation/MeasureReportBuilder.ts
```diff
@@ -550,8 +671,8 @@ export default class MeasureReportBuilder<T extends PopulationGroupResult> {
     });
   }
 
-  public getReport(): fhir4.MeasureReport {
-    this.calculateGroupScores();
+  public getReport(results?: ExecutionResult<PopulationGroupResult>[]): fhir4.MeasureReport {
```
hossenlopp, line 727: The method signature for this should not have been changed. This is intended to be called after all patients (or single) have had their results added with `addPatientResults`. This function should just be completing the MeasureReport and returning it, not taking any additional data.

src/calculation/MeasureReportBuilder.ts
```diff
@@ -497,27 +629,12 @@ export default class MeasureReportBuilder<T extends PopulationGroupResult> {
...
-          // TODO: what if value for denominator 0? ... do we need to subtract denex, dexecep... probably, as https://ecqi.healthit.gov/system/files/eCQM-Logic-and-Guidance-2018-0504.pdf
-          value: denominatorCount2 === 0 ? 0 : (numeratorCount2 / denominatorCount2) * 1.0
-        };
       default:
         throw new UnsupportedProperty(`Measure score type \"${scoringCode}\" not supported`);
     }
   }
 
-  private calculateGroupScores() {
+  public calculateGroupScores(results?: ExecutionResult<PopulationGroupResult>[]) {
```
hossenlopp, line 686: This function should be kept private and only use existing collected state in the report builder.

hossenlopp (changes requested):
Partial review. The external interface to the MeasureReportBuilder should not have been changed for this work.

src/calculation/MeasureReportBuilder.ts
sarahmcdougall: Ah that makes sense - this was the only way that I could come up with for getting all the data needed for computing an aggregate measure score but I agree that it should not be changed. Did we confirm that the "count" for each measure observation should be the observation value rather than the count of how many measure observations there are? Under that assumption, I think some refactoring could be done that would make it easier to access all the necessary data for calculating the aggregate measure score. If not, do you know how to get all the observations for all patients in one place for calculating the aggregate measure score that does not require bringing in the entire `ExecutionResult`?

hossenlopp: I think for this specific measure it will be the sum of all measures observations across all patients. So I think you could add a field to the MeasureReportBuilder instance that is used to hold on to or progressively count all of the observations for the numerator and denominator as each patient result is added. Then when getReport is called it can use that to make the overall score from that instance data instead of using the ExecutionResults.

mgramigna (approved):
¯\\\_(ツ)_/¯ works

hossenlopp (approved):

PR #137, opened 2022-09-08
Fix and Refactor Recursive ELM Parsing

mgramigna:
# Summary

This PR fixes issues uncovered in our RetrievesFinder where we were not handling all parts of a query that could contain an expression. This PR updates the recursion to handle the following cases, per the [Logical Model for a Query](http://cql.hl7.org/04-logicalspecification.html#queries):

1. `Query.source` (already handled before this PR but including it in the list for completeness)
2. `Query.let.expression`
3. `Query.where`
4. `Query.return.expression`
5. `Query.relationship.expression` (typically a `With` or `Without` expression)
6. `Query.relationship.suchThat.expression`

(Technically, `Query.aggregate` could have expressions, but we have never seen that used in an eCQM before, so couldn't properly draft a test case for it, and we are ignoring it for now)

Any of the above can be any ELM expression, and therefore _could_ be an ELM `Retrieve` expression, regardless of how unlikely that is in the measure logic. To see examples of how usage of retrieves look in this expressions in CQL, see the [unit test fixture](https://github.com/projecttacoma/fqm-execution/blob/cefa2bf0a447cd8480663f28a6b2ef673ef74d9e/test/fixtures/elm/queries/ComplexQueries.cql#L67) included in this PR. For all of these test cases, we expect all the data types and ValueSets used within these CQL statements to show up in the data requirements. That is what this PR addresses.

This PR also does a refactor to the `findRetrieves` function. There was a lot of duplicate recursive calls and results logging that I moved to a helper function with smart defaults to reduce code duplication and allow for more readability of the code given the fact that the developer now needs to explicitly specify which properties are changing during the recursive call.

Also fixed exponential growth of `withErrors` array.

## New behavior

Data types and ValueSets that appear at deeper nesting levels of a query via the above 6 cases are now included in data requirements output

## Code changes

* Add recursive calls for above 6 cases in `findRetrieves`
  * Here: https://github.com/projecttacoma/fqm-execution/pull/137/files#diff-3cfc16e9767c6fe6b87fa1f233e9ab7fb01a4dd9ba9f5a70d49dfd152a430f3dR183
* Add new unit tests and unit test CQL cases for all the above
* Remove shallow copying of `withErrors` array with each recursive call to avoid exponential growth for no reason
* Attached local IDs to graceful error messages for readability
* Refactor `findRetrieves` to use a helper function for all recursive calls (helps readability)
* Refreshed local IDs in unrelated unit test given that the translator updated them after the unit test fixture changed (expected)

# Testing guidance

EXM506 is a good test case for this. Compare the data requirements results with master and this branch, and you should see diffs picking up new resources that are further embedded in queries that weren't before :). Here is a suggestion for how to view these diffs:

1. Checkout `master` branch
2. run data requirements for EXM506 on master branch:

``` bash
./src/cli.ts dataRequirements -m path/to/connectathon/EXM506-2.2.000/EXM506-2.2.000-bundle.json -o master-dr.json
```
3. Checkout this `additional-query-parsing` branch
4. Run the same command but to a different output file:

``` bash
./src/cli.ts dataRequirements -m path/to/connectathon/EXM506-2.2.000/EXM506-2.2.000-bundle.json -o new-dr.json
```
5. Look at the diffs:

```bash
diff master-dr.json new-dr.json
```

You should see additions of Conditions, ServiceRequests, Procedures, and MedicationRequests. You can cross-reference the CQL to see why these weren't there before and why they are now


elsaperelli (approved):
Looks great to me! Only one typo :)

package.json
```diff
@@ -50,8 +50,8 @@
     "lint:fix": "tsc --noEmit && eslint \"**/*.{js,ts}\" --quiet --fix",
     "prettier": "prettier --check \"**/*.{js,ts}\"",
     "prettier:fix": "prettier --write \"**/*.{js,ts}\"",
-    "test": "jest --silent",
-    "test:coverage": "jest --collectCoverage --silent",
+    "test": "jest",
+    "test:coverage": "jest --collectCoverage",
```
natjoe4, line 54: Just double checking this was an intentional change...

src/gaps/RetrievesFinder.ts
```diff
@@ -135,57 +175,53 @@ export function findRetrieves(
...
+    // NOTE: ignoring "aggregate" for now, as we have no precedent for its use within queries for eCQMs
+    const query = expr as ELMQuery;
+
+    // Overwrite queryLocalId with new ID of the current query expression
+    const recursiveOpts: RecursiveCallOptions = {
+      ...defaultRecursiveOpts,
+      newQueryLocalId: query.localId
+    };
+
+    query.source?.forEach(s => {
```
natjoe4, line 191:
```suggestion
    query.source.forEach(s => {
```
I think we can ditch this question mark since source is 1...*

src/gaps/RetrievesFinder.ts
```diff
@@ -135,57 +175,53 @@ export function findRetrieves(
...
+      recurse(results, s.expression, recursiveOpts);
+    });
+
+    query.let?.forEach(letClause => {
```
natjoe4, line 195: We should make let an optional property in ElmQuery since its 0...*

natjoe4 (changes requested):
Just some tiny changes! Looks great. Love the clean look of the recurse function. Nice job!

package.json
mgramigna: haha yep it was, should've done this a while ago when we removed console logs from the source code

natjoe4 (approved):
Looks good to me!

PR #139, opened 2022-09-21
Cache ElmJsons

natjoe4:
# Summary
Added optional caching for cqls, elmJSONs, rootLibIdentifier, codeService, rep, vsMap when running execution

## New behavior
When passed the `useElmJsonsCaching` option, execution will check the existing ELMInfoCache to see if a measure with the same id and version has already been run in the last 10 minutes. It will use the previously extracted results for `cqls, elmJSONs, rootLibIdentifier, codeService, rep, vsMap` during the execution if the results exist and are valid.

## Code changes
- New file `ElmInfoCache` storing the cache, and functions to retrieve info from it as well as clear it
- Added `useElmJsonsCaching` option on calculation options to enable caching
- Added 'clearElmJsonsCache` option to force clear the existing ElmInfoCache before calculation
- New cli option `cache-elm` to enable ElmJsons caching
- Added testing

# Testing guidance
- `npm run check`
- Run a measure report calculation with the `cache-elm` option set to true and ensure that behavior is unchanged
- From `fqm-execution`, run `npm link`
- Open fqm-testify and run `rm -rf node_modules/fqm-execution` then `npm link fqm-execution`
- In fqm-testify in `util/MeasureCalculation.ts`, add `useElmJsonsCaching: true` following to the `options` const on line 15
- `npm run dev`
- Upload a measure
- Create a patient, select and ensure calculation results are correct
- Create a test resource and ensure calculation results are correct
- Use the chrome profiler to see the reduction in calculation time when caching is enabled. If you are unsure how to do this, reach out to me and I can walk you through!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,27 +24,27 @@
       }
     },
     "@babel/compat-data": {
-      "version": "7.18.13",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.18.13.tgz",
-      "integrity": "sha512-5yUzC5LqyTFp2HLmDoxGQelcdYgSpP9xsnMWBphAscOdFrHSAVbLNzWiy32sVNDqJRDiJK6klfDnAgu6PAGSHw==",
+      "version": "7.19.1",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.19.1.tgz",
+      "integrity": "sha512-72a9ghR0gnESIa7jBN53U32FOVCEoztyIlKaNoU05zRhEecduGK9L9c3ww7Mp06JiR+0ls0GBPFJQwwtjn9ksg==",
       "dev": true
     },
     "@babel/core": {
-      "version": "7.18.13",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.18.13.tgz",
-      "integrity": "sha512-ZisbOvRRusFktksHSG6pjj1CSvkPkcZq/KHD45LAkVP/oiHJkNBZWfpvlLmX8OtHDG8IuzsFlVRWo08w7Qxn0A==",
+      "version": "7.19.1",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.19.1.tgz",
+      "integrity": "sha512-1H8VgqXme4UXCRv7/Wa1bq7RVymKOzC7znjyFM8KiEzwFqcKUKYNoQef4GhdklgNvoBXyW4gYhuBNCM5o1zImw==",
       "dev": true,
       "requires": {
         "@ampproject/remapping": "^2.1.0",
         "@babel/code-frame": "^7.18.6",
-        "@babel/generator": "^7.18.13",
-        "@babel/helper-compilation-targets": "^7.18.9",
-        "@babel/helper-module-transforms": "^7.18.9",
-        "@babel/helpers": "^7.18.9",
-        "@babel/parser": "^7.18.13",
+        "@babel/generator": "^7.19.0",
+        "@babel/helper-compilation-targets": "^7.19.1",
+        "@babel/helper-module-transforms": "^7.19.0",
+        "@babel/helpers": "^7.19.0",
+        "@babel/parser": "^7.19.1",
         "@babel/template": "^7.18.10",
-        "@babel/traverse": "^7.18.13",
-        "@babel/types": "^7.18.13",
+        "@babel/traverse": "^7.19.1",
+        "@babel/types": "^7.19.0",
         "convert-source-map": "^1.7.0",
         "debug": "^4.1.0",
         "gensync": "^1.0.0-beta.2",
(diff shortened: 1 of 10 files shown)


package-lock.json
```diff
@@ -1185,7 +1185,7 @@
     "arr-diff": {
       "version": "4.0.0",
       "resolved": "https://registry.npmjs.org/arr-diff/-/arr-diff-4.0.0.tgz",
-      "integrity": "sha512-YVIQ82gZPGBebQV/a8dar4AitzCQs0jjXwMPZllpXMaGjXPYVUawSxQrRsjhjupyVxEvbHgUmIhKVlND+j02kA==",
+      "integrity": "sha1-1kYQdP6/7HHn4VI1dhoyml3HxSA=",
```
mgramigna, line 1188: same integrity issue as before, try the steps and see if they work for not updating the integrity:

```
rm -rf node_modules
npm cache clear --force
rm package-lock.json
npm install
```

hopefully should be no diffs in the integrity after this? Should all be 512s

src/execution/Execution.ts
```diff
@@ -58,35 +55,17 @@ export async function execute(
   // These can be provided directly as an argument or via the caching behavior
   valueSets.push(...valueSetCache);
 
-  const vsMap = valueSetsForCodeService(valueSets);
-
-  const { cqls, rootLibIdentifier, elmJSONs } = MeasureBundleHelpers.extractLibrariesFromBundle(measureBundle);
-
+  //const vsMap = valueSetsForCodeService(valueSets);
```
mgramigna, line 58: assuming this line isn't needed anymore
```suggestion
```

src/helpers/elm/ELMInfoCache.ts
```diff
@@ -0,0 +1,93 @@
...
+    const rep = new Repository(elmJSONs);
+    if (useElmJsonsCaching) {
+      ELMInfoCache.set(key, { cqls, elmJSONs, rootLibIdentifier, codeService, rep, vsMap, lastAccessed: new Date() });
+    } else {
+      return { cqls, elmJSONs, rootLibIdentifier, codeService, rep, vsMap };
+    }
+  } else {
+    ELMInfoCache.set(key, { ...(ELMInfoCache.get(key) as ELMInfoCacheType), lastAccessed: new Date() });
+  }
+  return { ...(ELMInfoCache.get(key) as ELMInfoCacheType) };
```
mgramigna, line 75: is there a reason that we return a new object with a destructured object inside? Can we not just return `ELMInfoCache.get(key) as ELMInfoCacheType`? 

Also I think I see why the type casting is necessary but it took me a second. Can you add a comment just saying the cast is used since we know that the cache entry is guaranteed to be set or the function will have already returned by the time we get to this return statement?

src/types/Calculator.ts
```diff
@@ -42,6 +42,10 @@ export interface CalculationOptions {
   verboseCalculationResults?: boolean;
   /** if true trust the content of meta.profile as a source of truth for what profiles the data that cql-exec-fhir grabs validates against */
   trustMetaProfile?: boolean;
+  /** if true, cache ELM JSONs and associated data for access in subsequent runs within 10 minutes */
+  useElmJsonsCaching?: boolean;
+  /** if true, clears ElmJsons cache before running calculation */
+  clearElmJsonsCache?: boolean;
```
mgramigna, line 48: Can you update the README adding information about these new calculation options?

mgramigna (changes requested):
Sweet! I was able to get this working in testify and could see the proper behavior of extracting it once, then using the cache for subsequent calculations. Very cool. I have a few inline comments.

src/types/Calculator.ts
natjoe4: ooh good catch!

src/helpers/elm/ELMInfoCache.ts
natjoe4: Yeah I wasn't entirely sure if functions would manipulate this info in ways that could mess up what the cache points to, so thought it safer to just return a copy. If not tho then will def change it to a direct return

src/cli.ts
```diff
@@ -85,6 +85,7 @@ program
     '-o, --out-file [file-path]',
     'Path to a file that fqm-execution will write the calculation results to (default: output.json)'
   )
+  .option('--cache-elm', 'Whether or not to cache the ELM JSONs and associated information from calculation', false)
```
elsaperelli, line 88: Question: do we want an option for clearElmJsonsCache?

elsaperelli (changes requested):
Looks great to me! Just one question and it looks like package-lock still doesn't have the sha512. Other than that looks good!

src/cli.ts
natjoe4: I left out this option because I think the cache should only work when fqm-execution is run as a package as part of a running program, like in fqm-testify. The cli commands will only be run between isolated runs of fqm-execution, so I think the cache will reset regardless. @mgramigna am I right about this? Happy to add a clear cache option if not.

elsaperelli: Ah that makes sense now!!

mgramigna: correct

mgramigna (approved):

src/cli.ts
hossenlopp: I hate to throw a wrench in this PR review. But there isn't a need for the `--cache-elm` option in the CLI at all since it's just an in memory cache. So it will be built fresh every single run.

natjoe4: Lol that's a great point. Will remove

mgramigna (approved):

elsaperelli (approved):
Nice job!

PR #147, opened 2022-10-07
Use search parameters to identify date based query string to add to t…

lmd59:
…he data requirement extension

# Summary

Uses (only date-based) search parameters pulled from [here](https://hl7.org/fhir/search-parameters.json) to identify the code that matches the dateFilter path, and add that code to the dataRequirement query pattern extension.

## New behavior
The dataRequirement extension should have a query string that at times uses a different code specific to the dateFilter path, whereas it had previously defaulted to `'date'`.

## Code changes

- Added SearchParameters.ts to allow code lookup based on the dateFilter path (looks for the expression that contains that path).
- Updated DataRequirementsHelpers.ts to adjust dateFilter path, do lookup from SearchParameters, and add to query string using the appropriate code (and adjusting for the dateFilter value type.
- Updated and added tests for DataRequirementsHelpers `addFhirQueryPatternToDataRequirements` function.

Question: Is the console log acceptable error handling, or should we just ignore these situations and accept leaving off date query information without explicit user notification?

# Testing guidance
`npm run test`
Run data requirements request with various measures. Sample launch.json line for vscode: `"args": ["dataRequirements", "--debug","-m", "${workspaceFolder}/test/fixtures/EXM124-9.0.000-bundle.json", "-s", "20190101", "-e", "20191231"]`
Most measures I tested did had a non-'date' code for the ServiceRequest (path `authoredOn` -> query `authored=`), so this shows a good example to the previous approach, which would have used a `date=` query.

### DIFF
--- a/src/helpers/DataRequirementHelpers.ts
+++ b/src/helpers/DataRequirementHelpers.ts
@@ -166,10 +166,10 @@ export function addFhirQueryPatternToDataRequirements(dataRequirement: fhir4.Dat
       // add resource type
       path = `${dataRequirement.type}.${path}`;
 
-      // identify search paramters that should be used for the query
+      // identify search parameters that should be used for the query
       const foundParams = SearchParameters.entry.filter(searchParam => searchParam.resource.expression.includes(path));
 
-      if (foundParams.length == 1) {
+      if (foundParams.length === 1) {
         if (dateFilter.valueDateTime) {
           queryString = queryString.concat(`${foundParams[0].resource.code}=${dateFilter.valueDateTime}&`);
         } else if (dateFilter.valuePeriod) {
--- a/src/helpers/DataRequirementHelpers.ts
+++ b/src/helpers/DataRequirementHelpers.ts
@@ -157,7 +157,10 @@ export function addFhirQueryPatternToDataRequirements(dataRequirement: fhir4.Dat
   if (dataRequirement.dateFilter) {
     dataRequirement.dateFilter.forEach(dateFilter => {
       let path = dateFilter.path || '';
-      // remove period path pieces - TODO: do we need to use these pieces in our query string?
+      // remove period path pieces
+      // NOTE: Currently we are limited in doing period comparisons within our query pattern
+      // because search parameters do not allow a use of ".end" or ".start"
+      // and prefixes are not sufficient for certain comparisons https://www.hl7.org/fhir/search.html#prefix
       if (path?.endsWith('.end')) {
         path = path.slice(0, -4);
       } else if (path?.endsWith('.start')) {
--- a/src/helpers/DataRequirementHelpers.ts
+++ b/src/helpers/DataRequirementHelpers.ts
@@ -14,7 +14,7 @@ import {
   IsNullFilter
 } from '../types/QueryFilterTypes';
 import { PatientParameters } from '../compartment-definition/PatientParameters';
-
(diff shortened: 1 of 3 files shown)


natjoe4 (commented):
These changes look good and work as described. Nice work! Just a few small comments and I'd look into adding tests to cover the uncovered lines (since it seems like this could be done with minor tweaks to the existing test cases) as well as one that covers the case when a dataRequirement has multiple dateFilters.

src/helpers/DataRequirementHelpers.ts
```diff
@@ -154,15 +154,42 @@ export function addFhirQueryPatternToDataRequirements(dataRequirement: fhir4.Dat
...
+      // remove period path pieces - TODO: do we need to use these pieces in our query string?
+      if (path?.endsWith('.end')) {
+        path = path.slice(0, -4);
+      } else if (path?.endsWith('.start')) {
+        path = path.slice(0, -6);
+      }
+      // add resource type
+      path = `${dataRequirement.type}.${path}`;
+
+      // identify search paramters that should be used for the query
```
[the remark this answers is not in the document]
lmd59: Oh dang- you got me on a spelling mistake!

test/DataRequirementHelpers.test.ts
```diff
@@ -511,7 +511,68 @@ describe('DataRequirementHelpers', () => {
...
+              unit: 'days',
+              value: 10
+            }
+          }
+        ]
+      };
+      DataRequirementHelpers.addFhirQueryPatternToDataRequirements(testDataReqWithDateFilter);
+      expect(testDataReqWithDateFilter.extension?.length).toEqual(2);
+      if (testDataReqWithDateFilter.extension) {
+        expect(testDataReqWithDateFilter.extension[0].valueString).toEqual(
```
natjoe4, line 561: Nitpicking, but I think it's better to just continue to use `?.` chaining here rather than stick these inside an if

test/DataRequirementHelpers.test.ts
```diff
@@ -511,7 +511,68 @@ describe('DataRequirementHelpers', () => {
...
+              start: '2019-01-01',
+              end: '2019-12-31'
+            }
+          }
+        ]
+      };
+      DataRequirementHelpers.addFhirQueryPatternToDataRequirements(testDataReqWithDateFilter);
+      expect(testDataReqWithDateFilter.extension?.length).toEqual(2);
+      if (testDataReqWithDateFilter.extension) {
+        expect(testDataReqWithDateFilter.extension[0].valueString).toEqual(
```
natjoe4, line 532: Same optional chaining up here :)

natjoe4 (changes requested):
Just some tiny code style stuff in the test cases but otherwise ready to approve :)

test/DataRequirementHelpers.test.ts
lmd59: Great call

test/DataRequirementHelpers.test.ts
lmd59: Much cleaner - nice :-)

sarahmcdougall (approved):
Tested with multiple measure bundles and it gave the correct output every time 👍 Good call adding the comment about the period comparisons.

Just needs a rebase, but otherwise should be ready to merge!

natjoe4 (approved):
Nice work! Looks great to me :)

PR #148, opened 2022-10-10
Fix functions incorrectly highlighting as false

### LINKED ISSUE #143: CQL fictions are added in coverage highlighting
Hello,

I think earlier, functions were not added to the coverage(highlighted HTML). After switching to [v1.0.0-beta.4](https://github.com/projecttacoma/fqm-execution/releases/tag/v1.0.0-beta.4), noticed that functions are being added to the HTML highlighting in red color(screenshot attached). I'm not sure if this is an intentional shift from the earlier approach. Even in cqm-execution, we used to exclude functions from coverage.

Screen Shot 2022-10-07 at 11 30 36 AM

Also, in `detailedResults.clauseResults` functions are marked as `NA` however in `detailedResults.statementResults` thier relevance is TRUE/FALSE. I'm using clauseResults to calculate the overall coverage percentage for a measure group. So, this causes some mismatch in coverage percentage and highlighting. In the above screenshot, I've 100% coverage but CQL functions are highlighted in red color. Is this a potential bug or the way I calculate overall coverage for a measure group is incorrect?

Thanks!

mgramigna:
Fixes #143

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "fqm-execution",
-  "version": "1.0.0-beta.4",
+  "version": "1.0.0-beta.5",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fqm-execution",
-  "version": "1.0.0-beta.4",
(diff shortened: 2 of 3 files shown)


hossenlopp (approved):
LGTM

hossenlopp:
Actually looks good and tests fine. Version bump is good.

Issue #144, opened 2022-10-07
Missing results for measure observations in some scenarios


nmorasb:
Expected behavior: 

- Boolean Continuous Variable measure includes observation results 
- Boolean and non-boolean Ratio measures include observation results 

Actual behavior: 

Only non-boolean Continuous Variable measure includes observation results 

populationResults results field only includes a boolean result for observations, and we always expect a numeric value for observation results. So far, we have only seen a numeric observation result in episodeResults, which is only populated for episode based measures. 

For all instances of Ratio, we have not seen any observation results in the episodeResults. The observations field of the measure-observation populationType is always undefined. 

[observation-resources.zip](https://github.com/projecttacoma/fqm-execution/files/9735555/observation-resources.zip)

mgramigna:
Thanks for the issue! And thanks for providing examples as well.

In our initial investigation/exploration with supporting Ratio measures, we did not really have any examples to go off of, so this helps a lot.

For the CV Boolean measure, it looks like the measure scoring is defined using the [cqfm-scoring extension](https://build.fhir.org/ig/HL7/cqf-measures/StructureDefinition-cqfm-scoring.html) on the Measure.group. I believe `fqm-execution` only looks at the root `Measure.scoring` attribute defined in base FHIR to determine the scoring. We will update this, and I believe this will help us fix the issue about observation results not being included in the return result. The raw results from the engine (before they get passed through all of fqm-executions eCQM-specific logic), do include a numerical value on the measure observation function:

```
"obs_func_fun_boolDenom": {
  "episode": true,
  "observation": 1
}
```

So we just need to make sure this gets handled properly when putting it on the detailed results JSON.

---

For the ratio measures, I think they key for us was seeing your use of [cqfm-criteriaReference](https://build.fhir.org/ig/HL7/cqf-measures/StructureDefinition-cqfm-criteriaReference.html) in the measure population for the observation(s) of the ratio measure. With the one example we had, we had to bake some assumptions into fqm-execution, as the example did not use this extension to actually identify which observation was a numerator or a denominator observation. I think this is contributing to the observations field not being properly reported. Our plan is to update `fqm-execution` to use this extension when determining what the observation function is in order to properly report the observation results.

---

Hopefully this helps

nmorasb:
Sounds good, thank you! If any additional examples or info would help with building in support for this then please just let me know!

mgramigna:
@nmorasb I believe #152 resolves these issues, see the PR description for more

[...]

PR #151, opened 2022-10-14
Handle cqfm-scoring Extension

natjoe4:
# Summary
Part 1 of a fix for #144. The `cqfm-scoring` extension allows MeasureGroups to specify the group's scoring code on the group rather than the measure. Prior to this PR, we only looked at the Measure.scoring to determine a group's score. Now fqm-execution prioritizes scoring codes specified on groups via the `cqfm-scoring` extension and falls back to the Measure.scoring when the extension is not present

## New behavior
Measure groups that specify their scoring code via the `cqfm-scoring` extension will now be properly accounted for in the MeasureReportsBuilder. All other functionality should remain unchanged

## Code changes
 - Added code in MeasureReportBuilder.ts to check for presence of `cqfm-scoring` extension, then pass that score through for groups that use it. Default to `Measure.scoring` when absent
 - Added testing

# Testing guidance
 - `npm run check`
 - Testing here is probably most easily done in the debugger. Contact me for an altered MeasureBundle which should trigger the newly added code and fail on master. Place a breakpoint wherever `groupScoringCode` is used in MeasureReportBuilder.ts and ensure that the correct value is retrieved and persisted through for the group in calculation

### DIFF
--- a/src/calculation/MeasureReportBuilder.ts
+++ b/src/calculation/MeasureReportBuilder.ts
@@ -49,10 +49,10 @@ export default class MeasureReportBuilder<T extends PopulationGroupResult> {
   }
 
   private getGroupScoringCode(group: fhir4.MeasureGroup) {
-    const debug = group?.extension?.find(
-      ext => ext.url === 'http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqfm-scoring'
+    return (
+      group?.extension?.find(ext => ext.url === 'http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqfm-scoring')
+        ?.valueCodeableConcept?.coding?.[0].code || this.scoringCode
     );
-    return debug?.valueCodeableConcept?.coding?.[0].code || this.scoringCode;
   }
 
   private setupBasicStructure() {
--- a/src/calculation/MeasureReportBuilder.ts
+++ b/src/calculation/MeasureReportBuilder.ts
@@ -48,6 +48,13 @@ export default class MeasureReportBuilder<T extends PopulationGroupResult> {
     this.setupPopulationGroups();
   }
 
+  private getGroupScoringCode(group: fhir4.MeasureGroup) {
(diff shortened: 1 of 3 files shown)


src/calculation/MeasureReportBuilder.ts
```diff
@@ -434,6 +452,7 @@ export default class MeasureReportBuilder<T extends PopulationGroupResult> {
   // http://build.fhir.org/ig/HL7/cqf-measures/ValueSet-aggregate-method.html
   private aggregate(aggregation: string, observations: number[]) {
     // Note: add break for any non-returning case
+    console.log(aggregation);
```
mgramigna, line 455:
```suggestion
```

test/MeasureReportBuilder.test.ts
```diff
@@ -22,36 +22,22 @@ const patient2Id = '08fc9439-b7ff-4309-b409-4d143388594c';
...
+    entry: [
+      {
+        resource: measure
+      }
+    ]
+  };
+}
+const simpleMeasureBundle: fhir4.Bundle = buildTestMeasureBundle(simpleMeasure);
+const ratioMeasureBundle = buildTestMeasureBundle(ratioMeasure);
+const cvMeasureBundle: fhir4.Bundle = buildTestMeasureBundle(cvMeasure);
```
mgramigna, line 40:
```suggestion
const simpleMeasureBundle = buildTestMeasureBundle(simpleMeasure);
const ratioMeasureBundle = buildTestMeasureBundle(ratioMeasure);
const cvMeasureBundle = buildTestMeasureBundle(cvMeasure);
```

Should prob just use the inference here in all 3 cases

mgramigna (changes requested):
Works quite well with the CV boolean example for measure reports. I was able to get it properly calculating when rebased off of my branch

mgramigna (approved):

elsaperelli (approved):
Nice job!

PR #161, opened 2022-11-02
Add populationId and criteriaReferenceId to all results when present

### LINKED ISSUE #146: Enhancement request: include IDs & criteriaReference on population & episode results output
Would it be possible to include IDs and criteriaReference fields in the execution output for populationResults, episodeResults, and stratificationResults? 

Currently it is difficult to match up observation and stratification output to the inputs for comparison of expected vs actual values. 

Outputs do now include the criteria expression, which helps, but it is possible for users to use the same expression multiple times (for observations). 

Similarly, they could have multiple stratifications on the same population. The outputs are then difficult to map back to the inputs for comparison of expected vs actual values.

mgramigna:
# Summary

Marking this as draft for now since it is branched off of #160. I made that the base branch so the code diff is clean to look at here

Fixes #146 

This PR adds more information to all population and stratifier results when defined on the Measure group

## New behavior

* If a population has an `id`, that `id` will be included on the root populationResults for that population, and on any populationResults for individual episodes in the episodeResults
* If a population references another using `cqfm-criteriaReference`, the `id` of that referenced population (aka the `valueString` of that extension) will be included on all relevant results similar to above
* If a strata in the Measure Group's stratifier has an `id`, that will show up as `strataId` on all stratifier results (including for individual episodes)

## Code changes

* New helper function for adding populationId and criteriaReferenceId to a given populationResult
* Anywhere a populationResult is created, add the relevant IDs using the above function
* Detect when a strata has an ID and ensure it is present on all created stratifier results

# Testing guidance

* Unit tests are the best bet for the stratifier stuff
* For everything else, test detailed results using the following cases (the test cases provided in #152 and #160 might be useful here)
  * Boolean measure with population IDs
  * Boolean measure with a measure observation (uses criteriaReference)
  * Episode measure with population IDs
  * Episode measure with a measure observation (uses criteriaReference)

For the above cases, you should see IDs and criteriaReferenceIds present on the results where appropriate

### DIFF
--- a/src/calculation/ClauseResultsBuilder.ts
+++ b/src/calculation/ClauseResultsBuilder.ts
@@ -606,8 +606,21 @@ export function buildPopulationRelevanceForAllEpisodes(
     masterRelevanceResults.forEach(masterPopResults => {
       // find relevance in episode and if true, make master relevance true, only if not already true
       if (masterPopResults.result === false) {
-        if (getResult(masterPopResults.populationType, episodeRelevance) === true) {
-          masterPopResults.result = true;
+        const matchingEpisodeResult = findResult(
+          masterPopResults.populationType,
+          episodeRelevance,
+          masterPopResults.criteriaExpression
+        );
+        if (matchingEpisodeResult) {
+          if (matchingEpisodeResult.populationId) {
+            masterPopResults.populationId = matchingEpisodeResult.populationId;
+          }
+          if (matchingEpisodeResult.criteriaReferenceId) {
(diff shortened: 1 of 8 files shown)


sarahmcdougall (commented):
Will re-review once this comes out of draft (just to make sure nothing breaks), but for now the only suggestion I have is to consider adding a unit test for the `findPopulationInGroup()` function in `MeasureBundleHelpers.ts` to prevent test coverage from going down. (I don't actually see this function used anywhere but I'm assuming it's just a nice-to-have function that we might want later on? unless my vscode searching is broken lol)

PR #163, opened 2022-11-10
Fix observation results handling for multiple measure observations

### LINKED ISSUE #159: Receiving unexpected measure-observation result on Ratio group
Expected behavior: Measure Observation output is tied to the associated population results.
Actual behavior: In some cases, we are receiving measure observation results when associated population results are false.

We have a Ratio scoring measure group with observations on both denom and numer.
In the execution results, IP is true, denom is false, and numer is false, but we are getting a true result and observation for the numerator observation.
Screen Shot 2022-10-28 at 12 57 07 PM

The bundle input includes the criteria reference extension on the observation populations to indicate which population (denom or numer) the observation is for.
Screen Shot 2022-10-28 at 12 58 35 PM

Measure bundle and test case JSON are attached for reference.
[bundles.zip](https://github.com/projecttacoma/fqm-execution/files/9890488/bundles.zip)

natjoe4:
# Summary
Fixes #159 

Populations that have measure-observations associated with them through the cqfm-criteriaReference extension will now have the associated measure-observation's result set to false and observations set to null when the population itself is false

## New behavior
When running detailed results calculation for measures with both single and multiple IPPs, the epsiodeResults array will now null out observations and false out results for measure-observation populations that reference populations with `result:false` through the criteria reference extension

## Code changes
- Added branches in `handlePopulationValues` function to track down and null out necessary measure-observations
- Added helper functions in `MeasureBundleHelpers.ts` to more easily track populations by criteria reference
- Added testing

# Testing guidance
 - Run detailed results on the bundles included in #159 and ensure that in all places in the episode results where`denom` is false, `numFunc` is also false with `observations: null`, and same for `numer` and `numerFunc`. Try bundles with multiple IPPs from #160 and normal bundles to ensure no other calculation is affected.


mgramigna:
I will be taking ownership of this. Changing the base branch to make the code diff cleaner

src/helpers/MeasureBundleHelpers.ts
```diff
@@ -122,6 +122,61 @@ export function hasMultipleIPPs(group: fhir4.MeasureGroup) {
...
+      measureObs.observations = null;
+    }
+  }
+}
+
+/*
+   Finds the measure observation that references the desired population in its criteria reference 
+   and returns its populationResult
+*/
+export function getCriteriaRefMeasureObs(
```
mgramigna, line 152: I wonder if we can unify this function with the `getRelevantIPPFromPopulation` since they're very very similar

mgramigna (changes requested):

test/fixtures/MeasureBundleHelpersFixtures/groupNumerAndDenomCriteria.json
```diff
@@ -0,0 +1,109 @@
...
+          {
+            "system": "http://terminology.hl7.org/CodeSystem/measure-population",
+            "code": "denominator",
+            "display": "Denominator"
+          }
+        ]
+      }
+    },
+    {
+      "id": "12c16634-ccf6-4d8f-acd3-5e52bdbfc5d3",
```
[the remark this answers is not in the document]
mgramigna: Technically the denex is relevant because there's a branch of logic there, but I'll remove the numex since it isn't used.

I'll also go ahead and update the IDs to be clearer instead of random UUIDs for better readability

src/helpers/MeasureBundleHelpers.ts
mgramigna: Yeah.. good point. Though the functions do serve slightly different purposes. One is identifying a `fhir4.MeasurePopulation` within a group that is referenced via the criteriaReference extension. The other one is finding a `PopulationResult` for a measure observation that `references` a certain population.

I'll go ahead and update the implementation and the naming of this function to clearly reflect that, and will be sure to re-use existing helper functions where appropriate

mgramigna: I also think that, to be more explicit, I'll go ahead and update it to return null instead of undefined.

src/calculation/DetailedResultsBuilder.ts
```diff
@@ -151,31 +153,39 @@ export function handlePopulationValues(
   ) {
     setResult(PopulationType.DENEX, false, populationResults);
     setResult(PopulationType.DENEXCEP, false, populationResults);
-
-    if (!MeasureBundleHelpers.hasMultipleIPPs(group)) {
+    DetailedResultsHelpers.nullCriteriaRefMeasureObs(group, populationResults, PopulationType.DENOM);
+
+    // If there is a MSRPOPL, all observations point to it, so null them out
+    if (hasResult(PopulationType.MSRPOPL, populationResults)) {
```
mgramigna, line 159: It looks like this branch and the MSRPOPLEX branch are not covered by tests. Can we add tests for them?

mgramigna (approved):

elsaperelli (changes requested):
Looks good to me! Just some questions that we discussed

src/helpers/DetailedResultsHelpers.ts
```diff
@@ -100,3 +100,30 @@ export function addIdsToPopulationResult(populationResult: PopulationResult, pop
...
+ *  the nulling of irrelevant observations happens already via `handlePopulationValues`
+ */
+export function nullCriteriaRefMeasureObs(
+  group: fhir4.MeasureGroup,
+  populationResults: PopulationResult[],
+  desiredPopulationType: PopulationType
+): void {
+  const popResults = populationResults.filter(result => result.populationType === PopulationType.OBSERV);
+  if (desiredPopulationType === PopulationType.NUMER && popResults.length === 1) {
+    // If only one measure-observation population, we know it relates to numerator
```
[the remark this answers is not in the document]
mgramigna: I think technically it could yeah. The safer move is just to null out the result regardless of what the desired populationType is. We should never really even be in this case in actual examples

elsaperelli (approved):
Lgtm! Nice job Nat&Matt

PR #170, opened 2022-11-30
Clause coverage highlighting for multiple groups

elsaperelli:
# Summary
Before, the coverage highlighting only happened for one measure group. Now, it calculates the coverage for all patients for each individual group.

## New behavior
`generateClauseCoverageHTML` now returns a lookup object where the key is the groupId and the value is the clause coverage html for that group. There is an additional property on the CalculationOutput interface for `groupClauseCoverageHTML`. 

## Code changes
- `src/calculation/Calculator.ts`- The lookup object created by `generateClauseCoverageHTML` is now returned as part of the CalculationOutput in the `calculate` function. If the debug output is enabled, an html file for the overall clause coverage html is created and contains the html for each group. In addition, an html file is created for each of the groups clause coverage html.
- `src/Calculation/HTMLBuilder.ts`- `generateClauseCoverageHtml` iterates through each execution result and adds the detailedResult to a lookup object where the key is its corresponding groupId. Then, it iterates through that lookup object and calculates the clauseCoverageHtml for each of the groups in that lookup object. The html for each group is added to another lookup object where the key is the groupId and the value is the clause coverage html for that group. That lookup object is then returned.
- `src/types/Calculator.ts`- The optional `groupClauseCoverageHTML` property was added to the `CalculationOutput` interface.
- `test/HTMLGenerator.test.ts`- Tests were changed to reflect the changes to the return type of `generateClauseCoverageHTML`.
- `test/fixtures/html/SimpleCoverageAnnotation.html`- This file was modified to include the group name.
- `test/fixtures/html/SimpleCoverageAnnotation2.html`- This file was added for the unit test that tests `generateClauseCoverageHTML` on multiple groups.

# Testing guidance
1. `npm run check` to ensure that unit tests pass and there are no prettier or lint errors/warnings.
2. Using the files attached below, run the following command: `npm run cli -- detailed -m <path-to-measure-bundle.json> -p <path-to-pt1.json> <path-to-pt2.json> -o --debug`
3. Look at the html files in the debug directory. There should be 4 files: two for each group that includes that group's clause coverage and two for each group that includes that group's measure highlighting.
4. Let me know if you have any questions and more importantly if you have any suggestions for additional unit tests or testing methods. 

[coverage-test-case.zip](https://github.com/projecttacoma/fqm-execution/files/10125052/coverage-test-case.zip)

### DIFF
--- a/src/calculation/HTMLBuilder.ts
+++ b/src/calculation/HTMLBuilder.ts
@@ -178,19 +178,28 @@ export function generateClauseCoverageHTML(
 
     const clauses = detailedResults.flatMap(c => (c.clauseResults ? c.clauseResults : []));
     clauseResults.push(clauses);
-
     const flattenedStatementResults = statementResults.flatMap(s => s);
     const flattenedClauseResults = clauseResults.flatMap(c => c);
 
-    // get all "unique" statements (by library name and localid) and filter by relevance
+    // Grab every statement with any relevance other than N/A
+    // There may be multiple entries for a given statement across the results,
+    // but we know that non of them can be irrelevant
     const relevantStatements = uniqWith(
       flattenedStatementResults,
+      (s1, s2) => s1.libraryName === s2.libraryName && s1.localId === s2.localId && s1.relevance === s2.relevance
+    ).filter(s => s.relevance !== Relevance.NA);
+
+    // From all the relevant ones, filter out any duplicate statements
+    // uniqWith appears to pick the first element in encounters that matches the uniqueness condition
+    // when iterating, which is fine because the relevance not being N/A is the only thing that matters now
+    const uniqueRelevantStatements = uniqWith(
+      relevantStatements,
       (s1, s2) => s1.libraryName === s2.libraryName && s1.localId === s2.localId
-    ).filter(s => s.relevance === Relevance.TRUE);
+    );
 
     // assemble array of statement annotations to be templated to HTML
     const statementAnnotations: { libraryName: string; annotation: Annotation[] }[] = [];
-    relevantStatements.forEach(s => {
+    uniqueRelevantStatements.forEach(s => {
       const matchingLibrary = elmLibraries.find(e => e.library.identifier.id === s.libraryName);
       if (!matchingLibrary) {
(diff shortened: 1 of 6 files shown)


natjoe4 (changes requested):
Just a couple small changes here. Functionality looks great in the provided example! Nice job

mgramigna (approved):

src/calculation/HTMLBuilder.ts
```diff
@@ -6,6 +6,7 @@ import mainTemplate from '../templates/main';
 import clauseTemplate from '../templates/clause';
 import { UnexpectedProperty, UnexpectedResource } from '../types/errors/CustomErrors';
 import { uniqWith } from 'lodash';
+import { connect } from 'http2';
```
natjoe4, line 9: Looks like this snuck in in the last commit

src/calculation/HTMLBuilder.ts
```diff
@@ -146,72 +147,92 @@ export function generateHTML(
...
+      flattenedStatementResults,
+      (s1, s2) => s1.libraryName === s2.libraryName && s1.localId === s2.localId && s1.relevance === s2.relevance
+    ).filter(s => s.relevance !== Relevance.NA);
 
-    const matchingExpression = matchingLibrary.library.statements.def.find(e => e.name === s.statementName);
-    if (!matchingExpression) {
-      throw new UnexpectedProperty(`No statement ${s.statementName} found in library ${s.libraryName}`);
-    }
+    // From all the relevant ones, filter out any duplicate statements
+    // uniqWith appears to pick the first element in encounters that matches the uniqueness condition
```
natjoe4, line 187:
```suggestion
    // uniqWith appears to pick the first element it encounters that matches the uniqueness condition
```

natjoe4 (changes requested):
Good catch on the flatmap being unnecessary! Just a couple little linting issues

src/calculation/HTMLBuilder.ts
mgramigna: Haha sorry

mgramigna (approved):

natjoe4 (approved):
Looks great!

PR #171, opened 2022-11-30
 Upgraded cql-exec-fhir and got rid of .d.ts file

elsaperelli:
# Summary
Upgrades cql-exec-fhir to newest patch version since https://github.com/cqframework/cql-exec-fhir/releases/tag/v2.1.3 was released.

## New behavior
No new behavior

## Code changes
- Bumps cql-exec-fhir version
- Deletes `CQLExecFHIR.d.ts`

# Testing guidance
- `npm install` and make sure nothing breaks.
- `npm run test` to ensure that unit tests still pass.


natjoe4 (approved):
All unit tests pass and no linting errors, so looks good to me! I think the `as any as CQLPatient` syntax is a little goofy, but seems necessary for the changes :)

test/queryFilters/interpretExpression.test.ts
```diff
@@ -10,7 +10,7 @@ const complexQueryELM = getELMFixture('elm/queries/ComplexQueries.json');
 const PATIENT = FHIRWrapper.FHIRv401().wrap({
   resourceType: 'Patient',
   birthDate: '1988-09-08'
-}) as CQLPatient;
+}) as any as CQLPatient;
```
sarahmcdougall, line 13: It looks like the `wrap()` function used to return type `any` but now returns type `FHIRObject`, which is why we need to cast to `any`. @mgramigna do you think this is an okay approach so that we can still cast to type `CQLPatient`? (asking you since you made the return type change to `wrap()`)

sarahmcdougall (changes requested):
Looks like the `package-lock` got changed to version 1, so we should change it back to version 2.

test/queryFilters/interpretExpression.test.ts
mgramigna: Haha yeah I ok-ed this because of how CQLPatient is defined. I think in the future there's a chance we can unify it better with the cql-exec-fhir types, but I don't think we should worry about it now.

The `as any` was to avoid tsserver yelling about insufficient overlap during a type cast. Happens to the best of us

sarahmcdougall (approved):

natjoe4 (approved):
Looks good!

PR #175, opened 2022-12-13
Integration testing setup

mgramigna:
# Summary

Adds necessary scripts and setup for the new integration testing approach. Gets rid of old approach and sets up structure for new one.

## New behavior

None in the actual codebase.

Main new behavior to call out is new scripts:

* `npm run build:test-data`: Builds all measure bundles in `test/integration`
* `npm run test:integration`: Runs all the `.test.ts` files inside the `test/integration` directory/subdirectories

## Code changes

* Move all unit tests from `test` to `test/unit` to keep things clean (updated import relative paths as well)
* Create `test/integration` directory with a placeholder for `proportion-boolean` measure
  * Placeholder cql file for measure logic
  * Script to generate the bundle using `ecqm-bundler`
  * Example test file that does nothing
  * Makefile for starting the translator container and generating everything
  * Resulting measure bundle from generation
* Update `integration` step of github actions yml to reflect new script
* Add README to integration test directory detailing steps
* Update  README to add link to integration testing README

# Testing guidance

Check out the instructions in the integration test README and make sure it's logical. You can even try following the steps for creating a new one and seeing if it works.

Try out `npm run build:test-data` and `npm run test:integration`

Visually inspect everything in `test/integration`


mgramigna (commented):
One thing I just realized is that re-building all the test data will generate new UUIDs on the bundles even if the underlying CQL has not changed.

I am going to look into ensuring that bundles are re-generated only if their cql has changed

mgramigna:
> One thing I just realized is that re-building all the test data will generate new UUIDs on the bundles even if the underlying CQL has not changed.
> 
> I am going to look into ensuring that bundles are re-generated only if their cql has changed

Resolved in https://github.com/projecttacoma/fqm-execution/pull/175/commits/8504e4e7a9991b818803a8a89e86bb50eccda17d

test/integration/Makefile
```diff
@@ -0,0 +1,30 @@
...
+	done
+
+.start-translator:
+	docker pull cqframework/cql-translation-service:latest
+	docker run --name cql-translation-service --rm -dit -p 8080:8080 cqframework/cql-translation-service:latest
+	sleep 5
+	touch .start-translator
+
+proportion-boolean:
+	$(BASE_PATH)/proportion-boolean/generate.sh
```
mgramigna, line 26: Open question from me:

Does it even make sense to have a `generate.sh` script in each subdirectory? The alternative is to put the ecqm-bundler command directly in this make target instead of having it call out to a script

Pros of current approach:

* Keeps Makefile clean (avoids long ecqm-bundler commands embedded in the mMkefile)

Cons of current approach:

* Requires a new file to be created in every directory (more steps)
* Adds some bloat

Pros of new approach:

* No need for additional generate.sh file

Cons of new approach:

* ecqm-bundler commands may get quite long and could be hard to read within the makefile

test/integration/README.md
```diff
@@ -0,0 +1,76 @@
...
+BASE_PATH=$(shell dirname $BASH_SOURCE)
+
+# ...
+
+all: [scoring-code]-[population-basis] # ...
+
+# ...
+
+[scoring-code]-[population-basis]:
+    $(BASE_PATH)/[scoring-code]-[population-basis]/generate.sh
```
elsaperelli, line 71: I think we can remove `BASE_PATH` now?

elsaperelli (changes requested):
This is really really awesome. Testing went super smoothly for me and the integration test README was very easy to understand. Just a few things in the README but other than that this is great.

test/integration/Makefile
mgramigna: The Makefile target would become this:


```makefile
proportion-boolean:
	ecqm-bundler -c ./proportion-boolean/cql/proportion-boolean.cql --deps ./helpers/cql/FHIRHelpers-4.0.1.cql --ipop ipp --numer numer --denom denom --scoring-code proportion --basis boolean -o ./proportion-boolean/proportion-boolean-bundle.json
```

test/integration/proportion-boolean/cql/proportion-boolean.cql
```diff
@@ -0,0 +1,14 @@
+library ProportionBooleanTest version '0.0.1'
+
+using FHIR version '4.0.1'
+
+include FHIRHelpers version '4.0.1'
+
```
hossenlopp, line 6:
```suggestion
parameter "Measurement Period" Interval<DateTime>
  default Interval[@2019-01-01T00:00:00.0, @2020-01-01T00:00:00.0)

context Patient

```

This example should have the MP and needs `context Patient`

hossenlopp (approved):
Looks like it will do the trick to get us started. One suggested change on the example you have there now.

test/integration/proportion-boolean/cql/proportion-boolean.cql
mgramigna: Good call! Gonna make the default 2022

elsaperelli (approved):
Lgtm! This structure is really great and everything is explained very thoroughly and logically. Nice :)

PR #178, opened 2022-12-15
Proportion Boolean Integration Test

natjoe4:
# Summary
Created the first integration testing suite in fqm-execution! This one's for a simple proportion-boolean measure.
 - CQL was bundled with @mgramigna 's[ecqm-bundler](https://github.com/mgramigna/ecqm-bundler)
 - Patient bundles generated with [fqm-testify](https://github.com/projecttacoma/fqm-testify)
 
## New behavior
 - On `npm run build:test-data` && `npm run test:integration` we now have a test suite that runs!
 
## Code changes
 - updated makefile with appropriate code paths
 - added MATGlobalCommonFunctionsFHIR4 to shared CQL directory (is this ok?)
 - created a `getJSONFixture` function to more easily pull and type JSON fixtures from patient directories
 - created `proportion-boolean.cql` by hand, which eventually gets bundled into `proportion-boolean-bundle.json`
 - created `patient-denom-bundle`, 'patient-ipp-bundle`, and `patient-numer-bundle` from fqm-testify and added to `patients` directory.
 - created `exampleValueset1`, `exampleValueset2`, and `exampleValueset3`, for use across any integration test
 - updated `proportion-boolean-bundle` to include custom CQL
 - filled in `proportion-boolean.test.ts`
 
# Testing guidance
- run `npm run build:test-data`
- run `npm run test:integration`
- take a closer look at our custom CQL and make sure everything makes sense and is sufficiently simple
- can you think of any other tests that make sense for the given test case? positive or negative?


test/integration/proportion-boolean/cql/proportion-boolean.cql
```diff
@@ -3,17 +3,33 @@ library ProportionBooleanTest version '0.0.1'
...
 include FHIRHelpers version '4.0.1'
+include MATGlobalCommonFunctionsFHIR4 version '6.1.000' called Global
+
+valueset "Example": 'http://example.com/exampleId'
+valueset "Example2": 'http://example.com/exampleId2'
+valueset "Example3": 'http://example.com/exampleId3'
+
+parameter "Measurement Period" Interval<DateTime>
+
+context Patient
```
mgramigna, line 14: These are defined right below this, no need to duplicate

```suggestion
```

test/integration/proportion-boolean/patients/patient-denom-bundle.json
```diff
@@ -0,0 +1,178 @@
...
+              {
+                "code": {
+                  "coding": [
+                    {
+                      "system": "http://terminology.hl7.org/CodeSystem/measure-population",
+                      "code": "initial-population"
+                    }
+                  ]
+                },
+                "count": 0
```
mgramigna, line 141: It doesn't _really_ matter for the sake of this integration test, but maybe in the future if we ever want to do anything with the test case measure report, we should properly reflect it in fqm-testify.

AKA can you actually select the desired populations in testify before exporting the patient?

test/integration/proportion-boolean/patients/patient-ipp-bundle.json
```diff
@@ -0,0 +1,151 @@
...
+              {
+                "code": {
+                  "coding": [
+                    {
+                      "system": "http://terminology.hl7.org/CodeSystem/measure-population",
+                      "code": "initial-population"
+                    }
+                  ]
+                },
+                "count": 0
```
mgramigna, line 114: It doesn't _really_ matter for the sake of this integration test, but maybe in the future if we ever want to do anything with the test case measure report, we should properly reflect it in fqm-testify.

AKA can you actually select the desired populations in testify before exporting the patient?

test/integration/proportion-boolean/patients/patient-numer-bundle.json
```diff
@@ -0,0 +1,205 @@
...
+              {
+                "code": {
+                  "coding": [
+                    {
+                      "system": "http://terminology.hl7.org/CodeSystem/measure-population",
+                      "code": "initial-population"
+                    }
+                  ]
+                },
+                "count": 0
```
mgramigna, line 168: It doesn't _really_ matter for the sake of this integration test, but maybe in the future if we ever want to do anything with the test case measure report, we should properly reflect it in fqm-testify.

AKA can you actually select the desired populations in testify before exporting the patient

test/integration/proportion-boolean/proportion-boolean.test.ts
```diff
@@ -1,3 +1,61 @@
...
+    const results = await calculate(MEASURE_BUNDLE, [IPP_PATIENT_BUNDLE], DEFAULT_CALCULATION_OPTIONS);
+    expect(results.results[0].detailedResults?.[0]?.populationResults).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({ populationType: 'initial-population', result: true }),
+        expect.objectContaining({ populationType: 'denominator', result: false }),
+        expect.objectContaining({ populationType: 'numerator', result: false })
+      ])
+    );
+  });
+  it('calculates denom patient into denom', async () => {
```
mgramigna, line 31: plz

```suggestion

  it('calculates denom patient into denom', async () => {
```

test/integration/proportion-boolean/proportion-boolean.test.ts
```diff
@@ -1,3 +1,61 @@
...
+    const results = await calculate(MEASURE_BUNDLE, [DENOM_PATIENT_BUNDLE], DEFAULT_CALCULATION_OPTIONS);
+    expect(results.results[0].detailedResults?.[0]?.populationResults).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({ populationType: 'initial-population', result: true }),
+        expect.objectContaining({ populationType: 'denominator', result: true }),
+        expect.objectContaining({ populationType: 'numerator', result: false })
+      ])
+    );
+  });
+  it('calculates numer patient into numer', async () => {
```
mgramigna, line 41:
```suggestion

  it('calculates numer patient into numer', async () => {
```

test/integration/proportion-boolean/proportion-boolean.test.ts
```diff
@@ -1,3 +1,61 @@
...
+    const results = await calculate(MEASURE_BUNDLE, [NUMER_PATIENT_BUNDLE], DEFAULT_CALCULATION_OPTIONS);
+    expect(results.results[0].detailedResults?.[0]?.populationResults).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({ populationType: 'initial-population', result: true }),
+        expect.objectContaining({ populationType: 'denominator', result: true }),
+        expect.objectContaining({ populationType: 'numerator', result: true })
+      ])
+    );
+  });
+  it('calculates numer into denom for different measurement period', async () => {
```
mgramigna, line 51:
```suggestion

  it('calculates numer into denom for different measurement period', async () => {
```

test/integration/proportion-boolean/proportion-boolean.test.ts
```diff
@@ -1,3 +1,61 @@
...
+
+const IPP_PATIENT_BUNDLE: fhir4.Bundle = getJSONFixture('proportion-boolean/patients/patient-ipp-bundle.json');
+const DENOM_PATIENT_BUNDLE: fhir4.Bundle = getJSONFixture('proportion-boolean/patients/patient-denom-bundle.json');
+const NUMER_PATIENT_BUNDLE: fhir4.Bundle = getJSONFixture('proportion-boolean/patients/patient-numer-bundle.json');
+const MEASURE_BUNDLE: fhir4.Bundle = getJSONFixture('proportion-boolean/proportion-boolean-bundle.json');
+
+describe('proportion boolean measure', () => {
+  it('calculates ipp patient into ipp', async () => {
+    const results = await calculate(MEASURE_BUNDLE, [IPP_PATIENT_BUNDLE], DEFAULT_CALCULATION_OPTIONS);
+    expect(results.results[0].detailedResults?.[0]?.populationResults).toEqual(
```
mgramigna, line 23: This assertion is going to be rather annoying to repeat. We can probably create something in `testHelpers` that is a re-usable assertion helper for this kind of thing which will be common in the integration tests. You don't have to take this code exactly, but I was thinking something like:

```ts
function assertPopulationResults(
  result: ExecutionResult<DetailedPopulationGroupResult>,
  expectedPopulations: Partial<Record<PopulationType, boolean>>
) {
  const expectedPopulationResults: PopulationResult[] = Object.entries(expectedPopulations).map(([popType, value]) =>
    expect.objectContaining({
      populationType: popType as PopulationType,
      result: value
    })
  );

  expect(result).toEqual(
    expect.objectContaining({
      detailedResults: expect.arrayContaining([
        expect.objectContaining({
          populationResults: expectedPopulationResults
        })
      ])
    })
  );
}
```

Then in your tests, all you have to do is this:

```ts
assertPopulationResults(results[0], {
  [PopulationType.IPP]: true,
  [PopulationType.NUMER]: false,
  [PopulationType.DENOM]: false
});
```

thoughts?

(side note: typescript enums are terrible, we really should get rid of them lol)

mgramigna (changes requested):

test/integration/proportion-boolean/patients/patient-denom-bundle.json
natjoe4: Gotcha. We were on the fence about this and opted not to. Happy to add them back :)

hossenlopp (changes requested):

test/integration/helpers/testHelpers.ts
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Formats expectedPopulations object into an array of objects which can be passed into an
+ * expect.arrayContaining and compared with result
+ */
+export function assertPopulationResults(
+  result: ExecutionResult<DetailedPopulationGroupResult>,
+  expectedPopulations: Partial<Record<PopulationType, boolean>>
+) {
+  const expectedPopulationResults: PopulationResult[] = Object.entries(expectedPopulations).map(([popType, value]) =>
```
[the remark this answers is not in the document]
natjoe4: good catch! Just so I understand, you're asking for the `results` arg to be a bit more specific here? Do you think it's a good plan to parse through the measure on the test side to grab the groupId and then just drill down to the specific group and pass that in for results instead?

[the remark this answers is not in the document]
mgramigna: Good point. Maybe this helper should just take the results for a specific group (i.e. the `expect` starts on an entry of the `detailedResults` list rather than doing an expect.arraycontaining on it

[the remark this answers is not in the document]
hossenlopp: ^ I think this is the better way, even though it will push more logic into the actual test. Although I wonder having a helper to grab the population group results by index or id and throw an `expect` if it cannot be found. Then you can take it and pass into this function to compare results.

test/integration/helpers/testHelpers.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+export function getJSONFixture(path: string): any {
+  return JSON.parse(readFileSync(`test/integration/${path}`).toString());
+}
+
+/**
+ * Parses through the execution result and asserts that the population group at the specified index exists.
+ * Returns that group if so
+ */
+export function getGroupByIndex(
```
mgramigna, line 16: How about changing this to be by groupId, not by index? Every result should have a group ID

hossenlopp: I thought group id was not required, at least in base FHIR.

mgramigna: It's not, but I believe fqm-execution puts a default there when not provided to be `group-${index}`

See https://github.com/projecttacoma/fqm-execution/blob/90b1951e7ba5b8a41c9b93e3c873cae243496134/src/types/Calculator.ts#L111 -- groupId is always there

natjoe4: The reason I didn't do this is that regenerating the bundle will cause all the groupIds to be regenerated (right?) meaning every test will need to be updated any time we change/add to the CQL. Is this a concern? @mgramigna @hossenlopp

mgramigna: OH you're 100% right lol

mgramigna (approved):
sweeeet

mgramigna:
I think we're good here

PR #179, opened 2022-12-28
Coverage functions

natjoe4:
# Summary
Remove CQL functions from clause coverage calculation

## New behavior
Previously, functions counted as clauses when calculating clause coverage. This led to coverage numbers being lower than appropriate. Functions are now identified and marked during statement and clause result generation, then filtered out during clause coverage calculation

## Code changes
 - Add line to `buildStatementAndClauseResults` that sets `isFunction` property to true if appropriate
 - Filter out clauses with `isFunction` set to `TRUE` in `calculateClauseCoverage`
 - add `isFunction` property to `StatementResult` interface
 - Updated HTMLGenerator test file to HTMLBuilder to reflect file name change

# Testing guidance
 - `npm run check`
 - download the attached measure and patient bundle and store at root level of fqm-execution
 - run `npm run build`
 - `node build/cli.js detailed  -m "./coverage-calc/measure-bundle.json" -p "./coverage-calc/patient-numer-bundle.json" --debug -o -s "2022-01-01" -e "2022-12-31"`
 - Head into the generated `debug/html` folder and view the `clause-coverage` file. Coverage should now be `100%`
 - `node build/cli.js detailed  -m "./coverage-calc/measure-bundle.json" -p "./coverage-calc/patient-denom-bundle.json" --debug -o -s "2022-01-01" -e "2022-12-31"`
 - Head into the generated `debug/html` folder and view the `clause-coverage` file. Coverage should now be `75%`
[coverage-calc.zip](https://github.com/projecttacoma/fqm-execution/files/10314563/coverage-calc.zip)

### DIFF
--- a/src/types/Calculator.ts
+++ b/src/types/Calculator.ts
@@ -177,6 +177,8 @@ export interface StatementResult {
   raw?: any;
   /** Pretty result for this statement. */
   pretty?: string;
+  /** TRUE if the statement is a function */
+  isFunction?: 'TRUE' | 'FALSE';
 }
 
 /**
--- a/src/types/Calculator.ts
+++ b/src/types/Calculator.ts
@@ -178,7 +178,7 @@ export interface StatementResult {
   /** Pretty result for this statement. */
   pretty?: string;
   /** TRUE if the statement is a function */
-  isFunction?: 'TRUE' | 'FALSE';
+  isFunction?: boolean;
 }
 
 /**
--- a/test/unit/HTMLBuilder.test.ts
+++ b/test/unit/HTMLBuilder.test.ts
@@ -4,7 +4,8 @@ import {
   objToCSS,
   cqlLogicClauseTrueStyle,
   cqlLogicClauseFalseStyle,
-  cqlLogicClauseCoveredStyle
+  cqlLogicClauseCoveredStyle,
+  calculateClauseCoverage
 } from '../../src/calculation/HTMLBuilder';
 import {
   StatementResult,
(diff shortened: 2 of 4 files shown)


src/types/Calculator.ts
```diff
@@ -177,6 +177,8 @@ export interface StatementResult {
   raw?: any;
   /** Pretty result for this statement. */
   pretty?: string;
+  /** TRUE if the statement is a function */
+  isFunction?: 'TRUE' | 'FALSE';
```
lmd59, line 181: Okay- I gotta know. Why is this a set of possible strings rather than just a boolean?

lmd59, line 181: Okay- I gotta know. Why is this a set of possible strings rather than just a boolean?

natjoe4: Lol no reason aside from a lot of the values in StatementResult wind up being all caps strings. Just changed :)

test/unit/HTMLBuilder.test.ts
```diff
@@ -152,4 +153,46 @@ describe('HTMLGenerator', () => {
       generateHTML([elm], badStatementResults, [], 'test');
     }).toThrowError();
   });
+  test('clause coverage percent with a function ignores function in calculation', () => {
```
elsaperelli, line 156: I would just add a space above this test to be consistent with the rest of the file :)

elsaperelli (changes requested):
Nice! Just one inline comment and one overall suggestion. In `ClauseResultsBuilder.ts`, would it be possible to rearrange the `if/else` statements so that `isStatementFunction()` is only called once rather than twice in two different clauses? I didn't look too hard into whether this was possible but I think it might be if you think that would make the code a bit cleaner or if it's not worth it!

elsaperelli (approved):
Lgtm!

lmd59 (approved):

PR #180, opened 2022-12-28
Fix observation function reuse issue with results recording

### LINKED ISSUE #176: Ratio, episode-based measure observation results not appearing as expected
There are a few scenarios where we are not seeing the observation results that we would expect.
The first is for a ratio measure group where there is a single IP, and an observation on both the denominator and numerator.
We select the same function for both observations.
In the episode results for one episode, IP is true, Denominator is true and Numerator is true. The Denominator Observation is true but has two results in its observations array.
The Numerator Observation is false and has no results in its observations array.
We'd expect that the Numerator Observation result be true and to have a result when the Numerator passes. Instead it seems like the numerator observation result is being added to the denominator observation result array.
To verify which observation is for numerator and denominator we are going by the criteriaReferenceId, as that ties back to either the numerator or denominator populationId.
image

The second scenario is ratio measure group with two IPs, and an observation on both denominator and numerator.
Again, we select the same function for both observations.
After execution again we'd expect one observation result for the denominator and one for numerator. (This is even more important in this case as denominator and numerator have separate IPs.)
Instead, we see both results on the denominator observation.
image

Another episode for the same test doesn't appear to have any observation result even though the second IP and associated Numerator both pass. The Numerator Observation has a false and no observations array:
image

We'd expect the Numerator Observation here to have a true result and contain the result of the observation function.

Bundles for these scenarios are attached to this issue. Please let me know if any other information is needed or if we can help explain the scenarios any better.
Thanks!
[ratio-encounter-1ip.zip](https://github.com/projecttacoma/fqm-execution/files/10221392/ratio-encounter-1ip.zip)
[ratio-encounter-2ips.zip](https://github.com/projecttacoma/fqm-execution/files/10221393/ratio-encounter-2ips.zip)

hossenlopp:
# Summary

Fixes #176 

Fixes results recording issues with a re-used observation function a ratio measure as exhibited in #176. Now ensures the criteriaReferenceId matches for populations when adding observation results to populations seen before in tallying.

## New behavior

Bug is fixed and re-uses of a function for observation will have their results routed to the correct episode and population result objects.

## Code changes

Added `ratio-Encounter-reuseObservationFunction` integration test based on data from #176.
DetailedResultsBuilder.ts - Fixes in collection of population results to make sure criteriaReferenceId is used to match up results.
ClauseResultsBuilder.ts - Fixes to helper functions for setting results so they match on criteriaReferenceId.

# Testing guidance

Run integration tests.


test/integration/Makefile
```diff
@@ -23,6 +23,9 @@ all: .start-translator
 proportion-boolean:
 	ecqm-bundler -c ./proportion-boolean/cql/proportion-boolean.cql -v ./valuesets --deps-directory ./helpers/cql --ipop ipp --numer numer --denom denom --scoring-code proportion --basis boolean -o ./proportion-boolean/proportion-boolean-bundle.json
 
+ratio-Encounter-reuseObservationFunction:
+	ecqm-bundler -c ./ratio-Encounter-reuseObservationFunction/cql/ratio-Encounter-reuseObservationFunction.cql -v ./valuesets --deps-directory ./helpers/cql --ipop ipp --numer numer --denom denom --detailed-msrobs "daysObs|numer" "daysObs|denom" --scoring-code ratio --basis Encounter -o ./ratio-Encounter-reuseObservationFunction/ratio-Encounter-reuseObservationFunction-bundle.json
```
mgramigna, line 27: beautiful, glad this worked lol

test/integration/ratio-Encounter-reuseObservationFunction/cql/ratio-Encounter-reuseObservationFunction.cql
```diff
@@ -0,0 +1,24 @@
...
+  ["Encounter"] E where E.period.end during "Measurement Period" and E.status = 'finished'
+
+define "numer":
+  "ipp"
+
+define function daysObs(e Encounter):
+  duration in days of e.period
+
+define function hoursObs(e Encounter):
+  duration in hours of e.period
```
mgramigna, line 24: should we get rid of hoursObs in this specific fixture since it is not applicable to the test case?

test/integration/ratio-Encounter-reuseObservationFunction/patients/patient-bundle.json
```diff
@@ -0,0 +1,184 @@
+{
```
mgramigna, line 1: I wonder if for consistency we should create a new patient in fqm-testify just so the format of patient data is consistent across all integration tests?

test/integration/ratio-Encounter-reuseObservationFunction/ratio-Encounter-reuseObservationFunction.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    const numerObservResults = popResults.find(
+      r => r.populationType === PopulationType.OBSERV && r.criteriaReferenceId === numerResults.populationId
+    );
+    const denomObservResults = popResults.find(
+      r => r.populationType === PopulationType.OBSERV && r.criteriaReferenceId === denomResults.populationId
+    );
+    expect(numerObservResults).toBeDefined();
+    expect(denomObservResults).toBeDefined();
+    expect(numerObservResults?.observations).toEqual([1]);
+    expect(denomObservResults?.observations).toEqual([1]);
```
mgramigna, line 43: Not strictly necessary, but just thinking ahead here. Do you think it would be beneficial to have a test helper that allows for repeatable assertions of observation results? Similar to what we did for `assertPopulationResults`? Then maybe there's a world where we can do something like (pseudocode):

```ts
assertMeasureObservation(groupInfo, criteriaReferenceId, [1])
```

src/calculation/DetailedResultsBuilder.ts
```diff
@@ -52,7 +52,9 @@ export function createPopulationValues(
         episodeResult.populationResults.forEach(popResult => {
           const measureObsPop = populationResults.find(
             pop =>
-              pop.populationType === PopulationType.OBSERV && popResult.criteriaExpression === pop.criteriaExpression
+              pop.populationType === PopulationType.OBSERV &&
+              popResult.criteriaExpression === pop.criteriaExpression &&
+              pop.criteriaReferenceId === popResult.criteriaReferenceId
```
mgramigna, line 57: I remember when the only measure type where we were dealing with measure observations was a CV episode measure (cms111). This measure had one measure observation without a criteriaReference extension, and so we assumed it was observing the measure observation.

This was a while ago though, and looking at the IG, it looks like even if there's only 1 measure observation in a CV measure, it SHALL use the criteriaReferenceId: https://build.fhir.org/ig/HL7/cqf-measures/measure-conformance.html#continuous-variable-measure

I was initially concerned that we could be dealing with measure observations without a criteriaReference extension, which I think this code would break on and not find the result. But just confirming that maybe we don't have to worry about that anymore?

If we wanted to maintain backwards compatibility with that assumption, we'd want to only do the criteriaReferenceId check if it exists on the measure observation population def to begin with.

Although now that I typed this out, it might technically still work for `undefined === undefined` lol, but i still think we should talk about it

mgramigna (changes requested):
Nice! Seems to solve the bug. I have 1 IG-related question that isn't a change request and then some additional suggestions on the integration tests.

p9g:
Note my QMIG ballot comment to clarify ratio measure requirements
  https://jira.hl7.org/browse/FHIR-39376

test/integration/ratio-Encounter-reuseObservationFunction/ratio-Encounter-reuseObservationFunction.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const PATIENT_BUNDLE: fhir4.Bundle = getJSONFixture(
+  'ratio-Encounter-reuseObservationFunction/patients/patient-bundle.json'
+);
+
+describe('ratio Encounter reuse observation function measure', () => {
+  it('does not put all results on observation usage of a function on one observation in population results', async () => {
+    const results = await calculate(MEASURE_BUNDLE, [PATIENT_BUNDLE], CALCULATION_OPTIONS);
+    if (!results.results[0].detailedResults) {
+      fail('Population DetailedResults not found');
```
lmd59, line 24: So there are a couple of `fail` calls when I might expect `expect` calls in these tests (I suspect to avoid typescript not recognizing that a value is non-null), and in wondering about whether there were any alternative approaches, I ended up in a conversation with @mgramigna about `fail` being removed from jest in later versions. We've used `fail` in tests before, but it might be time to start moving to other approaches to avoid incurring additional technical debt when we update jest some time in the future.

lmd59: There are a couple of potential workarounds for using `expect` instead. After the `expect`, we can throw an error for a null object, we can define our own fail that throws an error, we can explicitly cast the object to a non-null type, we can use the non-null operator `!` (and tell eslint to shutup about it) , or we could use the classic `?` even though we're not actually unsure about it.

lmd59: Thoughts on approach? Want to balance not cluttering code too much with sustainable practices, etc. (Happy to discuss in a more real-time forum!)

test/integration/ratio-Encounter-reuseObservationFunction/cql/ratio-Encounter-reuseObservationFunction.cql
hossenlopp: Oh true. Will remove.

hossenlopp: Removed in a771ed4.

test/integration/ratio-Encounter-reuseObservationFunction/patients/patient-bundle.json
hossenlopp: Did this in f207793. Want to add another test to better test it after I get the shared function for checking observ results done.

test/integration/ratio-Encounter-reuseObservationFunction/ratio-Encounter-reuseObservationFunction.test.ts
hossenlopp: Ooh! I didn't know about `fail` being removed. I really was trying to work against TS because the ? and arrays are a pain. I will re-work this after the new year with adding a helper function for looking at the observation results.

lmd59: Love it - sounds good!

test/integration/ratio-Encounter-reuseObservationFunction/ratio-Encounter-reuseObservationFunction.test.ts
hossenlopp: Implemented this in 435be92. Made it so the criteria reference is passed in by the population type of the referenced criteria. The helper function will look up that population by type, then use the id to find the right observation to assert on.

test/integration/ratio-Encounter-reuseObservationFunction/ratio-Encounter-reuseObservationFunction.test.ts
hossenlopp: Reworked the tests massively with 435be92. `expect(x).toBeDefined()` is used instead and then `as` is used afterwards to make TS happy.

mgramigna (approved):
Love the new test helpers. This sets us up well for the future. I don't think my last comment above should halt the merging, but let me know if you have thoughts on it

test/integration/ratio-Encounter-reuseObservationFunction/ratio-Encounter-reuseObservationFunction.test.ts
lmd59: Looks great!

lmd59 (approved):

PR #181, opened 2023-01-06
Implement library data requirements

elsaperelli:
# Summary
Before, we had only the option to calculate dataRequirements on a measure bundle. Now, we can calculate dataRequirements on a collection of Library resources (library bundle). This will be helpful when we implement Library/$data-requirements in the measure-repository-service.

## New behavior
New function in `Calculator.ts` calculates library dataRequirements from a Library bundle. This is in addition to the calculateDataRequirements function that calculates the dataRequirements on a measure bundle.

## Code changes
- `src/calculation/Calculator.ts` - new function `calculateLibrarydataRequirements` gets data requirements for a collection of Library resources. A lot of this code was identical to that in `calculateDataRequirements` but we wanted them to be separate functions so I refactored the duplicate code into a function in `DataRequirementHelpers.ts`.
- `src/cli.ts` - new cli option to run `libraryDataRequirements` with a flag with the `rootLibRef`.
- `src/helpers/DataRequirementHelpers.ts` - refactored a lot of the identical code in `calculateDataRequirements` and `calculateLibraryDataRequirements` to the `getDataRequirements` function.
- `src/helpers/MeasureBundleHelpers.ts` - `extractLibrariesFromLibraryBundle` gets libraries from a library bundle and `extractLibrariesFromMeasureBundle` gets libraries from a measure bundle. They both use `extractLibrariesFromBundle` because a lot of the code was the same. The main difference is that `extractLibrariesFromLibraryBundle` has to take a reference to the root library as a parameter, since there is no measure on the bundle to indicate the root library like in `extractLibrariesFromMeasureBundle`.
- `src/helpers/elm/ELMInfoCache.ts` - changed `extractLibrariesFromBundle` to `extractLibrariesFromMeasureBundle`.
- `src/types/Calculator.ts` - added `rootLibRef` as an optional CalculationOption which we need in `libraryDataRequirements` in the cli.
- `test/unit/helpers/ELMInfoCache.test.ts` - changed `extractLibrariesFromBundle` to `extractLibrariesFromMeasureBundle`.
- `test/unit/helpers/MeasureBundleHelpers.test.ts` - added a test for `extractLibrariesFromLibraryBundle`. This uses the `ELM130-7.3.000-bundles-nocodes.json` fixture that we already have but filters is so that it is a bundle of only its library resources. 

# Testing guidance
- `npm run check`
- Try the new cli command! 
- My motivation behind this test is I want to take a measure bundle and run dataRequirements on it. Then, I want to take that same measure bundle, but remove all resources that are not of type Library. Finally, I run libraryDataRequirements on that library bundle and then compare the two outputs. They should be the same.
- With this approach, I used the `EXM130-7.3.000-bundle-nocodes.json` fixture because it is there. I suggest you do the same since I attached that file and the one with only the library resources below, but feel free to try other bundles!
- `npm run cli -- dataRequirements -m <path-to-EXM130-7.3.000-bundle-nocodes.json> -o` 
- Save that `output.json` file
- `npm run cli -- libraryDataRequirements -m <path-to-EXM130.7.3.000-library-nocodes.json --root-lib-ref "Library-library-EXM130-7.3.000 -o`
- Save that `output.json` file.
- Compare the two files and make sure they are the same!

[LibraryDataRequirementsTestFiles.zip](https://github.com/projecttacoma/fqm-execution/files/10362517/LibraryDataRequirementsTestFiles.zip)

### DIFF
--- a/src/calculation/Calculator.ts
+++ b/src/calculation/Calculator.ts
@@ -516,9 +516,9 @@ export async function calculateGapsInCare<T extends OneOrMultiPatient>(
 }
 
 /**
- * Get data requirements for this library
+ * Get data requirements for a Library bundle
  *
- * @param libraryBundle Bundle with a LibraryResource and all necessary data for execution
+ * @param libraryBundle Bundle of library resources
  * @param options Options for calculation.
  *
  * @returns FHIR Library of data requirements
@@ -538,75 +538,6 @@ export async function calculateLibraryDataRequirements(
   );
 
   return DataRequirementHelpers.getDataRequirements(cqls, rootLibIdentifier, elmJSONs, options);
-
-  // const rootLib = elmJSONs.find(ej => ej.library.identifier == rootLibIdentifier);
-
-  // const { startCql, endCql } = Execution.getCQLIntervalEndpoints(options);
-
-  // // We need a root library to run dataRequirements properly. If we don't have one, error out.
-  // if (!rootLib?.library) {
-  //   throw new UnexpectedResource("root library doesn't contain a library object"); //unexpected resource
-  // }
-
-  // const parameters = { 'Measurement Period': new Interval(startCql, endCql) };
-
-  // const withErrors: GracefulError[] = [];
-  // // get the retrieves for every statement in the root lirbary
-  // const allRetrieves = rootLib.library.statements.def.flatMap(statement => {
-  //   if (statement.expression && statement.name != 'Patient') {
-  //     const retrievesOutput = RetrievesHelper.findRetrieves(rootLib, elmJSONs, statement.expression);
-  //     withErrors.push(...retrievesOutput.withErrors);
-  //     return retrievesOutput.results;
-  //   } else {
-  //     return [] as DataTypeQuery[];
-  //   }
-  // });
-
-  // const allRetrievesPromises = allRetrieves.map(async retrieve => {
-  //   // If the retrieves have a localId for the query and a known library name, we can get more info
-  //   // on how the query filters the sources.
-  //   if (retrieve.queryLocalId && retrieve.queryLibraryName) {
-  //     const library = elmJSONs.find(lib => lib.library.identifier.id === retrieve.queryLibraryName);
-  //     if (library) {
-  //       retrieve.queryInfo = await parseQueryInfo(
-  //         library,
-  //         elmJSONs,
-  //         retrieve.queryLocalId,
-  //         retrieve.valueComparisonLocalId,
-  //         parameters
-  //       );
-  //     }
-  //   }
-  // });
-
-  // await Promise.all(allRetrievesPromises);
-
-  // const results: fhir4.Library = {
-  //   resourceType: 'Library',
-  //   type: { coding: [{ code: 'module-definition', system: 'http://terminology.hl7.org/CodeSystem/library-type' }] },
-  //   status: 'unknown'
-  // };
-  // results.dataRequirement = uniqBy(
-  //   allRetrieves.map(retrieve => {
-  //     const dr = generateDataRequirement(retrieve);
-  //     GapsInCareHelpers.addFiltersToDataRequirement(retrieve, dr, withErrors);
-  //     addFhirQueryPatternToDataRequirements(dr);
-  //     return dr;
-  //   }),
-  //   JSON.stringify
-  // );
-
-  // return {
-  //   results: results,
-  //   debugOutput: {
-  //     cql: cqls,
-  //     elm: elmJSONs,
-  //     gaps: {
-  //       retrieves: allRetrieves
-  //     }
-  //   },
-  //   withErrors
-  // };
 }
 
 /**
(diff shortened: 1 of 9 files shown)


src/helpers/DataRequirementHelpers.ts
```diff
@@ -36,6 +45,84 @@ export function flattenFilters(filter: AnyFilter): AnyFilter[] {
   }
 }
 
+/**
+ * Returns a
```
sarahmcdougall, line 49: This should be updated - maybe something like "Returns FHIR library containing data requirements, given a root library"?

src/calculation/Calculator.ts
```diff
@@ -516,6 +515,31 @@ export async function calculateGapsInCare<T extends OneOrMultiPatient>(
...
+ */
+export async function calculateLibraryDataRequirements(
+  libraryBundle: fhir4.Bundle,
+  options: CalculationOptions = {}
+): Promise<DRCalculationOutput> {
+  if (options.rootLibRef === undefined) {
+    throw new UnexpectedProperty('Root lib ref must be provided in order to calculate library dataRequirements');
+  }
+
+  // Extract the library ELM, and the id of the root library, from the measure bundle
```
sarahmcdougall, line 534:
```suggestion
  // Extract the library ELM, and the id of the root library, from the library bundle
```

src/cli.ts
```diff
@@ -113,6 +118,8 @@ async function calc(
     result = await calculateGapsInCare(measureBundle, patientBundles, calcOptions, valueSetCache);
   } else if (program.outputType === 'dataRequirements') {
     result = calculateDataRequirements(measureBundle, calcOptions);
+  } else if (program.outputType === 'libraryDataRequirements') {
+    result = calculateLibraryDataRequirements(measureBundle, calcOptions);
```
sarahmcdougall, line 122: This is a little misleading since library data requirements does not make use of a measure. Would it be too much to rename `measureBundle` to `measureOrLibraryBundle`? I can't think of a better name for this lol. At the very least, we should add a comment about the bundle not containing a measure

src/helpers/MeasureBundleHelpers.ts
```diff
@@ -206,25 +206,51 @@ export function isValidLibraryURL(libraryName: string) {
...
-  const measure = extractMeasureFromBundle(measureBundle);
-  const rootLibRef = measure.library[0];
   let rootLibId: string;
   if (isValidLibraryURL(rootLibRef)) rootLibId = rootLibRef;
   else rootLibId = rootLibRef.substring(rootLibRef.indexOf('/') + 1);
 
+  const { cqls, rootLibIdentifier, elmJSONs } = extractLibrariesFromBundle(libraryBundle, rootLibId);
+
+  if (rootLibIdentifier.id === '') {
+    throw new UnexpectedResource('No Root Library could be identified in provided measure bundle');
```
sarahmcdougall, line 228:
```suggestion
    throw new UnexpectedResource('No Root Library could be identified in provided library bundle');
```

src/types/Calculator.ts
```diff
@@ -42,6 +42,8 @@ export interface CalculationOptions {
   useElmJsonsCaching?: boolean;
   /** if true, clears ElmJsons cache before running calculation */
   clearElmJsonsCache?: boolean;
+  /** Root lib ref, to be used in calculateLibraryDataRequirements */
```
sarahmcdougall, line 45: It might be more informative if we change "root lib ref" to "reference to the root library" (or something similar)

src/helpers/MeasureBundleHelpers.ts
sarahmcdougall, line 228: I'm seeing GH Actions for the error statement not being covered in a unit test - can we add a unit test for this, similar to the one that we have for `extractLibrariesFromMeasureBundle`?

test/unit/helpers/MeasureBundleHelpers.test.ts
```diff
@@ -788,10 +789,35 @@ describe('MeasureBundleHelpers tests', () => {
     });
   });
 
-  describe('extractLibrariesFromBundle', () => {
+  describe('extractLibrariesFromLibraryBundle', () => {
+    it('properly gets libraries from EXM130 library bundle', () => {
```
sarahmcdougall, line 793: I'll defer to @mgramigna on this - is this unit test sufficient, or should we instead try creating a simple bundle fixture that isn't EXM130? I know the unit tests for `extractLibrariesFromMeasureBundle` kinda set the precedent for this, but I think we should try to come up with a simpler example to use so that we start stepping away from using EXM130 for all the tests

sarahmcdougall (changes requested):
Thanks for the detailed testing guidance! Let me know if you have any questions on my requested changes!

src/helpers/DataRequirementHelpers.ts
```diff
@@ -36,6 +45,84 @@ export function flattenFilters(filter: AnyFilter): AnyFilter[] {
...
+  elmJSONs: ELM[],
+  options: CalculationOptions = {}
+): Promise<DRCalculationOutput> {
+  const rootLib = elmJSONs.find(ej => ej.library.identifier == rootLibIdentifier);
+
+  const { startCql, endCql } = Execution.getCQLIntervalEndpoints(options);
+
+  // We need a root library to run dataRequirements properly. If we don't have one, error out.
+  if (!rootLib?.library) {
+    throw new UnexpectedResource("root library doesn't contain a library object"); //unexpected resource
```
mgramigna, line 63:
```suggestion
    throw new UnexpectedResource("root library doesn't contain a library object");
```

src/types/Calculator.ts
mgramigna: We should also note that it _should_ be a canonical URL, but a resource ID will also work if a matching one exists in a bundle.

test/unit/helpers/MeasureBundleHelpers.test.ts
mgramigna: Yeah good question, we ultimately decided to just re-use the fixture due to the precedent already set in this file. We did discuss an end of sprint activity to scrub the fixtures of these unit tests and make them better, which would involve getting rid of EXM130.

Ultimately, I think since this just re-uses an existing fixture without creating a new one, it's fine, and would be updated along with all the other tests whenever that refactor gets done

test/unit/helpers/MeasureBundleHelpers.test.ts
```diff
@@ -788,10 +789,35 @@ describe('MeasureBundleHelpers tests', () => {
...
+      const measureBundle = getJSONFixture('EXM130-7.3.000-bundle-nocodes.json') as fhir4.Bundle;
+      const libraryBundle: fhir4.Bundle = {
+        resourceType: 'Bundle',
+        id: 'EXM130-7.3.000-bundle',
+        type: 'transaction'
+      };
+      libraryBundle.entry = measureBundle.entry?.filter(e => e.resource?.resourceType === 'Library');
+
+      const rootLibRef = 'Library/library-EXM130-7.3.000';
```
mgramigna, line 802: We should test root library extraction using canonical URLs as well, not just logical references.

The logical reference stuff is actually legacy support for how some old connectathon measures were defined. Using "url" as the root library reference is preferable, and conformant to the latest FHIR spec

mgramigna (commented):
Looks good. No huge change requests. Sucks that the helper is being yelled at for test coverage, but I think adding unit tests for the whole shebang might be a little out of scope. Idk

src/helpers/DataRequirementHelpers.ts
elsaperelli: Hahaha oops- thanks!

src/cli.ts
elsaperelli: Yeah I agree! I didn't want to make another input since it would've been a bit unnecessarily complicated but I can always do that instead.

test/unit/helpers/MeasureBundleHelpers.test.ts
sarahmcdougall: Cool sounds like a good plan 👍

sarahmcdougall (commented):
Changes look good! I forgot to include this in my first-round review, but we should make some updates to the README so it's up to date. I think this will include (1) adding an example of `calculateLibraryDataRequirements` to the "API" section, (2) adding `libraryDataRequirements` as a command in the "CLI" section, and (3) adding `root-lib-ref` as a CLI option in the "CLI" section. Otherwise everything looks good to me

mgramigna (approved):
Agreed with Sarah's readme updates, otherwise looks good

sarahmcdougall (approved):

PR #186, opened 2023-01-18
Upgrade cql-execution; Fix VSAC and ts-node config

### LINKED ISSUE #172: e.union is not a function
Good afternoon. The following error is being received from fqm-execution: 

   _e.union is not a function_

The measure bundle is available [here](https://github.com/srankins/ecqm-content-qicore-2022/blob/UnionNotRecognizedFunction/input/resources/atm-bundle-0.0.000.json). 

The patient bundle is [here](https://github.com/srankins/ecqm-content-qicore-2022/blob/UnionNotRecognizedFunction/input/tests/measure/test-bundle.json).

This is being run against [RC2 ](https://github.com/projecttacoma/fqm-execution/releases/tag/v1.0.0-rc.2) release.

I'm assuming this is bubbling up from cql-execution, but I thought I would start here. Bug or is this anticipated behavior?

mgramigna:
# Summary

Fixes #172 

This PR exposes the ValueSetResolver class that we use as an export from `fqm-execution` so it can be used in other libraries (main use case was use of it within a FHIR server). VSAC also changed their base URL recently for the FHIR API so I updated that too. I also updated the ts-node config to no longer require the use of `--files` option when testing the CLI

## New behavior

* Fixes issue brought up in #172 by upgrading the version of cql-execution (actual bug was fixed in https://github.com/cqframework/cql-execution/pull/306 thanks to @cmoesel)

## Code changes

* Upgrade cql-execution dependency
* Export ValueSetResolver from index.ts
* Fix canonical URL used for all VSAC requests, and included a link to where that URL is specified in case it breaks again in the future
* Move the `ts-node --files` option to `tsconfig` and update docs

# Testing guidance

Mainly just make sure the `npm run cli` command or using `ts-node` directly works without passing in the `--files` option. You can test valueset resolution if you want.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,7 +251,7 @@ Optionally, you can install the `ts-node` utility globally to execute the TypeSc
 
 ```bash
 npm install -g ts-node
-ts-node --files src/cli.ts [command] [options]
+ts-node src/cli.ts [command] [options]
 ```
 
 ### Debug Option
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "axios": "^0.21.1",
         "commander": "^6.1.0",
         "cql-exec-fhir": "^2.1.3",
-        "cql-execution": "^3.0.0-beta.2",
+        "cql-execution": "^3.0.0-beta.3",
         "handlebars": "^4.7.7",
         "lodash": "^4.17.21",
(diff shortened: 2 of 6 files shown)


elsaperelli (approved):
Lgtm! cli works

Issue #182, opened 2023-01-06
Single IPP Ratio Measure numerator/denominator populations are not treated independently


mgramigna:
Discussed in meeting with @srankins, @p9g, and others.

Per HQMF, ratio measure calculation should always treat numerator and denominator as independent populations. See below pseudo-cql from HQMF:

![image](https://user-images.githubusercontent.com/16297930/211086955-0fd373a9-ff07-4617-a5bd-c4b52146fe4b.png)

Some diagrams in the [cqfm IG](http://hl7.org/fhir/us/cqfmeasures/2023Jan/measure-conformance.html#ratio-measures) seem to be consistent with this, however as @p9g noted, the [pseudo-cql](http://hl7.org/fhir/us/cqfmeasures/2023Jan/measure-conformance.html#patient-based-calculation-1) does not align with HQMF.

Neither does fqm-execution, per the following example:

```cql
// snip
define "ipp":
  true

define "denom":
  false

define "numer":
  true
```

In the context of a ratio measure, "numer" should still be true despite "denom" being false. This is the current behavior in fqm-execution *only* when the measure has multiple IPPs, but fqm-execution should also behave this way even with a single IPP. Currently, it does not:

```json
[
  {
    "populationType": "initial-population",
    "criteriaExpression": "ipp",
    "result": true,
    "populationId": "ec813784-2eb4-46a9-b86a-bb03bf1c5b25"
  },
  {
    "populationType": "numerator",
    "criteriaExpression": "numer",
    "result": false,
    "populationId": "304711ea-6ded-4145-b429-20551850fc83"
  },
  {
    "populationType": "denominator",
    "criteriaExpression": "denom",
    "result": false,
    "populationId": "4c919f82-a385-4cda-9324-ae6c1722dec6"
  }
]
```

Measure Bundle and Patient Bundle for really simple demonstration of the issue: [issue.zip](https://github.com/projecttacoma/fqm-execution/files/10362543/issue.zip) (the above example)

[CMS871](https://github.com/cqframework/ecqm-content-r4-2021/tree/master/bundles/measure/HospitalHarmHyperglycemiainHospitalizedPatientsFHIR) is also an example of this issue in fqm-execution.

p9g:
https://github.com/HL7/cqf-measures/blob/master/archived/old_images/RatioVenn.jpg
https://github.com/HL7/cqf-measures/blob/master/archived/old_images/Ratio2Venn.png
[...]

PR #190, opened 2023-01-23
Independent numer/denom for ratio measures

sarahmcdougall:
# Summary
Addresses https://github.com/projecttacoma/fqm-execution/issues/182 (single IPP ratio measure numerator/denominator populations should be treated independently in fqm-execution).

## New behavior
Fqm-execution currently does not align with HQMF in the fact that ratio measure calculation should always treat numerator and denominator as independent populations. In fqm-execution, this currently only happens when the measure has multiple IPPs.

Fqm-execution now handles the numerator and denominator populations as independent populations for all ratio measures (so, including single IPP).

For example, we should now be able to achieve the following `populationResults`, meaning the numer is not automatically set to be false if the denom is false:

```
 "populationResults": [
          {
            "populationType": "initial-population",
            "criteriaExpression": "ipp",
            "result": true,
            "populationId": “ipp-id”
          },
          {
            "populationType": "numerator",
            "criteriaExpression": "numer",
            "result": true,
            "populationId": “numer-id”
          },
          {
            "populationType": "denominator",
            "criteriaExpression": "denom",
            "result": false,
            "populationId": “denom-id”
          }
        ]
```

## Code changes
Overall change: in `ClauseResultsBuilder` and `DetailedResultsBuilder`, check that the measure is a ratio measure (rather than checking that it has multiple IPPs) in spots where there is an assumed dependence between the numer and denom populations.

Note that the scoring code can occur at both the group and measure root level, so we should check for the presence of a scoring code in both spots. The measure scoring code is passed in to these functions as a fallback in case the scoring code does not appear at the group level. Open to feedback on a better way to pass in the measure scoring code… the only spot that I could see was passing it in through `calculate()` since the ClauseResultsBuilder/DetailedResultsBuilder  take in the measure group as input rather than the entire measure.

Code changes to focus on:
* Update `buildPopulationRelevanceMap` to check if measure is a ratio measure when deciding whether to set NUMER/NUMEX to false if DENOM is false
* Update `buildPopulationRelevanceMap` to check if measure is a ratio measure when deciding whether to se NUMER/NUMEX to false if DENEX is true
* Update `handlePopulationValues` to do similar checks based on scoring code as those in `buildPopulationRelevanceMap`
* Add unit test in DetailedResultsBuilder for checking numer/denom independence for single IPP ratio measure (plus some rearranging of unit tests - sorry the code diff is really ugly)
* Update unit tests in ClauseResultsBuilder to take in the new argument for measure scoring code

# Testing guidance
* Run unit tests
* Run detailed results against patient/measure bundle in https://github.com/projecttacoma/fqm-execution/issues/182 - on the main branch the numerator should false out, but on this branch the numerator should stay as “true” in the `populationResults`
* Run detailed results against the same patient/measure bundle but change the scoring code to “proportion” instead of “ratio” - the numerator should false out since it is dependent with the denominator (since this only applies to ratio measures)

### DIFF
--- a/src/calculation/ClauseResultsBuilder.ts
+++ b/src/calculation/ClauseResultsBuilder.ts
@@ -589,15 +589,15 @@ export function doesResultPass(result: any | null): boolean {
  * together so that a patient level population relevance set can be made.
  *
  * @private
- * @param {fhir4.MeasureGroup} populationGroup population group from measure
  * @param {EpisodeResults[]} episodeResultsSet - Results for each episode
- * @param {string | null} measureScoringCode scoring code for measure (used if scoring code not provided at the group level)
+ * @param {fhir4.MeasureGroup} populationGroup population group from measure
+ * @param {string} measureScoringCode scoring code for measure (used if scoring code not provided at the group level)
  * @returns {PopulationResult[]} List denoting if population calculation was considered/relevant in any episode
  */
 export function buildPopulationRelevanceForAllEpisodes(
-  populationGroup: fhir4.MeasureGroup,
   episodeResultsSet: EpisodeResults[],
-  measureScoringCode: string | null
+  populationGroup: fhir4.MeasureGroup,
+  measureScoringCode?: string
 ): PopulationResult[] {
   const masterRelevanceResults: PopulationResult[] =
     populationGroup.population?.map(population => {
@@ -611,8 +611,8 @@ export function buildPopulationRelevanceForAllEpisodes(
   episodeResultsSet.forEach(episodeResults => {
     const episodeRelevance = buildPopulationRelevanceMap(
       episodeResults.populationResults,
-      measureScoringCode,
-      populationGroup
+      populationGroup,
+      measureScoringCode
     );
     masterRelevanceResults.forEach(masterPopResults => {
       // find relevance in episode and if true, make master relevance true, only if not already true
@@ -653,14 +653,14 @@ export function buildPopulationRelevanceForAllEpisodes(
  * was kept to make it more maintainable.
  *
  * @param {PopulationResult[]} results - The population results list for the population results.
- * @param {string | null} measureScoringCode scoring code for measure (used if scoring code not provided at the group level)
  * @param {fhir4.MeasureGroup} group - The full group of the Measure, which is useful for resolving references between different populations
+ * @param {string} measureScoringCode scoring code for measure (used if scoring code not provided at the group level)
  * @returns {PopulationResult[]} The population relevance set.
  */
 export function buildPopulationRelevanceMap(
   results: PopulationResult[],
-  measureScoringCode: string | null,
-  group?: fhir4.MeasureGroup
+  group?: fhir4.MeasureGroup,
+  measureScoringCode?: string
 ): PopulationResult[] {
   // Create initial results starting with all true to create the basis for relevance.
   const relevantResults: PopulationResult[] = results.map(result => {
@@ -810,16 +810,16 @@ export function buildPopulationRelevanceMap(
 
 export function buildPopulationGroupRelevanceMap(
   results: DetailedPopulationGroupResult,
-  measureScoringCode: string | null,
-  group: fhir4.MeasureGroup
+  group: fhir4.MeasureGroup,
+  measureScoringCode?: string
 ): PopulationResult[] {
   // Episode of care measure
   if (results.episodeResults) {
-    return buildPopulationRelevanceForAllEpisodes(group, results.episodeResults, measureScoringCode);
+    return buildPopulationRelevanceForAllEpisodes(results.episodeResults, group, measureScoringCode);
 
     // Normal patient based measure
   } else if (results.populationResults) {
-    return buildPopulationRelevanceMap(results.populationResults, measureScoringCode, group);
+    return buildPopulationRelevanceMap(results.populationResults, group, measureScoringCode);
   } else {
     // this shouldn't happen
     return [];
--- a/src/calculation/ClauseResultsBuilder.ts
+++ b/src/calculation/ClauseResultsBuilder.ts
@@ -6,7 +6,7 @@ import * as cql from '../types/CQLTypes';
 import { Interval, DateTime, Code, Quantity } from 'cql-execution';
 import moment from 'moment';
 
-import { FinalResult, MeasureScoreType, PopulationType, Relevance } from '../types/Enums';
+import { FinalResult, PopulationType, Relevance } from '../types/Enums';
 import {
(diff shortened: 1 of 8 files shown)


src/calculation/ClauseResultsBuilder.ts
```diff
@@ -736,7 +753,14 @@ export function buildPopulationRelevanceMap(
...
   if (hasResult(PopulationType.DENEX, results) && getResult(PopulationType.DENEX, results) === true) {
-    if (!(group && MeasureBundleHelpers.hasMultipleIPPs(group))) {
+    // Do not apply this to ratio measures (numerator and denominator are independent from each other)
+    if (
+      !(
+        group &&
+        (MeasureBundleHelpers.getScoringCodeFromGroup(group) === MeasureScoreType.RATIO ||
+          measureScoringCode === MeasureScoreType.RATIO)
+      )
+    ) {
```
elsaperelli, line 763: I am having a hard time with this logic- 

We don't want to go into this if statement if it's a ratio measure, right?

If group doesn't exist but measureScoringCode === MeasureScoreType.RATIO, then the && calculates to false which then !false = true so it _does_ go into the if statement even though I don't think we want to since it's a ratio measure? Am I doing this logic correctly?
If group exists and its a ratio measure, then

test/unit/DetailedResultsBuilder.test.ts
```diff
@@ -583,7 +585,7 @@ describe('DetailedResultsBuilder', () => {
       );
     });
 
-    describe('multiple IPPs', () => {
+    describe('handlePopulationValues for group with observations and multiple IPPs', () => {
```
elsaperelli, line 588: Add something here about ratio measures?

elsaperelli (changes requested):
Nice work on this bug! I just have a couple comments and questions.

One overall thing I noticed though was there don't seem to be any unit tests for testing numer/denom independence when there is only 1 ipp and its a ratio measure. This should maybe be something we add unless it's there and I just missed it.

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -589,12 +589,15 @@ export function doesResultPass(result: any | null): boolean {
  * together so that a patient level population relevance set can be made.
  *
  * @private
+ * @param {fhir4.MeasureGroup} populationGroup population group from measure
  * @param {EpisodeResults[]} episodeResultsSet - Results for each episode
+ * @param {string | null} measureScoringCode scoring code for measure (used if scoring code not provided at the group level)
```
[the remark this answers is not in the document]
mgramigna: Yeah I've encountered stuff like this before. It's because the helper function itself returns `string | null`. (undefined is not assignable to type null).

I'm not opposed to making the helper functions return `string | undefined` instead if it makes these function signatures a little easier to understand.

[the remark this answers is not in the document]
sarahmcdougall: Cool I'll try reworking the helper function to return `string | undefined` which should simplify a lot of things

sarahmcdougall:
> One overall thing I noticed though was there don't seem to be any unit tests for testing numer/denom independence when there is only 1 ipp and its a ratio measure. This should maybe be something we add unless it's there and I just missed it.

They are there (they probably got overshadowed by the rest of the code diff since I made a lot of little changes surrounding them) - the unit tests are in a describe block at the bottom of `DetailedResultsBuilder.test.ts`: 'handlePopulationValues for group with single IPP - ratio measure'

src/helpers/MeasureBundleHelpers.ts
```diff
@@ -5,6 +5,7 @@ import { UnexpectedProperty, UnexpectedResource } from '../types/errors/CustomEr
 import { getMissingDependentValuesets } from '../execution/ValueSetHelper';
 import { ValueSetResolver } from '../execution/ValueSetResolver';
 import { ExtractedLibrary } from '../types/CQLTypes';
+import { MeasureBundleHelpers } from '..';
```
mgramigna, line 8: Looks like this file is importing itself? You should be able to just call the local function below rather than importing it and using this as a prefix

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -736,7 +747,8 @@ export function buildPopulationRelevanceMap(
 
   // If DENEX is true then NUMER, NUMEX and DENEXCEP not calculated
   if (hasResult(PopulationType.DENEX, results) && getResult(PopulationType.DENEX, results) === true) {
-    if (!(group && MeasureBundleHelpers.hasMultipleIPPs(group))) {
+    // Do not apply this to ratio measures (numerator and denominator are independent from each other)
+    if (!MeasureBundleHelpers.isRatioMeasure(group, measureScoringCode)) {
```
mgramigna, line 751: I think there's one more case of these that is missing (down on line 772).

There is a block of code that falses out the DENEXCEP if numer is true, but we should not do this for ratio measures according to the Venn D since DENEXCEP is a subset of the denom which is independent from the numer

mgramigna (changes requested):

src/helpers/MeasureBundleHelpers.ts
sarahmcdougall: this was vscode not me I swear... apologies for letting this get pushed this is so embarrassing

src/calculation/ClauseResultsBuilder.ts
elsaperelli: Dang these were the two changes I was going to request hahah

elsaperelli (changes requested):
Just two small changes that it looks like Matt beat me to! Nice job with the new helper function I think that makes things look much better!

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -757,7 +769,11 @@ export function buildPopulationRelevanceMap(
   }
 
   // If NUMER is true then DENEXCEP is not calculated
-  if (hasResult(PopulationType.NUMER, results) && getResult(PopulationType.NUMER, results) === true) {
+  if (
+    hasResult(PopulationType.NUMER, results) &&
+    getResult(PopulationType.NUMER, results) === true &&
+    !MeasureBundleHelpers.isRatioMeasure(group, measureScoringCode)
```
elsaperelli, line 775: My only suggestion would be to maybe add a comment there about the ratio measure stuff like you did above (line 750) but that's pretty picky haha

elsaperelli (approved):
One optional suggestion otherwise looks great :)

mgramigna (approved):
Great work on this

PR #192, opened 2023-01-25
Removed default measurement period for dataRequirements

natjoe4:
# Summary
We've been using 2019-01-01 -> 2019-12-31 as our default interval if one is not passed-in as a calculation option. Now, when one is not provided, we use the effectivePeriod of the Measure, or, when absent, simply pass in no measurement period.

## New behavior
data-requirements operations will no longer default to a 2019 measurement period when one is not provided. Instead, the following will happen: 
 - If a measurement period start and end calculation option are provided, this period will be used
 - If a measurement period start calculation option is provided with no end, a period with duration 1 year beginning at the provided start will be used
 - If a measurement period end calculation option is provided with no start, a period with duration 1 year ending at the provided end will be used
 - If no measurement period start or end calculation option is provided, the measurement period will be determined by the effective period of the FHIR Measure, using the same logic as above to handle missing start and end dates
 - If no measurement period start or end calculation option is provided and no effectivePeriod is found on the measure, no measurement period will be used, and all dateFilters will be excluded

## Code changes
 - updated `getDataRequirements` function to take an additional `effectivePeriod` argument
 - created `extractDataRequirementsMeasurementPeriod` function
 - created `createIntervalFromEndpoints` function
 - created `hasMeasurementPeriodInfo` function
 - added testing

# Testing guidance
 - `npm run check`
 - download the included .zip file
 - unzip the file and copy the contents into the root level of `fqm-execution`
 - `sh run.sh`
 - now inspect each output file in `test-outputs` and ensure that the `dateFilters` all line up with the logic outlined in the new behavior section
 - Please reach out with questions!

[test-data.zip](https://github.com/projecttacoma/fqm-execution/files/10503813/test-data.zip)

### DIFF
--- a/src/gaps/QueryFilterParser.ts
+++ b/src/gaps/QueryFilterParser.ts
@@ -51,7 +51,8 @@ import {
   UnknownFilter,
   ValueFilter,
   ValueFilterComparator,
-  IsNullFilter
+  IsNullFilter,
+  QueryParserParams
 } from '../types/QueryFilterTypes';
 import { findLibraryReference, findLibraryReferenceId } from '../helpers/elm/ELMDependencyHelpers';
 import { findClauseInLibrary } from '../helpers/elm/ELMHelpers';
@@ -79,7 +80,7 @@ export async function parseQueryInfo(
   allELM: ELM[],
   queryLocalId?: string,
   valueComparisonLocalId?: string,
-  parameters: Record<string, any> = {},
+  parameters: QueryParserParams = {},
   patient?: CQLPatient
 ): Promise<QueryInfo> {
   if (!queryLocalId) {
--- a/src/gaps/QueryFilterParser.ts
+++ b/src/gaps/QueryFilterParser.ts
@@ -79,7 +79,7 @@ export async function parseQueryInfo(
   allELM: ELM[],
   queryLocalId?: string,
   valueComparisonLocalId?: string,
-  parameters: { [key: string]: any } = {},
+  parameters: Record<string, any> = {},
   patient?: CQLPatient
 ): Promise<QueryInfo> {
   if (!queryLocalId) {
--- a/src/helpers/DataRequirementHelpers.ts
+++ b/src/helpers/DataRequirementHelpers.ts
@@ -23,7 +23,7 @@ import * as GapsInCareHelpers from '../gaps/GapsReportBuilder';
 import { parseQueryInfo } from '../gaps/QueryFilterParser';
 import * as RetrievesHelper from '../gaps/RetrievesFinder';
 import { uniqBy } from 'lodash';
-import { Interval } from 'cql-execution';
+import { DateTime, Interval } from 'cql-execution';
 import { parseTimeStringAsUTC } from '../execution/ValueSetHelper';
 const FHIR_QUERY_PATTERN_URL = 'http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqfm-fhirQueryPattern';
 
@@ -392,6 +392,10 @@ export function codeLookup(dataType: string, attribute: string): string | null {
   return null;
 }
 
+/**
+ * Extracts the measurement period information from either the options or effective period (in that order depending on presence)
+ * and populates a parameters object including the extracted info to be passed into the parseQueryInfo function
+ */
 export function extractDataRequirementsMeasurementPeriod(options: CalculationOptions, effectivePeriod?: fhir4.Period) {
   if (!hasMeasurementPeriodInfo(options, effectivePeriod)) {
     return {};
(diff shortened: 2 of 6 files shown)


src/gaps/QueryFilterParser.ts
```diff
@@ -79,7 +79,7 @@ export async function parseQueryInfo(
   allELM: ELM[],
   queryLocalId?: string,
   valueComparisonLocalId?: string,
-  parameters: { [key: string]: any } = {},
+  parameters: Record<string, any> = {},
```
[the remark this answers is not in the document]
elsaperelli: On further inspection, while I think the Record is much better than what it was, just out of curiosity, why not make this stricter? This was what it was previously so I may have missed the reason why but if "MeasurementPeriod" is the only supported parameter at the moment, then could we change this type? Or at least make it a Record<string, fhir4.Interval> ? Or do we anticipate having more parameters soon

src/helpers/DataRequirementHelpers.ts
```diff
@@ -391,3 +391,57 @@ export function codeLookup(dataType: string, attribute: string): string | null {
...
+    parameters['Measurement Period'] = createIntervalFromEndpoints(effectivePeriod?.start, effectivePeriod?.end);
+  }
+  return parameters;
+}
+
+/**
+ * Creates a cql-execution interval from start to end. If either start or end is not present,
+ * creates an interval with duration exactly one year using the present endpoint
+ */
+export function createIntervalFromEndpoints(start?: string, end?: string) {
```
elsaperelli, line 420: Out of curiosity, before our default was from 2019-01-01 to 2019-12-31 but now we have intervals from the first of one year to the first of the next (2019-01-01 to 2020-01-01 for example). Was there a reason for this change?

src/helpers/DataRequirementHelpers.ts
```diff
@@ -391,3 +391,57 @@ export function codeLookup(dataType: string, attribute: string): string | null {
...
+    }
+    startCql = DateTime.fromJSDate(startCql, 0);
+    endCql = DateTime.fromJSDate(endCql, 0);
+  }
+  return new Interval(startCql, endCql);
+}
+
+function hasMeasurementPeriodInfo(options: CalculationOptions, effectivePeriod?: fhir4.Period) {
+  return (
+    options.measurementPeriodStart || options.measurementPeriodEnd || effectivePeriod?.start || effectivePeriod?.end
```
elsaperelli, line 445: I have no problem with this being a separate function but do we think it's necessary? Looks like we only use it in one spot and the only reason really why I am asking is it looks like this Branch is not covered through GitHub Actions and adding a test unit for it might just be unnecessary.

test/unit/DataRequirementHelpers.test.ts
```diff
@@ -625,4 +627,65 @@ describe('DataRequirementHelpers', () => {
...
+      expect(DataRequirementHelpers.extractDataRequirementsMeasurementPeriod({}, {})).toEqual({});
+    });
+
+    test('uses measurement period from options when both provided', () => {
+      const start = '2019-01-01';
+      const end = '2020-01-01';
+      const startCql = DateTime.fromJSDate(moment.utc(start, 'YYYYMDDHHmm').toDate(), 0);
+      const endCql = DateTime.fromJSDate(moment.utc(end, 'YYYYMDDHHmm').toDate(), 0);
+      const options: CalculationOptions = { measurementPeriodStart: start, measurementPeriodEnd: end };
+      const mp: fhir4.Period = { start: '2000-01-01', end: '2001-01-01' };
```
elsaperelli, line 642:
```suggestion
      const mp: fhir4.Period = { start: '2000-01-01', end: '2001-01-01' };
      
```

test/unit/DataRequirementHelpers.test.ts
```diff
@@ -625,4 +627,65 @@ describe('DataRequirementHelpers', () => {
...
+    test('returns interval from start to end if both provided', () => {
+      const start = '2019-01-01';
+      const end = '2022-01-01';
+      const startCql = DateTime.fromJSDate(moment.utc(start, 'YYYYMDDHHmm').toDate(), 0);
+      const endCql = DateTime.fromJSDate(moment.utc(end, 'YYYYMDDHHmm').toDate(), 0);
+
+      expect(DataRequirementHelpers.createIntervalFromEndpoints(start, end)).toEqual(new Interval(startCql, endCql));
+    });
+
+    test('returns interval from start with duration 1 year if end not provided provided', () => {
```
elsaperelli, line 671:
```suggestion
    test('returns interval from start with duration 1 year if end not provided', () => {
```

test/unit/DataRequirementHelpers.test.ts
```diff
@@ -625,4 +627,65 @@ describe('DataRequirementHelpers', () => {
...
+      const start = '2019-01-01';
+      const expectedEnd = '2020-01-01';
+      const startCql = DateTime.fromJSDate(moment.utc(start, 'YYYYMDDHHmm').toDate(), 0);
+      const endCql = DateTime.fromJSDate(moment.utc(expectedEnd, 'YYYYMDDHHmm').toDate(), 0);
+
+      expect(DataRequirementHelpers.createIntervalFromEndpoints(start)).toEqual(new Interval(startCql, endCql));
+    });
+
+    test('returns interval up to end with duration 1 year if end not provided provided', () => {
```
elsaperelli, line 680:
```suggestion
    test('returns interval up to end with duration 1 year if start not provided', () => {
```

test/unit/DataRequirementHelpers.test.ts
```diff
@@ -625,4 +627,65 @@ describe('DataRequirementHelpers', () => {
...
+      expect(DataRequirementHelpers.createIntervalFromEndpoints(start, end)).toEqual(new Interval(startCql, endCql));
+    });
+
+    test('returns interval from start with duration 1 year if end not provided provided', () => {
+      const start = '2019-01-01';
+      const expectedEnd = '2020-01-01';
+      const startCql = DateTime.fromJSDate(moment.utc(start, 'YYYYMDDHHmm').toDate(), 0);
+      const endCql = DateTime.fromJSDate(moment.utc(expectedEnd, 'YYYYMDDHHmm').toDate(), 0);
+
+      expect(DataRequirementHelpers.createIntervalFromEndpoints(start)).toEqual(new Interval(startCql, endCql));
```
elsaperelli, line 677:
```suggestion
      expect(DataRequirementHelpers.createIntervalFromEndpoints(start, undefined)).toEqual(new Interval(startCql, endCql));
```
Thoughts? Just makes things more clear but idk

elsaperelli (changes requested):
This is really great work. Lots of attention to detail in testing- made life really easy! Just some very small comments

src/helpers/DataRequirementHelpers.ts
sarahmcdougall, line 445: This is really minor but the function returns a string even tho it's called `hasMeasurementPeriodInfo`, which I would assume returns a boolean. Thoughts on making this return a boolean instead (if any exist, return true, o/w return false)?

sarahmcdougall (commented):
Woohoo great work! tysm for creating a script for all the different scenarios, it was super helpful for testing :) I just have one comment but otherwise should be good to merge after making Elsa's changes

src/helpers/DataRequirementHelpers.ts
natjoe4: In all honesty just because `setYear(getYear() + 1)` was easier to type than `setYear(getYear() + 1).setDays(getDays() - 1)`. I can't imagine any consequences for adding a day to our default measurement period, but happy to change it if there's something I'm missing :)

src/helpers/DataRequirementHelpers.ts
natjoe4: Lol I totally thought this would return a boolean... Good catch

src/helpers/DataRequirementHelpers.ts
elsaperelli: I agree I don't see anything wrong with it I was just making sure that was intentional!

src/helpers/DataRequirementHelpers.ts
natjoe4: Yeah I really just did this for readability since its kinda an ugly boolean statement that ultimately accomplishes something really simple. I could convert to const, or if you think it's more readable to just insert the boolean into the function that calls this, happy to do that also :)

elsaperelli: Gotcha yeah I agree I mostly just pointed this out for branch coverage awareness but not a huge deal at all! Perfect as is as long as branch coverage isn't a concern

elsaperelli (approved):
Nice job!

sarahmcdougall (approved):
yay looks good!

PR #196, opened 2023-02-03
Added issue template

mgramigna:


natjoe4 (approved):
This looks great and should streamline the issue reporting process. My one, optional, comment is that a lot of the semantic bits issue requests have included screenshots, which I at least have found helpful when addressing issues. Could we add a designated `Evidence` or `Supporting Documents`, or even just `Output` section for screenshots and JSON files? Or do you think it is self-explanatory to include those in `Summary` or `Expected Behavior`

mgramigna:
> This looks great and should streamline the issue reporting process. My one, optional, comment is that a lot of the semantic bits issue requests have included screenshots, which I at least have found helpful when addressing issues. Could we add a designated `Evidence` or `Supporting Documents`, or even just `Output` section for screenshots and JSON files? Or do you think it is self-explanatory to include those in `Summary` or `Expected Behavior`

I think it's mostly covered by the summer/expected behavior, but maybe I'll add a catchall "any additional info" section or something

PR #199, opened 2023-02-14
Fix ratio function reuse numer only

### LINKED ISSUE #187: Ratio, episode-based measure observation results, for same function, not appearing correctly
After upgrading to 1.0.3 (measure bundle and test case are in the attached tar.gz file)

We are executing a testcase where denominator and numerator results in incorrect "results"

With fqm-execution 1.0.2, this scenario results in correct results for Denom, but no results for Num

Now, with fqm-execution 1.0.3, this scenario results in incorrect results for Denom and correct results for Num

In this case, the IPP is where encounter is in a period, the DENOM is equal to IPP.  The NUMER is where Encounter.status == finished

So, for this test case, all Encounters are in the period.  So IPP = 3 & DENOM = 3.  Only one Encounter (Encounter 1) have a status of finished, so NUMER should = 1

There are Observations for DENOM & NUMER where each Observation is fun3 ( which is hardcoded to = 3)

We expect the results to show measure observation matches for denom & numer on Encounter 1 and measure observation matches for denom only for Encounter 2 & 3

The results for this execution show Encounter 1 matches as expected on denom & num as expected 

image

However, for Encounter 2 & 3 (which should match denom, but not numer), the results show

image

[measure_files.tar.gz](https://github.com/projecttacoma/fqm-execution/files/10448855/measure_files.tar.gz)

elsaperelli:
# Summary
Fixes #187 
This pull request (by @hossenlopp) fixes the bug in [#187](https://github.com/projecttacoma/fqm-execution/issues/187).
Fixes ratio, episode-based measure observation results, for same function, not appearing correctly. Now, we ensure that observations are not nulled out by finding the populationResult whose criteriaExpression matches the desiredPopulation criteriaExpression AND whose populationId matches the desiredPopulation id.

## New behavior
Before, in `getObservationResultForPopulation()` in `MeasureBundleHelpers.ts`, we found the `populationResult` of the measure observation that references the desired population in its criteria reference by doing a `.find` on populationResults and looking for the result whose criteriaExpression matched the criteriaExpression of the desiredObservation. In cases where populationResults had the same criteriaExpression, this could lead to issues because the `.find` would result in returning the first result that had a matching criteriaExpression. In order to fix this, we added a check to the `.find` to ensure that the population id of the population result matched the population id of the desired observation. If the populationId does not exist in the population results, then `true` is returned. 

## Code changes
- `src/helpers/MeasureBundleHelpers.ts` - where the bug fix occurs
- `test/integration/helpers/testHelpers.ts` - add `| undefined` to `observations` and add check for `observations` in `assertObservations()`
- `test/integration/ratio-Encounter-reuseObservationFunction/patients` - additional patient bundles for integration testing
- `test/integration/ratio-Encounter-reuseObservationFunction/ratio-Encounter-reuseObservationFunction.test.ts` - additional integration tests for the new patient bundles

# Testing guidance
- `npm run test:integration`
- `npm run check`


test/integration/helpers/testHelpers.ts
```diff
@@ -105,7 +105,7 @@ function getPopulationResult(
 export function assertEpisodeObservations(
   result: DetailedPopulationGroupResult,
   episodeId: string,
-  observations: any[],
+  observations: any[] | undefined,
```
mgramigna, line 108:
```suggestion
  observations?: any[],
```

test/integration/helpers/testHelpers.ts
```diff
@@ -124,7 +124,7 @@ export function assertEpisodeObservations(
  */
 export function assertGroupObservations(
   result: DetailedPopulationGroupResult,
-  observations: any[],
+  observations: any[] | undefined,
```
mgramigna, line 127:
```suggestion
  observations?: any[],
```

test/integration/helpers/testHelpers.ts
```diff
@@ -137,7 +137,7 @@ export function assertGroupObservations(
  */
 function assertObservations(
   populationResults: PopulationResult[],
-  observations: any[],
+  observations: any[] | undefined,
```
mgramigna, line 140:
```suggestion
  observations?: any[],
```

mgramigna (commented):
LGTM, only one code styling comment in the tests

mgramigna (approved):

natjoe4 (approved):
Has to be the biggest `functional code / testing code` ratio I've ever seen in a PR. Nice work :)

PR #201, opened 2023-02-16
(integration tests): fix proportion boolean bundling and allow for force rebuilding

mgramigna:
# Summary

The existing `make` target for the proportion boolean integration test didn't work. This was due to the addition of MATGlobalCommonFunctions which included valuesets that ecqm-bundler couldn't resolve.

My solution for this was to just remove this library and uses of it for simplicity. It required deleting one test.

I think one of the reasons that this happened was the `git status --porcelain ...` command used in the Makefile to only re-build bundles that have a git diff in their main CQL. I added an option to force rebuilding of bundles which will be useful for testing and testing new updates of ecqm bundler.

## New behavior

* It works now
* Can use `FORCE=true npm run build:test-data` to force a rebuild of the test data

## Code changes

* Add detection of `FORCE=` environment var for Make targets. Presence of it will re-build the bundles for everything regardless of if the CQL has been modified
  * Update README in integration directory to describe this option
* Update to ecqm-bundler 0.2.2
* Remove use of MATGlobalCommon from proportion boolean test
  * IMO we prob just shouldn't use this. I'm amenable to this though and can add it back, but in that case I will remove the value set defs and any expressions that use valuesets, just keeping the helper functions (like normalize interval)
* Re-bundle proportion boolean test

# Testing guidance

* `npm run build:test-data` should do nothing
* `FORCE=true npm run build:test-data` should re-build everything
* `npm run test:integration should work both pre and post-rebuild


sarahmcdougall (approved):
Good catch with this PR - I remember when we added `MATGlobalCommonFunctions` I assumed it had no repercussions because I didn't see anything rebuild... makes sense now why it looked like nothing was affected. I followed the testing guidance and the results are what we expect. Good documentation additions as well.

In regard to whether or not we should remove the test that uses `MATGlobalCommonFunctions` (which I'll admit I had to google what amenable means so thank you for expanding my vocabulary), I think we should remove this test and library for simplicity, and my reasoning for this is the context of the test being removed. I don't think we necessarily _need_ to test an earlier measurement period in the integration tests, and I favor simplicity over having this test. If we see a stronger need for this test in the future (or if we have more tests get created for other measures that need the library), then we can consider adding the library back.

elsaperelli (approved):
Lgtm!

PR #206, opened 2023-03-03
Library data requirements should bypass query filter parsing

sarahmcdougall:
# Summary
Ensures query filter parsing is only performed on measure data requirements and not library data requirements.

## New Behavior
When running library data requirements, the detailed query filter parsing will no longer be performed. Library data requirements does not make use of a measurement period because there is no measure, and so it does not make sense to do detailed query filter parsing since a lot of the logic relies on constructing intervals for the (non-existent) measurement period.

## Code Changes
* Add check in the conditional of`getDataRequirements` for whether a measurement period is passed in (since it should only be passed in from `dataRequirements` and not `libraryDataRequirements`)
* `calculateLibraryDataRequirements` now omits `measurementPeriodStart`/`measurementPeriodEnd` if they are passed in as CLI options (we could change this to throw an error if that is more appropriate)

# Testing Guidance
* Run unit tests
* Run library data requirements on the master branch and compare the output to the results on this branch to see the effects of not performing detailed query filter parsing for library data requirements.
    * See the testing guidance in  https://github.com/projecttacoma/fqm-execution/pull/181 for an example library bundle and CLI command that can used for running library data requirements.
    * A good way to go about this is to save the output from the master branch to a file, then save the output from this branch to a file, and then use VSCode’s “Select for Compare”/“Compare with Selected” to see the code diff

### DIFF
--- a/src/calculation/Calculator.ts
+++ b/src/calculation/Calculator.ts
@@ -38,6 +38,7 @@ import { Interval, DataProvider } from 'cql-execution';
 import { PatientSource } from 'cql-exec-fhir';
 import { pruneDetailedResults } from '../helpers/DetailedResultsHelpers';
 import { clearElmInfoCache } from '../helpers/elm/ELMInfoCache';
+import { omit } from 'lodash';
 
 /**
  * Calculate measure against a set of patients. Returning detailed results for each patient and population group.
@@ -532,6 +533,9 @@ export async function calculateLibraryDataRequirements(
   libraryBundle: fhir4.Bundle,
   options: CalculationOptions = {}
 ): Promise<DRCalculationOutput> {
+  // omit measurementPeriodStart/measurementPeriodEnd since there is no measure
+  options = omit(options, 'measurementPeriodStart', 'measurementPeriodEnd');
+
   if (options.rootLibRef === undefined) {
     throw new UnexpectedProperty('Root lib ref must be provided in order to calculate library dataRequirements');
   }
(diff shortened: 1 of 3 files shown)


src/calculation/Calculator.ts
```diff
@@ -532,6 +533,9 @@ export async function calculateLibraryDataRequirements(
   libraryBundle: fhir4.Bundle,
   options: CalculationOptions = {}
 ): Promise<DRCalculationOutput> {
+  // omit measurementPeriodStart/measurementPeriodEnd since there is no measure
```
elsaperelli, line 536: I think this is good and we don't need to throw an error, but should we have a warning that the measurement period is omitted? That way a user isn't confused by the output? A warning might be unnecessary so in that case maybe just adding something small to the README that addresses this.

elsaperelli (approved):
Looks good to me! I just have some suggestion but I'm open to debate on that so otherwise approved!

src/calculation/Calculator.ts
natjoe4: Yeah I agree. I think best course would just be to add a note to the documentation (and maybe the cli descriptions as well) that says measurementPeriodStart and measurementPeriodEnd are ignored for libraryDataRequirements

natjoe4 (changes requested):
This looks great! Nice work. I want to point out that this code change also turns off queryFilterParsing for regular `calculateDataRequirements` calls when no measurement period is provided by the user or found in the measure. Is this the expected behavior? Otherwise, just minimal comments :)

src/calculation/Calculator.ts
sarahmcdougall: Good call - I think we should definitely add something to the documentation. Thoughts on adding a `console.warn` as well or do we think that's unnecessary?

sarahmcdougall:
> I want to point out that this code change also turns off queryFilterParsing for regular `calculateDataRequirements` calls when no measurement period is provided by the user or found in the measure. Is this the expected behavior?

Yes this is expected. Based on this [Slack thread](https://mitrecorp.slack.com/archives/CD894AETZ/p1677088041502479) there exists at least one measure type that does not require `effectivePeriod`, but it is unlikely that we will be working with those measures, and if we are, providing the measurement period via the Calculation Options should allow us to still do the query filter parsing. Otherwise, we will bypass the query filter parsing for the measure.

src/calculation/Calculator.ts
natjoe4: I think that's probably unnecessary. We do use `console.warn` in fqm-execution, but not widely, and I think a documentation update is enough info here.

natjoe4 (approved):

elsaperelli (approved):
Lgtm!

PR #207, opened 2023-03-07
Age highlighting issue

### LINKED ISSUE #202: Execution returns wrong highlighting and no data for AgeInYearsAt function
### Expected Behavior

AgeInYearsAt(date from start of "Measurement Period")>= 66 should be highlighted red, but it's green.

```json
{
   "resourceType":"Bundle",
   "id":"Patient-1-IPPass-66WithQualEnc",
   "type":"collection",
   "entry":[
      {
         "fullUrl":"633c9d020968f8012250fc60",
         "resource":{
            "resourceType":"Patient",
            "id":"Patient-1",
            "meta":{
               "profile":[
                  "http://hl7.org/fhir/us/qicore/StructureDefinition/qicore-patient"
               ]
            },
            "extension":[
               {
                  "url":"http://hl7.org/fhir/us/core/StructureDefinition/us-core-race",
                  "extension":[
                     {
                        "url":"ombCategory",
                        "valueCoding":{
                           "code":"2028-9",
                           "system":"urn:oid:2.16.840.1.113883.6.238",
                           "display":"Asian"
                        }
                     },
                     {
                        "url":"text",
                        "valueString":"Asian"
                     }
                  ]
               },
               {
                  "url":"http://hl7.org/fhir/us/core/StructureDefinition/us-core-ethnicity",
                  "extension":[
                     {
                        "url":"ombCategory",
                        "valueCoding":{
                           "code":"2135-2",
                           "system":"urn:oid:2.16.840.1.113883.6.238",
                           "display":"Hispanic or Latino"
                        }
                     },
                     {
                        "url":"text",
                        "valueString":"Hispanic or Latino"
                     }
                  ]
               }
            ],
            "identifier":[
               {
                  "system":"http://hospital.smarthealthit.org",
                  "value":"999999995"
               }
            ],
            "name":[
               {
                  "family":"Bertha",
                  "given":[
                     "Betty"
                  ]
               }
            ],
            "birthDate":"1957-01-02",
            "gender":"female"
         }
      },
      {
         "fullUrl":"633c9d020968f8012250fc61",
         "resource":{
            "resourceType":"Encounter",
            "id":"Encounter-1",
            "meta":{
               "profile":[
                  "http://hl7.org/fhir/us/qicore/StructureDefinition/qicore-encounter"
               ]
            },
            "subject":{
               "reference":"Patient/Patient-1"
            },
            "status":"finished",
            "class":{
               "code":"AMB",
               "system":"http://terminology.hl7.org/CodeSystem/v3-ActCode",
               "display":"ambulatory"
            },
            "type":[
               {
                  "coding":[
                     {
                        "code":"185317003",
                        "system":"http://snomed.info/sct",
                        "display":"Telephone encounter (procedure)"
                     }
                  ]
               }
            ],
            "period":{
               "start":"2023-01-02T01:00:00.000Z",
               "end":"2023-01-02T01:15:00.000Z"
            }
         }
      }
   ]
}
```

### Relevant Calculation Options (e.g. Measurement Period, meta.profile Validation):

measurement start: 01/01/2023, measurement end: 12/01/2023

elsaperelli:
# Summary
Fixes #202 
Even though we deduced that this error was the result of how the cql-to-elm translator works (see cql-to-elm translator issue [here](https://github.com/cqframework/clinical_quality_language/issues/1131)), this PR provides a work around for this issue in fqm-execution (see Matt's explanation on the fqm-execution issue [here](https://github.com/projecttacoma/fqm-execution/issues/202)). 

## New behavior
The cql-to-elm translator causes the ELM Binary Expressions where one side of the comparison is of type Literal to result in the operator (in the case of the issue, `<=`) being grouped in with the literal it is being compared to. Regardless of whether or not this is intentional or a mistake, this PR creates a workaround that makes sure the expression with the literal and comparison operator in it does not use its localId, but rather the localId of the actual comparison expression itself. Doing so makes sure that the highlighting is correct. In order to do this, special handling is added for cql comparison operations. 

## Code changes
- `src/calculation/ClauseResultsHelpers.ts` - Special case handling for when the statement is a ELM Binary Expression with a Comparison Operator in `findAllLocalIdsInStatement` that calls a new function `findLocalIdsForComparisonOperators` that handles this case. 
- `test/unit/ClauseResultsHelper.test.ts` - Two new unit tests to test the new special case handling.
- `test/unit/fixtures/elm/ComparisonWithLiteral.json` / `test/unit/fixtures/elm/ComparisonWithoutLiteral.json` - Two new fixtures that are used in the new unit tests. Are these okay to include? If not, is there a better way to function test this function?

# Testing guidance
1. `npm run check`
2. Download the below zip file and unzip it in the main directory of `fqm-execution`. It includes a `highlighting-issue` folder, `age-highlighting.ts` file, and a `run.sh` script. The `highlighting-issue` folder contains two patient bundles, `patient-bundle-fail.json` has a patient who is younger than 66 and `patient-bundle-pass.json` has a patient who is older than 66, `literal-on-left.json` elm json file where the CQL comparison expression has the literal on the left side, `literal-on-right.json` elm json file where the CQL comparison expression has the literal on the right side (the example in the issue), `highlightIssue.cql` file that contains minimum cql to replicate the issue, and `FHIRHelpers.cql`. I included the last two files incase anyone wanted to test another situation and bundle it using ecqm-bundler.
3. `./run.sh` This runs the detailed calculation on four different measure/patient bundle combinations and puts the resulting logic highlighting html in a folder called html. 
4. `cd html`
5. Look at the four resulting html files. They should look as follows:

`right-fail.html`
image

`left-fail.html`
image

These both make sense because the patient in `patient-bundle-fail.json` is younger than 66, so we want to make sure that the comparison is highlighted in red. This is the solution that was agreed upon in the issue thread.

`right-pass.html` & `left-pass.html`
image

This makes sense because the patient in `patient-bundle-pass.json` is older than 66 so it should be highlighted green.

6. If you want to use the cli instead of the script, you can run the following and then navigate to the debug html folder to see the html: `npm run cli -- detailed -m highlight-issue/literal-on-right.json -p highlight-issue/patient-bundle-fail.json -o --debug -s 2023-01-01 -e 2023-12-31`

[age-highlighting-testing.zip](https://github.com/projecttacoma/fqm-execution/files/10914320/age-highlighting-testing.zip)


sarahmcdougall:
I'm not an official reviewer on this, but I wanted to leave a comment to give kudos for such a well-written PR description. The testing guidance is very detailed, and linking all the related Github issues will make it easy for us to use this as a reference for why we implemented this workaround. Good work!

mgramigna (commented):
I tested the bejesus out of this:

image

Looks like it works perfectly for pretty much every case. The only one it doesn't is when using a `!=`:

image

It looks like the translator generates this as a nested

```
Not
  Equal
```

As opposed to using a `NotEqual` expression, which is the case handled by the code. We might need to determine if it's worth adding a special case for the above structure to make sure it works in all cases

src/calculation/ClauseResultsHelpers.ts
```diff
@@ -194,6 +207,29 @@ export function findAllLocalIdsInSort(
   })();
 }
 
+/**
+ * Checks if an operand in the ELM Binary Expression is of type Literal and if so,
+ * pushes its localId and the overall statement localId to emptyResultClauses
+ * @param statement
+ * @param libraryName
+ * @param emptyResultClauses
```
mgramigna, line 215: I'd either fill these out or delete them

src/calculation/ClauseResultsHelpers.ts
```diff
@@ -194,6 +207,29 @@ export function findAllLocalIdsInSort(
   })();
 }
 
+/**
+ * Checks if an operand in the ELM Binary Expression is of type Literal and if so,
+ * pushes its localId and the overall statement localId to emptyResultClauses
```
hossenlopp, line 212: Might want to add to the comment what we are trying to accomplish. Perhaps something like.
> Due to a quirk of the translator as of version 2.??, the comparison annotation includes the Literal and is marked as the localId of the Literal making it look like the comparison always passes. This function adds a mapping to say we should take the result of the comparison clause as the result of the Literal clause as well. This will make the highlighting of the comparison AND literal follow the result of the comparison.

test/unit/fixtures/elm/ComparisonWithLiteral.json
```diff
@@ -0,0 +1,521 @@
+{
```
hossenlopp, line 1: Where did the CQL for these tests come from? It might be nice to have them incase we want/need to re-translate.

hossenlopp (changes requested):
Small suggestions on adding some documentation and retaining the CQL for re-translation in the future.

src/calculation/ClauseResultsHelpers.ts
elsaperelli: Lol true whoops

test/unit/fixtures/elm/ComparisonWithLiteral.json
elsaperelli: I just made them myself. I still have them, would it be beneficial to commit them in the repository or just hang on to them?

mgramigna (approved):
Good stuff!

hossenlopp (approved):
Looks good and we got the CQL if we want to rebuild in the future.

PR #210, opened 2023-03-13
Log full error in CLI

mgramigna:
Examples using new error reporting from https://github.com/cqframework/cql-execution/pull/308:

Before:

image

After:

image

No need to limit ourselves to just logging the message and the stack in the CLI


hossenlopp (approved):
LGTM. Shows all the info now. 👍

PR #212, opened 2023-03-27
Proportion episode-based measure highlighting fix

### LINKED ISSUE #211: Proportion episode-based measure with falsy IPP shows no highlighting
### Summary

Consider the following example

```cql
define "Qualifying Encounter":
  ["Encounter": "Encounter Inpatient"] InpatientEncounter
    where InpatientEncounter.status = 'finished'

define "ipop":
  "Qualifying Encounter"
```

When a patient does not have a qualifying encounter (in this case, the example has a valid Encounter from that valueset, but the status is `in-progress` instead of `finished`), no highlighting shows up

image

### Expected Behavior

Highlighting should show up and look like this:

image

### Version or Commit

Verified in `v1.0.6` and `v1.0.7`

[relevance-highlighting-input.tar.gz](https://github.com/projecttacoma/fqm-execution/files/11055945/relevance-highlighting-input.tar.gz)

### Relevant Calculation Options (e.g. Measurement Period, meta.profile Validation)

N/A

2.4.0

### Any Additional Info

The engine might not be treating that statement as "relevant" even though it should be? I also tested it without using an expression ref and just putting the query directly in the "ipop" expression. Same behavior

elsaperelli:
# Summary
Fixes #211 

## New behavior
Before, `buildPopulationRelevanceForAllEpisodes()` was being called in scenarios where `episodeResults` is defined, but is an empty array. This caused the highlighting to be incorrect because `buildPopulationRelevanceForAllEpisodes` had no episode array to work with and therefore returned a `masterRelevanceResults` array where every entry had a result of false. Now, in `ClauseResultsBuilder.ts`, we check if `results.episodeResults` is defined AND has a length greater than 0. As a result, the correct highlighting is shown because it is treated as a normal patient based measure when calculating relevance. We set the precedent of using patient-level logic for this in `DetailedResultsBuilder.ts` (line 43) so I thought this fix was okay.

## Code changes
- `ClauseResultsBuilder.ts` - bug fix and some small grammar fixes

# Testing guidance
- In the issue, Matt attached a patient and measure bundle for this issue. 
- Add valuesets to the measure bundle.
- `npm cli -- detailed -m <path-to-measure-bundle> -b <path-to-patient-bundle> --debug`
- `cd debug/html`
- Open the clauses HTML and make sure the highlighting is what it is supposed to be (shown in the issue).
- Change the patient encounter in the patient bundle to have a status of `"finished"` rather than `"in-progress"`.
- Open the detailed results command again and make sure that highlighting appears, but that it is all green.

### DIFF
--- a/src/calculation/ClauseResultsBuilder.ts
+++ b/src/calculation/ClauseResultsBuilder.ts
@@ -46,8 +46,8 @@ import {
  * @param {string} mainLibraryId - The identifier of the main library.
  * @param {ELM[]} elmLibraries - All ELM libraries for the measure.
  * @param {fhir4.MeasureGroup} populationGroup - The population group being calculated.
- * @param {boolean} calculateSDEs - Wether or not to treat SDEs as calculed/relevant or not.
- * @returns {StatementResult[]} The StatementResults list for each statement with it its relevance status populated.
+ * @param {boolean} calculateSDEs - Whether or not to treat SDEs as calculated/relevant or not.
+ * @returns {StatementResult[]} The StatementResults list for each statement with its relevance status populated.
  */
 export function buildStatementRelevanceMap(
   measure: fhir4.Measure,
@@ -132,7 +132,7 @@ export function buildStatementRelevanceMap(
 /**
  * Recursive helper function for the buildStatementRelevanceMap function. This marks a statement as relevant (or not
  * relevant but applicable) in the StatementResults list. It recurses and marks dependent statements also relevant
- * unless they have already been marked as 'TRUE' for their relevance statue. This function will never be called on
+ * unless they have already been marked as 'TRUE' for their relevance status. This function will never be called on
  * statements that are 'NA'.
  *
  * @private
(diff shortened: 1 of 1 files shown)


src/calculation/ClauseResultsBuilder.ts
```diff
@@ -132,7 +132,7 @@ export function buildStatementRelevanceMap(
 /**
  * Recursive helper function for the buildStatementRelevanceMap function. This marks a statement as relevant (or not
  * relevant but applicable) in the StatementResults list. It recurses and marks dependent statements also relevant
- * unless they have already been marked as 'TRUE' for their relevance statue. This function will never be called on
+ * unless they have already been marked as 'TRUE' for their relevance status. This function will never be called on
```
mgramigna, line 135: Nah, this one was right as is

mgramigna (approved):

sarahmcdougall (approved):

PR #214, opened 2023-03-31
Upgrade to cql-execution v3.0.0-beta.4

mgramigna:
# Summary
## Code changes
# Testing guidance

We should now have detailed execution error reporting just from this upgrade. Test with a bundle that throws an error and it should log the full annotated error


elsaperelli (approved):
Tested with durationBetween-issue. Lgtm.

PR #215, opened 2023-03-31
DataRequirements flattened relatedArtifact list

hossenlopp:
# Summary
Adds a combined/flattened version of all `relatedArtifact` lists in a measure bundle to the measure data-requirements output.

## New behavior
Combines all `relatedArtifact` objects in a measure bundle's Measure and Library resources into one list, makes the list unique and puts the resulting list on the `relatedArtifact` property of the the data-requirements output.

## Code changes
`src/helpers/DataRequirementHelpers.ts` - New helper function `getFlattenedRelatedArtifacts` that compiles the flat list of `relatedArtifact`s.
`src/calculation/Calculator.ts` - Calls new helper function to get the list assembled and adds it to the dataRequirements output.

# Testing guidance
Run unit tests. Run data-requirements on a measure bundle and spot check that elements in the `relatedArtifacts` attributes on Measure and Library resources are ending up in the `relatedArtifacts` attribute in the resulting data-requirements output.

### DIFF
--- a/src/calculation/Calculator.ts
+++ b/src/calculation/Calculator.ts
@@ -570,7 +570,15 @@ export async function calculateDataRequirements(
     measureBundle,
     measure
   );
-  return DataRequirementHelpers.getDataRequirements(cqls, rootLibIdentifier, elmJSONs, options, effectivePeriod);
+  const dataRequirements = await DataRequirementHelpers.getDataRequirements(
+    cqls,
+    rootLibIdentifier,
+    elmJSONs,
+    options,
+    effectivePeriod
+  );
+  dataRequirements.results.relatedArtifact = DataRequirementHelpers.getFlattenedRelatedArtifacts(measureBundle);
+  return dataRequirements;
 }
 
 /**
--- a/src/calculation/Calculator.ts
+++ b/src/calculation/Calculator.ts
@@ -548,7 +548,12 @@ export async function calculateLibraryDataRequirements(
     options.rootLibRef
   );
 
-  return DataRequirementHelpers.getDataRequirements(cqls, rootLibIdentifier, elmJSONs, options);
+  const dataRequirements = await DataRequirementHelpers.getDataRequirements(cqls, rootLibIdentifier, elmJSONs, options);
+  dataRequirements.results.relatedArtifact = DataRequirementHelpers.getFlattenedRelatedArtifacts(
+    libraryBundle,
+    options.rootLibRef
+  );
+  return dataRequirements;
 }
 
 /**
--- a/src/helpers/DataRequirementHelpers.ts
+++ b/src/helpers/DataRequirementHelpers.ts
@@ -23,8 +23,9 @@ import * as GapsInCareHelpers from '../gaps/GapsReportBuilder';
 import { parseQueryInfo } from '../gaps/QueryFilterParser';
 import * as RetrievesHelper from '../gaps/RetrievesFinder';
 import { uniqBy } from 'lodash';
-import { DateTime, Interval } from 'cql-execution';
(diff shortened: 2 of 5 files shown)


src/helpers/DataRequirementHelpers.ts
```diff
@@ -23,8 +23,9 @@ import * as GapsInCareHelpers from '../gaps/GapsReportBuilder';
 import { parseQueryInfo } from '../gaps/QueryFilterParser';
 import * as RetrievesHelper from '../gaps/RetrievesFinder';
 import { uniqBy } from 'lodash';
-import { DateTime, Interval } from 'cql-execution';
+import { DateTime, Interval, Library } from 'cql-execution';
```
mgramigna, line 26: Looks like `Library` isn't used

```suggestion
import { DateTime, Interval } from 'cql-execution';
```

src/helpers/DataRequirementHelpers.ts
```diff
@@ -445,3 +446,40 @@ function hasMeasurementPeriodInfo(options: CalculationOptions, effectivePeriod?:
...
+ * @param measureBundle The measure bundle to fetch all RelatedArtifacts from
+ * @returns List of flattened related artifacts.
+ */
+export function getFlattenedRelatedArtifacts(measureBundle: fhir4.Bundle): fhir4.RelatedArtifact[] {
+  const measure = MeasureBundleHelpers.extractMeasureFromBundle(measureBundle);
+  // add the measure itself
+  const relatedArtifacts: fhir4.RelatedArtifact[] = [
+    {
+      type: 'depends-on',
+      display: 'Measure',
```
mgramigna, line 462: Maybe here we can try to use the Measure's `name` in the display if it is defined? It should be defined in most cases

```suggestion
      display: `Measure ${measure.name ?? ''}`,
```

mgramigna (changes requested):
Looks good! Just two small inline comments

src/calculation/Calculator.ts
```diff
@@ -564,7 +564,15 @@ export async function calculateDataRequirements(
...
   const effectivePeriod = MeasureBundleHelpers.extractMeasureFromBundle(measureBundle).effectivePeriod;
-  return DataRequirementHelpers.getDataRequirements(cqls, rootLibIdentifier, elmJSONs, options, effectivePeriod);
+  const dataRequirements = await DataRequirementHelpers.getDataRequirements(
+    cqls,
+    rootLibIdentifier,
+    elmJSONs,
+    options,
+    effectivePeriod
+  );
+  dataRequirements.results.relatedArtifact = DataRequirementHelpers.getFlattenedRelatedArtifacts(measureBundle);
```
mgramigna, line 574: I just realized that it might be useful to also include this on `libraryDataRequirements` (different API function).

In that case, there won't be a Measure resource (the input is assumed to be a bundle of libraries), so it might require a slight re-work. Could probably just have the helper function take in a boolean for `shouldAddMeasure` and only do the logic for extracting it and adding it to the bundle when you're doing regular data requirements?

mgramigna (changes requested):

sarahmcdougall (changes requested):
Great work on this! I just have one in-line comment. I agree with Matt that we should add this functionality for library data requirements. 

I noticed that we re-use some of the same measures and libraries in the new unit tests. As it is, it doesn't seem like too much code duplication, but if we add more unit tests for the library data requirements use case, then we might want to throw some of the measure/library fixtures into `const`s (specifically the Measure without `name` since it gets used a few times, and the main Library since that also gets used a few times).

mgramigna (approved):

sarahmcdougall (approved):
Great work!

PR #216, opened 2023-04-03
DRY extraction of measure for calculation

sarahmcdougall:
# Summary
Code refactors related to measure extraction for the `calculate` function and its helpers.

## New behavior
**Main motivation**: Future work related to composite measure calculation will require looping over component measures and calling various helpers on each component measure. Therefore, we want to limit the amount of code duplication related to measure extraction since we will want to specify the measure outside the helper functions when looping over the components rather than doing a `.find()` for the first measure in the given bundle every time.

No new behavior should be introduced. 

## Code changes
- Update `extractMeasurementPeriod` to take in a measure rather than a bundle
- Update `execute` to take in a measure *in addition to* a bundle
- Update `extractLibrariesFromMeasureBundle` to take in an *optional* measure *in addition to* a bundle
- Various unit test updates to reflect changes to function inputs

Some context behind these changes:

- The change to `extractMeasurementPeriod` is pretty self-explanatory - there's no reason to duplicate the measure extraction within `calculate` and within `extractMeasurementPeriod`
- The change to `execute` comes from the overarching motivation for composite measure calculation. We still need to pass in the entire measure bundle in order to resolve ValueSets. However, we want to specify the measure being passed in rather than extracting it from the bundle once we are in the `execute` function. This will allow us to call `execute` on each component (or on the single measure in the non-composite case).
- The change to `extractLibrariesFromMeasureBundle` is a little more nuanced. We call this function from `retrieveELMInfo`, which itself is called from `execute`. Therefore, when calculating detailed results, we will want to specify the measure that we are extracting the libraries from (following the same idea of looping over components), and this measure is already passed in to `retrieveELMInfo`. However, the `extractLibrariesFromMeasureBundle` function is also used for `calculateDataRequirements` and `calculateQueryInfo`, where we don't really need to specify the measure beforehand (at least not within the scope of this task since we are only focusing on composite measure calculation for detailed results). Therefore, `measure` is an optional input and will be extracted from the passed-in bundle if it is not supplied as a separate input (open to thoughts on this)
- While the `extractMeasureFromBundle` functionality is not necessarily what we will want to use with composite measures (since it does a `.find()` to get the first measure), I kept the existing functionality for now - figured it was out of scope to change this logic without adding full support for composite measures

# Testing guidance

- Run unit tests and basic `detailedResults` CLI commands to check that calculation behavior has not changed
- Inspect `calculate` to see if there are any other useful refactors we can make now to allow for more flexibility with measure calculation

### DIFF
--- a/src/helpers/elm/ELMInfoCache.ts
+++ b/src/helpers/elm/ELMInfoCache.ts
@@ -1,5 +1,10 @@
 import { CodeService, Repository } from 'cql-execution';
-import { MeasureBundleHelpers } from '../..';
+import {
+  isEpisodeOfCareGroup,
+  MeasureWithLibrary,
+  extractLibrariesFromMeasureBundle,
+  codeableConceptToPopulationType
+} from '../MeasureBundleHelpers';
 import {
   generateEpisodeELMJSONFunction,
   generateBooleanELMJSONFunction
@@ -34,7 +39,7 @@ const ELMInfoCache = new Map<string, ELMInfoCacheType>();
  * @returns {Object} an object containing cqls, elmJSONs, rootLibIdentifier, codeService, rep, and vsMap
  */
 export function retrieveELMInfo(
-  measure: MeasureBundleHelpers.MeasureWithLibrary,
+  measure: MeasureWithLibrary,
   measureBundle: fhir4.Bundle,
   valueSets: fhir4.ValueSet[],
   useElmJsonsCaching?: boolean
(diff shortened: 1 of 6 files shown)


src/helpers/elm/ELMInfoCache.ts
```diff
@@ -35,7 +34,7 @@ const ELMInfoCache = new Map<string, ELMInfoCacheType>();
  * @returns {Object} an object containing cqls, elmJSONs, rootLibIdentifier, codeService, rep, and vsMap
  */
 export function retrieveELMInfo(
-  measure: fhir4.Measure,
+  measure: MeasureBundleHelpers.MeasureWithLibrary,
```
elsaperelli, line 37:
```suggestion
  measure: MeasureWithLibrary,
```
And then import it from `MeasureBundleHelpers.ts`? This would be consistent with what's in `Execution.ts`, but it doesn't really matter.

src/helpers/elm/ELMInfoCache.ts
```diff
@@ -58,7 +60,7 @@ export function retrieveELMInfo(
           if (msrPop?.criteria?.expression && obsrvPop.criteria?.expression) {
             const mainLib = elmJSONs.find(elm => elm.library.identifier.id === rootLibIdentifier.id);
             if (mainLib) {
-              const elmFunction = isEpisodeOfCareGroup(measure, group)
+              const elmFunction = MeasureBundleHelpers.isEpisodeOfCareGroup(measure, group)
```
elsaperelli, line 63: Just out of curiosity, why the change? If this is preferred maybe ignore my above comment and then change the code in `Execution.ts` to use this pattern, but up to you :)

sarahmcdougall: I made the change because it looks like we are importing `MeasureBundleHelpers` twice, once to import everything and then to also import `isEpisodeOfCareGroup`. Might change this to import just what we need

elsaperelli (approved):
Lgtm!

mgramigna (approved):

PR #217, opened 2023-04-05
Add regression script for comparing previous calculation results

mgramigna:
# Summary

Adds scripts for comparing calculation results across `ecqm-content` repos. See additions to `contributing.md` for details on what exactly is happening.

## New behavior

None in source code, new script exists.

## Code changes

Le script. Update docs

# Testing guidance

I've simulated an error on the `simulated-regression-error` branch, where for a given calculation, there is a 25% chance that it will incorrectly set numer/denom to true when IPP is false. If you run the script and get no failures at all, go buy a lottery ticket! To test:

1) `git checkout simulated-regression-error`
2) Run the following:

```bash
./regression/run-regression.sh -b add-regression-script
```

This will run regression with this PR's branch as the base, and the error branch as the "one under test." Conceptually this is simulating a scenario where you might be reviewing the code on `simulated-regression-error` and want to make sure that it didn't change any functionality from its base branch (in this case `add-regression-script`). In general the base branch will be master, but right now it won't work against `master` since the regression script isn't checked in there yet.

You should get some errors (this won't be idempotent since errors are random):

* There will be errors reported during comparison that says that the calculation results for two test cases for a measure are different: 
image
* There will be a summary of all of the offending test cases at the end: 
image

This indicates that calculation results differ across the branches

3. To get more info, use `-v` during the regression run which will print the diffs of the actual results (NOTE: the errors you get will be different than step 2 since it's random):

```bash
./regression/run-regression.sh -b add-regression-script -v
```

Diffs will be printed to the terminal in realtime:

image

### DIFF
--- a/regression/regression.ts
+++ b/regression/regression.ts
@@ -39,14 +39,17 @@ async function main() {
     // where the `*-files` directory contains the patient
     const filesPath = path.join(basePath, `${dir.shortName}-files`);
 
-    // Skip measures with no test patients provided
+    // Skip measures with no `*-files` directory
     if (!fs.existsSync(filesPath)) continue;
+
+    const testFilePaths = fs.readdirSync(filesPath).filter(p => p.startsWith('tests-'));
+
+    // Skip measures with no test patients in the `*-files` directory
+    if (testFilePaths.length === 0) continue;
 
     const regressionResultsPath = path.join(REGRESSION_OUTPUT_DIR, dir.shortName);
 
     fs.mkdirSync(regressionResultsPath);
-
-    const testFilePaths = fs.readdirSync(filesPath).filter(p => p.startsWith('tests-'));
 
     // It is assumed that the bundle lives under the base directory with `-bundle.json` added to the extension
     const measureBundle = JSON.parse(
--- a/regression/regression.ts
+++ b/regression/regression.ts
@@ -70,15 +70,17 @@ async function main() {
         fs.writeFileSync(testResultsPath, JSON.stringify(results, undefined, verbose ? 2 : undefined), 'utf8');
         console.log(`${FG_GREEN}%s${RESET}: Results written to ${testResultsPath}`, 'SUCCESS');
       } catch (e) {
-        // Errors will not halt regression. For the purposes of these tests, what matters is that there aren't any new errors that weren't there before
-        // or that the behavior related to the error differs from the base branch to the branch in question.
-        // Errors that occur will be diffed just like normal calculation results
-        fs.writeFileSync(
-          testResultsPath,
-          JSON.stringify({ error: e.message }, undefined, verbose ? 2 : undefined),
-          'utf8'
-        );
-        console.log(`${FG_YELLOW}%s${RESET}: Results written to ${testResultsPath}`, 'EXECUTION ERROR');
+        if (e instanceof Error) {
+          // Errors will not halt regression. For the purposes of these tests, what matters is that there aren't any new errors that weren't there before
+          // or that the behavior related to the error differs from the base branch to the branch in question.
+          // Errors that occur will be diffed just like normal calculation results
+          fs.writeFileSync(
+            testResultsPath,
+            JSON.stringify({ error: e.message }, undefined, verbose ? 2 : undefined),
+            'utf8'
+          );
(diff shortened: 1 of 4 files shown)


regression/regression.ts
```diff
@@ -0,0 +1,84 @@
...
+import path from 'path';
+import { Calculator } from '../src';
+
+const regressionBaseName = process.argv[2] ?? 'regression-output';
+const verbose = process.argv[3] === 'true';
+
+const REGRESSION_OUTPUT_DIR = path.join(__dirname, `./output/${regressionBaseName}`);
+const CTHON_BASE_PATH = path.join(__dirname, './connectathon/fhir401/bundles/measure');
+const ECQM_CONTENT_BASE_PATH = path.join(__dirname, './ecqm-content-r4-2021/bundles/measure');
+const ECQM_CONTENT_QICORE_BASE_PATH = path.join(__dirname, './ecqm-content-qicore-2022/bundles/measure');
```
elsaperelli, line 11: Just out of curiosity, why are we including this? It looks like there is only one measure bundle and no test patients so there is no output.

regression/regression.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  for (const dir of allDirs) {
+    const basePath = dir.fullPath;
+
+    // Regression relies on the `*-files` approach used in ecqm content repositories
+    // where the `*-files` directory contains the patient
+    const filesPath = path.join(basePath, `${dir.shortName}-files`);
+
+    // Skip measures with no test patients provided
+    if (!fs.existsSync(filesPath)) continue;
```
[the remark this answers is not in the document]
elsaperelli: Yeah it appears that it is assumed that if the `*-files` exists, then there are test patients but that is not the case for `ecqm-content-r4-2021/ColorectalCancerScreeningsFHIR`.

regression/regression.ts
```diff
@@ -0,0 +1,84 @@
...
+      try {
+        const { results } = await Calculator.calculate(measureBundle, [patientBundle], {
+          verboseCalculationResults: false
+        });
+
+        fs.writeFileSync(testResultsPath, JSON.stringify(results, undefined, verbose ? 2 : undefined), 'utf8');
+        console.log(`${FG_GREEN}%s${RESET}: Results written to ${testResultsPath}`, 'SUCCESS');
+      } catch (e) {
+        // Errors will not halt regression. For the purposes of these tests, what matters is that there aren't any new errors that weren't there before
+        // or that the behavior related to the error differs from the main branch to the branch in question.
```
elsaperelli, line 71:
```suggestion
        // or that the behavior related to the error differs from the base branch to the branch in question.
```

elsaperelli (changes requested):
This will be a great addition to fqm-execution!! It'll make testing easier and definitely catch some bugs before they happen. Just a few inline comments!

regression/regression.ts
mgramigna: Mainly just future proofing. I didn't notice that there weren't any test patients for those, but hopefully there are eventually.

elsaperelli (approved):
Nice!

regression/regression.ts
```diff
@@ -0,0 +1,87 @@
...
+
+        fs.writeFileSync(testResultsPath, JSON.stringify(results, undefined, verbose ? 2 : undefined), 'utf8');
+        console.log(`${FG_GREEN}%s${RESET}: Results written to ${testResultsPath}`, 'SUCCESS');
+      } catch (e) {
+        // Errors will not halt regression. For the purposes of these tests, what matters is that there aren't any new errors that weren't there before
+        // or that the behavior related to the error differs from the base branch to the branch in question.
+        // Errors that occur will be diffed just like normal calculation results
+        fs.writeFileSync(
+          testResultsPath,
+          JSON.stringify({ error: e.message }, undefined, verbose ? 2 : undefined),
```
hossenlopp, line 78: With the new TS version this line fails to compile on the `e.message` with: 
```
regression/regression.ts:78:35 - error TS18046: 'e' is of type 'unknown'.
```

hossenlopp (changes requested):
I am unable to get this to run locally. But this does look good and simple.

regression/regression.ts
mgramigna: Ah whoops, fixed.

hossenlopp (approved):
Sweet!

PR #218, opened 2023-04-07
Logic highlighting documentation

elsaperelli:
# Summary
This PR adds more detailed descriptions of the measure logic highlighting to our documentation, specifically, the motivation behind unstyled highlighting.

## New behavior
- No new behavior.

## Code changes
- `README.md` - documentation additions
- `logic-highlighting-example-2.png` - screenshot of the relevant logic highlighting behavior.

# Testing guidance
- Check grammar
- Make sure everything makes sense and looks nice!


README.md
```diff
@@ -702,6 +702,14 @@ const { results } = await Calculator.calculateMeasureReports(measureBundle, pati
 
 **NOTE**: Measure logic highlighting is not available when calculating a population `MeasureReport` with `reportType = 'summary'`
 
+However, highlighted HTML provided by `fqm-execution` is not solely based on whether individual pieces of the measure logic CQL had "truthy" or "falsy" values. When there are populations that depend on each other, a relevance hierarchy is built to ensure that populations that are subsets of others aren't considered for membership until membership is determined for their superset population. If a statement's superset statement is false, then no highlighting will appear, regardless of whether that statement's value is "truthy" or "falsy". This pattern reflects the actual population results that are being returned during calculation.
+
+The following example shows what logic highlighting would look like for a proportion boolean measure where the Initial Population evaluates to false but the Denominator and Numerator statements both evaluated to true.
```
natjoe4, line 707: I find this line a tiny bit confusing. I wonder if it would be simpler to exclude "but the Denominator and Numerator statements both evaluated to true."? I think that might better explain that the value of these clauses is irrelevant, but def open to conversation here.

natjoe4 (changes requested):
Nice work. Definitely sheds light on some potentially confusing behavior. Just one small comment.

README.md
```diff
@@ -702,6 +702,14 @@ const { results } = await Calculator.calculateMeasureReports(measureBundle, pati
 
 **NOTE**: Measure logic highlighting is not available when calculating a population `MeasureReport` with `reportType = 'summary'`
 
+However, highlighted HTML provided by `fqm-execution` is not solely based on whether individual pieces of the measure logic CQL had "truthy" or "falsy" values. When there are populations that depend on each other, a relevance hierarchy is built to ensure that populations that are subsets of others aren't considered for membership until membership is determined for their superset population. If a statement's superset statement is false, then no highlighting will appear, regardless of whether that statement's value is "truthy" or "falsy". This pattern reflects the actual population results that are being returned during calculation.
```
mgramigna, line 705: Might want to consider separating this part out by a sub-heading or something, it's a lot of text to appear just under the previous explanation of measure highlighting. Maybe `### Population Relevance for Highlighting` or something. You don't have to include that section in the table of contents or anything, but just to break up the content a bit

elsaperelli: Good call, I always have trouble coming up with good sub-heading names so thank you!

README.md
elsaperelli: Definitely open to changing the wording on this, but my reason for adding that the Denominator and Numerator statements in this case evaluate to true was because that is the case where the confusion initially came from. The statement below the screenshot gets into the fact that the value of those clauses are irrelevant, so I could pull that up here. @mgramigna thoughts?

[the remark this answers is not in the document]
elsaperelli: Not entirely pleased with where I decided to add this so let me know if you think there's a better spot

mgramigna: Yeah maybe pulling up that statement before the screenshot? Could say something like "the following example shows how the logic highlighting will not style the numerator and denominator despite their truthy raw values bc ... " or something idk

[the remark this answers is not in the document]
mgramigna: Yeah, I think "This process is similar to proportion measure scoring for proportion measures depicted in [this flowchart](https://build.fhir.org/ig/HL7/cqf-measures/measure-conformance.html#proportion-measure-scoring)." is a little awkward. 

I would maybe just put it after your last statement since it only applies to the proportion measure example. And can just mention that the relevance calculation matches the population semantics for proportion measures defined in the flowchart

PR #219, opened 2023-04-11
Typescript upgrade

elsaperelli:
# Summary
This PR upgrades the typescript  version to the latest version, [5.0.4](https://www.npmjs.com/package/typescript?activeTab=readme).

## New behavior
- There should be no new behavior related to calculation.

## Code changes
- As anticipated, the only fixes involved needing to explicitly handle errors in try/catch statements.
- When I upgraded the typescript version, I got a warning that the jest version we were on only supported typescript versions <5.0.0, so I decided to upgrade our `jest` and `ts-jest` versions as well. If for whatever reason we want to maintain the old jest version, then I would have to downgrade the typescript version to `4.9.5`.
- I also upgraded `typescript-eslint` dependencies.
- I got a warning that `ts-jest` in globals was deprecated, so I changed `jest.config.js` to reflect the current `ts-jest` [installation docs](https://kulshekhar.github.io/ts-jest/docs/getting-started/installation/#jest-config-file).

Error handling changes:
- `ValueSetResolver.ts` - This approach is consistent to what we do in `Execution.ts`. The alternative is to do: `catch (e: any)` or `catch (e: unknown)`.
- `MeasureBundleHelpers.test.ts` / `parseQueryInfo.test.ts` - I thought this was a better approach than `catch (e: any)`, but let me know if that doesn't make sense.

# Testing guidance
- `npm run check` Make sure there are no new warnings or errors and that all tests pass.
- Make sure the changes I made to the try/catch errors make sense.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,4 +1,5 @@
 module.exports = {
+  preset: 'ts-jest',
   collectCoverageFrom: ['src/**/*.ts'],
   coveragePathIgnorePatterns: [
     'src/integration-tests',
@@ -7,11 +8,6 @@ module.exports = {
     'src/helpers/DebugHelpers.ts',
     'src/execution/Execution.ts'
   ],
-  globals: {
-    'ts-jest': {
-      tsconfig: 'tsconfig.json'
-    }
-  },
   moduleFileExtensions: ['ts', 'js', 'd.ts'],
   transform: {
     '^.+\\.(ts|tsx)$': 'ts-jest'
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,12 +37,12 @@
         "ecqm-bundler": "^0.2.2",
         "eslint": "^7.5.0",
         "eslint-config-prettier": "^6.11.0",
-        "jest": "^26.1.0",
+        "jest": "^29.3.1",
         "opener": "^1.5.2",
         "prettier": "^2.0.5",
-        "ts-jest": "^26.1.3",
+        "ts-jest": "^29.0.3",
         "ts-node": "^9.1.1",
-        "typescript": "^4.9.5"
+        "typescript": "^5.0.4"
       }
     },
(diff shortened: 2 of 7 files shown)


mgramigna (approved):
When I initially wrote this task, TypeScript 5.0 wasn't released yet. Cool that it is now. Normally I'd be really hesitant about skipping an entire major version (4.X), but after running a few tests and reading the [4.0 release notes](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/) and [5.0 release notes](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#const-type-parameters), we don't seem to be impacted by too many of the changes. Some of the [optimizations](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#speed-memory-and-package-size-optimizations) that we get from 5.0 are nice too; I saw a small but not insignificant speed improvement when running the regression tests:

Screen Shot 2023-04-11 at 1 54 02 PM

Screen Shot 2023-04-11 at 1 53 55 PM

Lastly, the diffs between the built JavaScript with the new TypeScript version are very minimal, with small syntax changes when assigning object properties. Overall LGTM.

mgramigna:
While we're here, can we finally just comment out this unused variable (don't remove it) so that lint passes without warnings? https://github.com/projecttacoma/fqm-execution/pull/219/files#file-src-calculation-clauseresultsbuilder-ts-L354

hossenlopp:
> While we're here, can we finally just comment out this unused variable (don't remove it) so that lint passes without warnings? https://github.com/projecttacoma/fqm-execution/pull/219/files#file-src-calculation-clauseresultsbuilder-ts-L354

Yes, OIDs were used in QDM land to reference codesystems. But now we use sane canonical URLs in FHIR. Definitely can be removed.

hossenlopp (approved):
LGTM.

PR #220, opened 2023-04-11
Readme update

elsaperelli:
# Summary
Correct use of HL7® FHIR® standard registered trademark.

## New behavior
None

## Code changes
- Change first instance of FHIR to HL7® FHIR® standard. Subsequent uses should be fine as is. Also added a footnote.

# Testing guidance
- Look at the trademark section in the spec [here](https://build.fhir.org/license.html#trademark) and make sure I didn't miss anything. Also check the slack thread.
- `npm run docs` should show what the docs page looks like. Make sure the link works there!

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
   <h1 align="center">fqm-execution</h1>
 <div align="center">
 
-Library for calculating HL7® FHIR® standard<sup id="fn-1">[\[1\]](#fnref-1)</sup>-based Electronic Clinical Quality Measures (eCQMs) written in Clinical Quality Language (CQL)
+Library for calculating Electronic Clinical Quality Measures (eCQMs) written in Clinical Quality Language (CQL) using the HL7® FHIR® standard<sup id="fn-1">[\[1\]](#fnref-1)</sup>
 
 </div>
 </p>
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 <p align="center">
   <h1 align="center">fqm-execution</h1>
-<center>
+<div align="center">
 
 Library for calculating HL7® FHIR® standard<sup id="fn-1">[\[1\]](#fnref-1)</sup>-based Electronic Clinical Quality Measures (eCQMs) written in Clinical Quality Language (CQL)
 
-</center>
+</div>
 </p>
 <div align="center">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,9 +1,10 @@
 <p align="center">
   <h1 align="center">fqm-execution</h1>
-  <p align="center">
-    Library for calculating FHIR-based Electronic Clinical Quality Measures (eCQMs) written in Clinical Quality Language (CQL)
-    <br/>
-  </p>
+<div align="center">
+
+Library for calculating HL7® FHIR® standard<sup id="fn-1">[\[1\]](#fnref-1)</sup>-based Electronic Clinical Quality Measures (eCQMs) written in Clinical Quality Language (CQL)
```
lmd59, line 5:
```suggestion
Library for calculating Electronic Clinical Quality Measures (eCQMs) written in Clinical Quality Language (CQL) using the HL7® FHIR® standard<sup id="fn-1">[\[1\]](#fnref-1)</sup>
```
The "-based" is kind of awkward now. What do you think of changing it around like this?

lmd59 (commented):
Because it should be trademarked at least once per page, I think this should also be done in the contributing.md doc, which I can do in my PR.

I also noticed that the basic type (i.e. boolean) links in the Calculation Options section appear to be broken. They're just empty links, and I'm not sure why they're linked at all. I would probably just remove the links for basic types. Is this in scope to fix in this PR, or do you think it should be a different one?

elsaperelli:
> Because it should be trademarked at least once per page, I think this should also be done in the contributing.md doc, which I can do in my PR.
> 
> I also noticed that the basic type (i.e. boolean) links in the Calculation Options section appear to be broken. They're just empty links, and I'm not sure why they're linked at all. I would probably just remove the links for basic types. Is this in scope to fix in this PR, or do you think it should be a different one?

Ooo good idea on adding to to the `contributing.md` doc- thanks for doing that! Regarding the basic type links, good catch! I can fix that in this PR :)

README.md
elsaperelli: It seems like "-based" is what is used in almost all of our other PRs, but we will have to update those to use the trademark too eventually, so I can happy to make this change.

lmd59 (approved):

PR #225, opened 2023-04-19
Sde properties

lmd59:
# Summary
Updates SDEResult interface to make `name` optional (not really a useful piece of information, but we're keeping it around for backwards compatibility) and add `id`, `criteriaExpression`, and `usage` as result properties (`usage` required). Updates `getSDEValues` function accordingly to populate these properties.
## New behavior
Output from calculation with SDEs includes these new properties
## Code changes
- `ClauseResultsBuilder.ts` includes updates to `getSDEValues` to populate properties
- `Calculator.ts` types file includes updates to the `SDEResult` interface
-  ClauseResultsBuilder.test.ts` includes added `getSDEValues` unit test
-  `simple-measure.json` fixture includes added `supplementalData` `id`
- Updates `MeasureReportBuilder.addSDE.test.ts` and `MeasureReportBuilder.test.ts` to include new properties as part of test data objects
# Testing guidance
- `npm run check`
- Testing with measure example input: `detailed -m "ecqm-content-r4-2021/bundles/measure/DiabetesHemoglobinA1cHbA1cPoorControl9FHIR/DiabetesHemoglobinA1cHbA1cPoorControl9FHIR-bundle.json" -p "ecqm-content-r4-2021/bundles/measure/DiabetesHemoglobinA1cHbA1cPoorControl9FHIR/DiabetesHemoglobinA1cHbA1cPoorControl9FHIR-files/tests-numer-CMS122-bundle.json" --debug -o` -> Search `output.json` file for `"supplementalData":` to find the updated supplemental data, which should include `id`, `criteriaExpression`, and `usage`

### DIFF
--- a/src/calculation/ClauseResultsBuilder.ts
+++ b/src/calculation/ClauseResultsBuilder.ts
@@ -14,7 +14,8 @@ import {
   PopulationResult,
   SDEResult,
   StatementResult,
-  SupplementalDataUsage
+  SupplementalDataUsage,
+  isSupplementalDataUsage
 } from '../types/Calculator';
 import { UnexpectedProperty } from '../types/errors/CustomErrors';
 
@@ -883,8 +884,7 @@ export function getSDEValues(measure: fhir4.Measure, statementResults: cql.State
             sde.usage.length > 0 &&
             sde.usage[0].coding &&
             sde.usage[0].coding.length > 0 &&
-            sde.usage[0].coding[0].code &&
-            ['supplemental-data', 'risk-adjustment-factor'].includes(sde.usage[0].coding[0].code)
+            isSupplementalDataUsage(sde.usage[0].coding[0].code)
           )
         ) {
           throw new UnexpectedProperty(
--- a/src/calculation/ClauseResultsBuilder.ts
+++ b/src/calculation/ClauseResultsBuilder.ts
@@ -878,20 +878,12 @@ export function getSDEValues(measure: fhir4.Measure, statementResults: cql.State
       if (sde.criteria?.expression) {
         const expression = sde.criteria.expression;
         const result = statementResults[expression];
-        if (
-          !(
-            sde.usage &&
-            sde.usage.length > 0 &&
-            sde.usage[0].coding &&
-            sde.usage[0].coding.length > 0 &&
-            isSupplementalDataUsage(sde.usage[0].coding[0].code)
-          )
-        ) {
+        const usage = sde.usage?.[0]?.coding?.[0].code;
+        if (!isSupplementalDataUsage(usage)) {
           throw new UnexpectedProperty(
             'Expected sde usage code from the MeasureDataUsage valueset: https://terminology.hl7.org/3.1.0/ValueSet-measure-data-usage.html'
           );
         }
-        const usage = sde.usage[0].coding[0].code;
         results.push({
           name: sde.code?.text || expression,
           rawResult: result,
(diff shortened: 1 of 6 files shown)


src/calculation/ClauseResultsBuilder.ts
```diff
@@ -875,10 +878,27 @@ export function getSDEValues(measure: fhir4.Measure, statementResults: cql.State
       if (sde.criteria?.expression) {
         const expression = sde.criteria.expression;
         const result = statementResults[expression];
+        if (
+          !(
+            sde.usage &&
+            sde.usage.length > 0 &&
+            sde.usage[0].coding &&
+            sde.usage[0].coding.length > 0 &&
+            isSupplementalDataUsage(sde.usage[0].coding[0].code)
```
natjoe4, line 887: If we define usage above this line as `const usage = sde.usage?.[0]?.coding?.[0].code;`. We should just be able to replace this whole string of `&&`s with `!isSupplementalDataUsage(usage)`. Or maybe `usage && !isSupplementalDataUsage(usage)` if we want to allow usage to be optional.

src/types/Calculator.ts
```diff
@@ -97,11 +97,27 @@ export interface ExecutionResult<T extends PopulationGroupResult> {
...
   /** Raw result of SDE clause */
   rawResult?: any;
   /** Pretty result for this SDE. */
   pretty?: string;
+  /** The id of the SDE as defined in the measure. */
+  id?: string;
+  /** Criteria expression for this SDE. */
+  criteriaExpression?: string;
+  /** Measure data usage for this SDE. */
+  usage: SupplementalDataUsage;
```
natjoe4, line 110: Just based on the task description, it looks like this was intended to be optional. Did that change?

src/types/Calculator.ts
```diff
@@ -97,11 +97,27 @@ export interface ExecutionResult<T extends PopulationGroupResult> {
...
+}
+/**
```
natjoe4, line 112: Nitpicking, but maybe a newline here for readability :)

natjoe4 (changes requested):
This looks great! A couple small improvements. Also, A unit test that expects an error when the usage is an invalid string could be nice :)

src/calculation/ClauseResultsBuilder.ts
lmd59: Oh man- that is so much better. Love it.

src/types/Calculator.ts
lmd59: yah- double checked with Matt on it this morning!

src/types/Calculator.ts
lmd59: I guess we don't really do that anywhere else. I'll just trust people to shift their eyes one more line if needed lol

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -875,10 +878,19 @@ export function getSDEValues(measure: fhir4.Measure, statementResults: cql.State
...
+          );
+        }
         results.push({
           name: sde.code?.text || expression,
           rawResult: result,
-          pretty: prettyResult(result)
+          pretty: prettyResult(result),
+          id: sde.id,
+          criteriaExpression: expression,
+          usage: usage as SupplementalDataUsage
```
mgramigna, line 893: Because you threw an error above if the type guard is not satisfied, TypeScript knows that from then on, `usage` must be of type `SupplementalDataUsage`, making this cast unnecessary now

```suggestion
          usage
```

natjoe4 (changes requested):

hossenlopp (approved):

natjoe4 (approved):
Great work!

PR #226, opened 2023-04-19
Place reasonDetail on reasonCode, not coding

sarahmcdougall:
# Summary
For care gaps, updates the `GuidanceResponse.reasonCode` to have the `reasonDetail` extension on the `reasonCode` itself rather than on the `coding`.

## New behavior
When running care gaps, the structure of the generated `GuidanceResponse` resources will be different if the reason contains a reference. Previously, the output would look like the following:

```json
"reasonCode": [
  {
    "coding": [
      {
        "system": ... ,
        "code": ... ,
        "display": ...
        "extension": [
          {
            "url": "http://hl7.org/fhir/us/davinci-deqm/StructureDefinition/reasonDetail",
            "extension": [<reason detail extension >]
          }
        ]
      }
    ]
  }
]
```

The expected output should now have the following format:

```json
"reasonCode": [
  {
    "coding": [
      {
        "system": ...,
        "code": ...,
        "display": ...
      }
    ],
    "extension": [
      {
        "url": "http://hl7.org/fhir/us/davinci-deqm/StructureDefinition/reasonDetail",
        "extension": [<reason detail extension>]
      }
    ]
  }
],
```

See the `GuidanceResponse` attachment in this [Jira issue](https://jira.hl7.org/browse/FHIR-40803) to get an understanding of the updated structure.
NOTE: it appears that there are several ways of attaching the extension (might be possible to attach it to the `reasonCode.coding` as well as including it on the root of `reasonCode`). For now, we will use this approach, but the extension may change in the future.

See the [resource profile for Detailed Care Gap Guidance Response](http://hl7.org/fhir/us/davinci-deqm/2023Jan/StructureDefinition-gaps-guidanceresponse-detailedcaregap.html) for more detail.

## Code changes
* `src/gaps/GapsReportBuilder` —> Change the `generateReasonCoding` function to now be named `generateReasonCode`. Now generates the `reasonCode` rather than just the `coding` (which would then be added to the `reasonCode` outside the function). Update `generateGuidanceResponse` function according to this change.
    * Reasoning for this: We still want to map all the reasons to their reason codings, but creating the whole reason code in this function means we won’t need to do an additional mapping to the detailed extension as well. We can just construct the reason code all at once in this function
* `test/unit/GapsInCareHelpers.test.ts` —> Updates unit test fixtures to have the extension defined on the `reasonCode` elements rather than on the `reasonCode`’s `coding`’s. Also updates the function name from `generateReasonCoding` to `generateReasonCode`.

# Testing guidance
* Run unit tests and check that the new unit test fixtures align with the spec
* Run care gaps via the CLI on a measure and patient bundle, and compare the resulting `GuidanceResponse` structure against the output received from the main branch. The only difference should be the placement of the `reasonDetail` extension (if present).
    * Example CLI command: 

```
npm run cli -- gaps -m <EXM 130 measure bundle> -p <EXM 130 denominator patient bundle> -o -s 2019-01-01 -e 2019-12-31
```

### DIFF
--- a/src/gaps/GapsReportBuilder.ts
+++ b/src/gaps/GapsReportBuilder.ts
@@ -215,12 +215,12 @@ export function generateGuidanceResponses(
       ];
     }
 
-    let gapCoding: fhir4.Coding[];
+    let gapStatus: fhir4.CodeableConcept[];
 
     if (q.reasonDetail?.hasReasonDetail && q.reasonDetail.reasons.length > 0) {
-      gapCoding = q.reasonDetail.reasons.map(generateReasonCoding);
+      gapStatus = q.reasonDetail.reasons.map(generateReasonCode);
     } else {
-      gapCoding =
+      const gapCoding: fhir4.Coding[] =
         improvementNotation === ImprovementNotation.POSITIVE
           ? [
               {
@@ -236,11 +236,12 @@ export function generateGuidanceResponses(
                 display: CareGapReasonCodeDisplay[CareGapReasonCode.PRESENT]
               }
             ];
+      gapStatus = [
+        {
+          coding: gapCoding
+        }
+      ];
     }
-
-    const gapStatus: fhir4.CodeableConcept = {
-      coding: gapCoding
-    };
 
     const dataRequirement: fhir4.DataRequirement = {
       type: q.dataType,
@@ -253,7 +254,7 @@ export function generateGuidanceResponses(
       resourceType: 'GuidanceResponse',
       id: uuidv4(),
       dataRequirement: [dataRequirement],
-      reasonCode: [gapStatus],
+      reasonCode: gapStatus,
       status: 'data-required',
       moduleUri: measureURL
     };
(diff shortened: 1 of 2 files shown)


p9g:
Can you add a test with 2 items in the reasonCode array, one with just a coding, and the other with a coding and the extension (with reference and path), for example 

```
 "reasonCode": [
    {
      "coding": [
        {
          "code": "Present",
          "system": "http://hl7.org/fhir/us/davinci-deqm/CodeSystem/care-gap-reason",
          "display": "Data Element is Present"
        }
      ]
    },
    {
      "coding": [
        {
          "code": "ValueInRange",
          "system": "http://hl7.org/fhir/us/davinci-deqm/CodeSystem/care-gap-reason",
          "display": "Value is within specified range"
        }
      ],
      "extension": [
        {
          "extension": [
            {
              "url": "reference",
              "valueReference": {
                "reference": "Patient/cfsb1681933662110"
              }
            },
            {
              "url": "path",
              "valueString": "generalPractitioner"
            }
          ],
          "url": "http://hl7.org/fhir/us/davinci-deqm/StructureDefinition/reasonDetail"
        }
      ]
    }
  ]
```

lmd59 (approved):
It's a little bit unclear the direction we're going with this in the future, but at least according to our current intentions, this all aligns with the ticket and expected behavior we're looking for!
Good job!

sarahmcdougall:
> Can you add a test with 2 items in the reasonCode array, one with just a coding, and the other with a coding and the extension (with reference and path), for example

Good suggestion, I added a test in https://github.com/projecttacoma/fqm-execution/pull/226/commits/5e087d6322be13b3f331d22437005aec43742612 to address this scenario

lmd59 (approved):
Re-review looks good!

elsaperelli (approved):
Nice job with this!

PR #229, opened 2023-04-24
Add Common Issues description of milli-presi issue to contributing do…

lmd59:
…cument

# Summary
Added documentation on time precision to the Contributing docs Common Issues section. 
## New behavior
None
## Code changes
None
# Testing guidance
Look at new section here: https://github.com/projecttacoma/fqm-execution/blob/milli-presi/contributing.md#population-results-with-equivalent-datetime-comparisons
Are there any issues or clarifications that would be helpful?

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -104,15 +104,15 @@ If everything from steps 1-3 seems correct, there may be an issue with the popul
 
 ### Population Results with Equivalent DateTime Comparisons
 
-A commonly seen unexpected population result may come up for when two compared `DateTimes` are thought to be equivalent. When considering `DateTime` equivalencies, it is important to note the precision of the compared `DateTime`s.
+A commonly seen unexpected population result may come up for when two compared `DateTimes` are thought to be equivalent. When considering `DateTime` equivalencies, it is important to note the precision of the compared `DateTime`'s.
 
 Consider the following date data:
 - a: "2021-12-01T08:00:00.000+00:00"
 - b: "2021-12-01T08:00:00+00:00"
 
-While these describe the same time as relevant to most human processes, the calculation engine must consider the uncertainty that comes from less precision in comparing these times. For the `b` `DateTime`, `cql-execution` constructs an uncertainty containing the range of 0 to 999 milliseconds for a given `DateTime`. 
+While these describe the "same time" according to human inspection, the calculation engine must consider the uncertainty that comes from the less precise `DateTime` `b`. For the `b` `DateTime`, `cql-execution` constructs an uncertainty containing the range of 0 to 999 milliseconds. 
 
-When evaluating a statement about `b`,`cql-execution` needs to account for all possible `b` values. This will be represented as an `Uncertainty`. `b` may be anywhere in the range "2021-12-01T08:00:00+00.000:00" to "2021-12-01T08:00:00.999+00:00". As such, if we're comparing `a` and `b`, the engine can guarantee that `a` is <= `b` but not that `b` is <= `a`. If `b` must be before or equivalent to `a`, the result of the comparison clause will be null and show falsy values in highlighting.
+When evaluating a statement about `b`,`cql-execution` needs to account for all possible `b` values. This will be represented as an `Uncertainty`. `b` may be anywhere in the range "2021-12-01T08:00:00.000+00:00" to "2021-12-01T08:00:00.999+00:00". As such, if we're comparing `a` and `b`, the engine can guarantee that `a` is <= `b` but not guarantee that `b` is <= `a`. If `b` must be before or equivalent to `a`, the result of the comparison clause will be `null` and show falsy values in highlighting.
 
 To fix: consider changing your test data to have matching precision, millisecond precision preferred.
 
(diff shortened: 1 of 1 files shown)


contributing.md
```diff
@@ -102,6 +102,23 @@ In some cases, `fqm-execution` will succeed in its calculation, but the data ret
...
+
+Consider the following date data:
+- a: "2021-12-01T08:00:00.000+00:00"
+- b: "2021-12-01T08:00:00+00:00"
+
+While these describe the "same time" according to human inspection, the calculation engine must consider the uncertainty that comes from the less precise `DateTime` `b`. For the `b` `DateTime`, `cql-execution` constructs an uncertainty containing the range of 0 to 999 milliseconds. 
+
+When evaluating a statement about `b`,`cql-execution` needs to account for all possible `b` values. This will be represented as an `Uncertainty`. `b` may be anywhere in the range "2021-12-01T08:00:00.000+00:00" to "2021-12-01T08:00:00.999+00:00". As such, if we're comparing `a` and `b`, the engine can guarantee that `a` is <= `b` but not guarantee that `b` is <= `a`. If `b` must be before or equivalent to `a`, the result of the comparison clause will be `null` and show falsy values in highlighting.
+
+To fix: consider changing your test data to have matching precision, millisecond precision preferred.
```
hossenlopp, line 117:
```suggestion
To fix: Consider changing your test data to have matching precision, millisecond precision preferred. Alternatively, the precision you wish to use for the comparison can be added to the logic. ex. `during seconds of`. 
```
Suggesting mentioning they can define precision in the operation as an alternate.

hossenlopp (changes requested):
Small suggestion with alternate fix.

contributing.md
```diff
@@ -102,6 +102,23 @@ In some cases, `fqm-execution` will succeed in its calculation, but the data ret
 
 If everything from steps 1-3 seems correct, there may be an issue with the population processing logic in `fqm-execution`. In this case, please feel free to submit an issue or a pull request.
 
+### Population Results with Equivalent DateTime Comparisons
+
+A commonly seen unexpected population result may come up for when two compared `DateTimes` are thought to be equivalent. When considering `DateTime` equivalencies, it is important to note the precision of the compared `DateTime`'s.
```
natjoe4, line 107:
```suggestion
A commonly seen unexpected population result may come up for when two compared `DateTime` properties are thought to be equivalent. When considering `DateTime` equivalencies, it is important to note the precision of the compared `DateTime` properties.
```

natjoe4: Just a thought to unify the grammar here. Open to suggestions

contributing.md
```diff
@@ -102,6 +102,23 @@ In some cases, `fqm-execution` will succeed in its calculation, but the data ret
...
+
+Consider the following date data:
```
natjoe4, line 109:
```suggestion
Consider the following `DateTime` data:
```

contributing.md
```diff
@@ -102,6 +102,23 @@ In some cases, `fqm-execution` will succeed in its calculation, but the data ret
...
+- a: "2021-12-01T08:00:00.000+00:00"
+- b: "2021-12-01T08:00:00+00:00"
+
+While these describe the "same time" according to human inspection, the calculation engine must consider the uncertainty that comes from the less precise `DateTime` `b`. For the `b` `DateTime`, `cql-execution` constructs an uncertainty containing the range of 0 to 999 milliseconds.
```
natjoe4, line 113:
```suggestion
While these describe the "same time" according to human inspection, the calculation engine must consider the uncertainty that comes from the less precise `DateTime` `b`. For the `b` `DateTime`, `cql-execution` constructs an `Uncertainty` containing the range of 0 to 999 milliseconds. 
```

natjoe4 (changes requested):
Just some small nitpicks here. Might it be valuable to add a warning to CQL execution that notifies the user when an `Uncertainty` is used in this way?

lmd59:
> Just some small nitpicks here. Might it be valuable to add a warning to CQL execution that notifies the user when an `Uncertainty` is used in this way?

We could... I can add a placeholder ticket for leadership consideration at least.

contributing.md
lmd59: Cool alternative!

contributing.md
```diff
@@ -102,6 +102,23 @@ In some cases, `fqm-execution` will succeed in its calculation, but the data ret
 
 If everything from steps 1-3 seems correct, there may be an issue with the population processing logic in `fqm-execution`. In this case, please feel free to submit an issue or a pull request.
 
+### Population Results with Equivalent DateTime Comparisons
+
+A commonly seen unexpected population result may come up for when two compared `DateTime` properties are thought to be equivalent. When considering `DateTime` equivalencies, it is important to note the precision of the compared `DateTime` properties.
```
elsaperelli, line 107:
```suggestion
A commonly seen unexpected population result may come up when two compared `DateTime` properties are thought to be equivalent. When considering `DateTime` equivalencies, it is important to note the precision of the compared `DateTime` properties.
```
I actually don't know but I think this sounds better? Grammar is not my specialty 😅

elsaperelli (commented):
Great description! Just one potential grammar fix but good to approve after

contributing.md
lmd59: I think the former is technically correct English, but this is definitely better English :-P

elsaperelli (approved):

hossenlopp (approved):

PR #230, opened 2023-04-25
Valuesets Fix

natjoe4:
[bundles.zip](https://github.com/projecttacoma/fqm-execution/files/11324588/bundles.zip)
# Summary
Fqm-execution was only pulling in valuesets listed in Library dataRequirements codeFilters, but needs all valueset relatedArtifacts as well to properly supply results. This PR makees updates to ValueSetHelper.ts to retrieve valuesets from VSAC included in the relatedArtifacts of a Library that are not included in the MeasureBundle

## New behavior
Calculation on MeasureBundles that do not include all necessary valuesets will now pull in valuesets listed in the `relatedArtifacts` property of included Libraries

## Code changes
Added additional checks to `getMissingDependentValuesets` that look for valuesets listed in `relatedArtifacts`

# Testing guidance
- `npm run check`
- run `npm run cli -- -p "issuePB.json" -m "issueMB.json" -o --vs-api-key {VSAC_API_KEY}` on the attached bundles in master and see that the patient incorrectly evaluates outside of the `IPP`
- run the same on my branch and see the patient calculate into the numerator
- run `npm run cli -- valueSets -m "issueMB.json" -o --vs-api-key {VSAC_API_KEY}` and ensure that the output bundle has 26 valueset resources.
- Please reach out if you need a VSAC API key for testing purposes!


src/execution/ValueSetHelper.ts
```diff
@@ -107,33 +107,51 @@ export function getMissingDependentValuesets(
     throw new UnexpectedResource('Expected measure bundle to contain entries');
   }
   const libraryEntries = measureBundle.entry?.filter(
-    e => e.resource?.resourceType === 'Library' && e.resource.dataRequirement
+    e => e.resource?.resourceType === 'Library' && (e.resource.dataRequirement || e.resource.relatedArtifact)
```
lmd59, line 110: Should we worry about any relatedArtifact(s) on the Measure resource as well?

hossenlopp: It actually does not look like ValueSets are referenced in the Measure resource by itself. So just looking at the Library is appropriate.

src/execution/ValueSetHelper.ts
```diff
@@ -107,33 +107,51 @@ export function getMissingDependentValuesets(
...
+        }, [] as string[])
+      );
+    }
+    if (libraryResource.relatedArtifact) {
+      libraryVSURL.push(
+        ...libraryResource.relatedArtifact.reduce((acc: string[], ra) => {
+          if (ra.url && ra.url.includes('ValueSet')) {
+            acc.push(ra.url);
+          }
+          return acc;
```
hossenlopp, line 151:
```suggestion
      libraryVSURL.push(
        ...libraryResource.relatedArtifact.reduce((accumulator: string[], ra) => {
          if (ra.url && ra.url.includes('ValueSet')) {
            accumulator.push(ra.url);
          }
          return acc;
```
To not confuse the `acc` in the parent closure, suggest using a different name like was done with the dataRequirement iteration.

hossenlopp (changes requested):
Works as expected. Just want a small change for more maintainable code.

src/execution/ValueSetHelper.ts
```diff
@@ -92,8 +92,8 @@ export function parseTimeStringAsUTCConvertingToEndOfYear(timeValue: string): Da
 }
 
 /**
- * Collates dependent valuesets from a measure by going through all of the measure bundle's libraries' dataCriteria's codeFilters' valueset.
- * Then finds all valuesets that are not already contained in the measure bundle.
+ * Collates dependent valuesets from a measure by going through all of the valuesets listed in the relatedArtifacts of the measure bundle's measure and libraries,
```
lmd59, line 95:
```suggestion
 * Collates dependent valuesets from a measure by going through all of the valuesets listed in the relatedArtifacts of the measure bundle's libraries,
```

lmd59: Since it seems we don't need to worry about the measure relatedArtifacts

lmd59 (commented):
Tiny update suggested!

src/execution/ValueSetHelper.ts
natjoe4: good catch!

lmd59 (approved):
Fix looks good- thanks for making updates!

hossenlopp (approved):

PR #233, opened 2023-05-03
Expose relevant interfaces from root exports; update docs

mgramigna:
# Summary

Can now export any interface directly from `fqm-execution`. Updated docs, and deprecated the previous type exports to be removed as a breaking change in 2.0.

Before:

```typescript
// These contain every interface/type exported in the corresponding files
import { CalculatorTypes, Enums, CQLTypes, ELMTypes } from 'fqm-execution';
```

After:

```typescript
// Only import specific types that you are using/need to use
import { CalculationOptions, PopulationType, MeasureScoreType, FinalResult, Relevance, /* ... */ } from 'fqm-execution';
```

## New behavior

None in runtime. Any previously export interface can now be imported directly from `fqm-execution` in a TypeScript project

## Code changes

* Add individual type file exports to `src/index.ts` and add deprecation comments/todo
* Update docs

# Testing guidance

`npm link` the project somewhere and try the new import style for the interfaces. Make sure the old style still works too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -819,52 +819,50 @@ const { results } = await Calculator.calculate(measureBundle, [], {
 
 ## Usage with TypeScript
 
-`fqm-execution` exports custom-defined TypeScript interfaces used within the code to allow for easy integration into other TypeScript projects. The TypeScript files defining these interfaces can be found [here](https://github.com/projecttacoma/fqm-execution/tree/master/src/types).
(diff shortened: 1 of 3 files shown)


natjoe4 (approved):
All docs compiled and look good. Linked into fqm-testify and new type defs look great :) Nice work.

elsaperelli (approved):
Nice! This will really clean up some of our code. My only suggestion is to create some backlog tasks for updating our repositories when fqm-execution 2.0 is released so we don't forget.

PR #235, opened 2023-05-10
Add Support for Composite Measure Calculation

mgramigna:
# Summary
The moment we’ve all been waiting for - composite measure support is here! Adds the ability to run measure calculation on a composite measure, and adds the ability to generate a composite measure report.

#### Composite measure calculation
`fqm-execution` can now calculate detailed results for a composite measure. To do so, a measure bundle must be passed in that contains the composite measure, all component measures, and all their dependencies. For each patient, calculates detailed results for each of the components.

#### Composite measure report building
`fqm-execution` can now generate a summary measure report. Running `reports` with a passed-in measure bundle containing a composite will calculate the measure score based on the composite scoring type that is provided. [All-or-nothing](https://build.fhir.org/ig/HL7/cqf-measures/composite-measures.html#all-or-nothing-scoring
), [opportunity](https://build.fhir.org/ig/HL7/cqf-measures/composite-measures.html#opportunity-scoring
), and [linear](https://build.fhir.org/ig/HL7/cqf-measures/composite-measures.html#patient-level-linear-combination-scoring
) are currently supported.

#### Composite measure calculation
* When it comes to `Calculator.ts`, the best approach for seeing code changes is to go through the file directly rather than looking through the code diff
* Summary of calculation changes are as follows:
    * Check if a composite measure is contained in the bundle - if so, extract all components, otherwise proceed as normal
    * Execution is now performed within a loop over the measure(s)
    * When making results for the patients, in the composite case, we search if execution results already exist for the patient so that we can append to them - each patient’s execution results contain the detailed results for each component
    * In the component case, we append `componentResults` onto the patient execution results - this is a simplified version of the detailed results that we can use for report building that gives us access to the `groupId`, `componentCanonical`, and `populationResults`
    * Group clause coverage and overall clause coverage are only created for non-composite measures

#### Composite measure report building
* New abstract class `AbstractMeasureReportBuilder`. Having this abstract class allows us to cleanly separate the composite functionality from the standard measure report building that we’ve been doing.
* New class `CompositeReportBuilder` that extends the abstract class:
    * creates a single group to store the numerator and denominator counts - these get populated when getting the report. When looping through patient results, the numerator and denominator in `this.compositeFraction` get incremented
    * `addPatientResults` increments the numerator and denominator of `this.compositeFraction` according to the composite scoring type defined on the composite report builder. `all-or-nothing`, `opportunity`, and `linear` are currently supported.

#### New integration tests
* Update `Makefile` to generate an all-or-nothing composite measure that consists of two components —> we restricted integration tests to all-or-nothing because the composite scoring type only affects the measure report content, not detailed results
* Added corresponding integration tests for all-or-nothing composite support

#### New unit tests
* Unit tests for composite report builder class, report builder factory
* New simple composite measure fixture
* Tests for new composite measure bundle helpers

#### Other changes
* README - Updated documentation on composite measure calculation support and how to interpret the calculation results
* Updated `cql-execution` to `3.0.0`
* Updated `ecqm-bundler` to `0.4.1`
* The standard `MeasureReportBuilder` no longer takes in the measure bundle as input - just takes in the measure 
* Support for specifying a patient directory as a CLI option instead of needing to list out the path to each patient bundle

# Testing guidance
1. Run integration tests for all-or-nothing composite
2. Run unit tests for composite report builder class, changes to standard measure report builder, composite measure bundle helpers, etc.
3. Test composite report building across the three different composite scoring types. This can be done by changing the composite scoring code on the measure.


test/unit/helpers/MeasureBundleHelpers.test.ts
```diff
@@ -822,6 +823,241 @@ describe('MeasureBundleHelpers tests', () => {
     });
   });
 
+  describe.only('extractComponentsFromMeasure', () => {
```
elsaperelli, line 826: Is this `.only` purposeful?

elsaperelli (changes requested):
Yay composite measures! Took me awhile to go through all of this but I only have a few comments! I also tested `opportunity` and `linear` scoring types and those looked good too!

README.md
```diff
@@ -619,6 +621,131 @@ When enabled, the overall patient-level results will contain the raw results of
 ];
 ```
 
+## Composite Measures
```
lmd59, line 624: Since we now have support for composite measures, we should probably note somewhere in this section what fqm-execution functions still do *not* support composite measures.

src/calculation/CompositeReportBuilder.ts
```diff
@@ -0,0 +1,209 @@
...
+          });
+        }
+      });
+    } else if (this.compositeScoringType === 'linear') {
+      // https://build.fhir.org/ig/HL7/cqf-measures/composite-measures.html#patient-level-linear-combination-scoring
+
+      // Always increment the denominator for linear scoring when processing a patient result
+      this.compositeFraction.denominator++;
+
+      const [patientNumerCount, patientDenomCount] = results.componentResults?.reduce(
```
elsaperelli, line 153: This is a very big nit pick so feel free to ignore, but I feel like we were maybe a little nervous about users being aware that the composite measure population list is a tuple of [denom, numer]. I know this is separate but maybe to reinforce that patten this could be [patientDenomCount, patientNumerCount]. Thoughts?

src/calculation/CompositeReportBuilder.ts
```diff
@@ -0,0 +1,209 @@
...
+> {
+  report: CompositeMeasureReport;
+  compositeScoringType: CompositeScoreType;
+  compositeFraction: { numerator: number; denominator: number };
+
+  constructor(compositeMeasure: fhir4.Measure, options: CalculationOptions) {
+    super(compositeMeasure, options);
+
+    this.compositeScoringType = getCompositeScoringFromMeasure(compositeMeasure) ?? 'all-or-nothing';
+    this.compositeFraction = { numerator: 0, denominator: 0 };
```
elsaperelli, line 30: Related to the one below although even less important because it is an object rather than a tuple, but thoughts on making this `{ denominator: 0, numerator: 0}` just to reinforce that pattern?

src/calculation/CompositeReportBuilder.ts
```diff
@@ -0,0 +1,209 @@
...
+                ]
+              }
+            }
+          ]
+        }
+      ]
+    };
+  }
+
+  public addPatientResults(results: ExecutionResult<T>) {
```
elsaperelli, line 73:
```suggestion
  public addPatientResults(result: ExecutionResult<T>) {
```

More of a question and it isn't really important for this PR but I was thinking about it for my PR branched off of this branch... but thoughts on changing to be singular? It's being called in a loop of results so in `Calculator.ts` it is actually called as `addPatientResults(result)`. I know this was here before this PR but it actually may make things less confusing in the future but we could also deal with that later...

README.md
```diff
@@ -619,6 +621,131 @@ When enabled, the overall patient-level results will contain the raw results of
...
+
+:warning: Composite measure support is highly experimental, and may change as the [specification](https://build.fhir.org/ig/HL7/cqf-measures/composite-measures.html) evolves :warning:
+
+`fqm-execution` supports the calculation of [composite measures](https://build.fhir.org/ig/HL7/cqf-measures/composite-measures.html) under the following conditions:
+
+- The provided measure bundle **must** contain only one composite `Measure` resource (i.e. a FHIR `Measure` resource with `composite` as the scoring code)
+  - See the [conformance requirements of a composite measure](https://build.fhir.org/ig/HL7/cqf-measures/composite-measures.html#component-quality-measures) and the [CQFMCompositeMeasure profile](https://build.fhir.org/ig/HL7/cqf-measures/StructureDefinition-composite-measure-cqfm.html) for specific guidance regarding a composite `Measure` resource
+- The provided measure bundle **must** contain all component `Measure` resources and their dependent `Library` resources needed for calculation
```
elsaperelli, line 632: If I remove the FHIRHelpers library from the composite measure bundle in the `test/integration` folder, the output is the same as when the FHIRHelpers library was included. Is this expected?

elsaperelli, line 632: I also tested removing one of the component measures from the bundle even though it was still a related artifact in the composite measure. Do we want to throw an error in this case? When I tested, it did not throw an error, the detailedResults just did not included population results for that component.

elsaperelli: Looks like the measure score gets calculated correctly, though.

elsaperelli (changes requested):
I tested some measure bundles that did not follow the conditions listed in the README.md and found some cases maybe an error should have been thrown but was not. Other than that just some other small comments!

src/calculation/CompositeReportBuilder.ts
mgramigna: Ideally users will never see this code! But I'll update it for consistency

src/calculation/CompositeReportBuilder.ts
mgramigna: I'm going to say no to this one since order never matters in an object

README.md
mgramigna: Probably, because in those libraries as written, FHIRHelpers is never actually used in the CQL AFAIK. So no statements will throw an error since that library is never actually needed. It probably shouldn't even be included on that composite bundle in the first place

src/calculation/CompositeReportBuilder.ts
mgramigna: https://github.com/projecttacoma/fqm-execution/pull/235/commits/d3657a34a5367c199725a566d599df445f8d025d

src/calculation/CompositeReportBuilder.ts
mgramigna: https://github.com/projecttacoma/fqm-execution/pull/235/commits/d3657a34a5367c199725a566d599df445f8d025d

README.md
mgramigna: https://github.com/projecttacoma/fqm-execution/pull/235/commits/c34809c7218c1e0cf54d30b17744cdffe7850bb8

elsaperelli (approved):
Looks good to me! I think the added error handling is very helpful...I can't think of any other obvious situations we would want to handle.

src/helpers/MeasureBundleHelpers.ts
hossenlopp, line 1: Down on line 315 in `extractLibrariesFromMeasureBundle` the error message could use a quick update to identify which measure we currently are trying to load to identify which measure we cannot find the main library for.

test/integration/composite-all-or-nothing/composite-all-or-nothing.test.ts
```diff
@@ -0,0 +1,166 @@
...
+  'composite-all-or-nothing/patients/patient-comp1-denom-comp2-numer-bundle.json'
+);
+const PATIENT_COMP1_DENOM_COMP2_DENOM: fhir4.Bundle = getJSONFixture(
+  'composite-all-or-nothing/patients/patient-comp1-denom-comp2-denom-bundle.json'
+);
+
+const COMPONENT_ONE_CANONICAL = 'http://example.com/Measure/measure-AllOrNothingComponentOne|0.0.1';
+const COMPONENT_TWO_CANONICAL = 'http://example.com/Measure/measure-AllOrNothingComponentTwo|0.0.1';
+
+describe('Composite measure all-or-nothing scoring', () => {
```
lmd59, line 22: Do we want another integration test that tests the measure report for all-or-nothing or do we feel report building is sufficiently addressed elsewhere?

mgramigna: When we initially created the integration testing we decided that it would only be for `calculate` for now, not measure reports.

MeasureReport integration testing could be a problem for future us, but out of scope for now heheh

lmd59 (approved):
Awesome work here!

src/calculation/CompositeReportBuilder.ts
```diff
@@ -0,0 +1,209 @@
...
+
+      if (patientDenomCount !== 0) {
+        this.compositeFraction.numerator += (patientNumerCount * 1.0) / patientDenomCount;
+      }
+    } else if (this.compositeScoringType === 'weighted') {
+      // https://build.fhir.org/ig/HL7/cqf-measures/composite-measures.html#weighted-scoring
+      throw new Error('Weighted scoring not implemented for composite measures');
+    }
+
+    if (result.evaluatedResource) {
```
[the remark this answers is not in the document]
mgramigna: https://github.com/projecttacoma/fqm-execution/pull/235/commits/ad5ecba3b5cc3ba620c0cd5b0aa2bd104da41105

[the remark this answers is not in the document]
mgramigna: Looks good!

PR #236, opened 2023-05-17
Composite measure weighted scoring

elsaperelli:
# Summary
This PR adds support for calculating measure reports for composite measures with the `weighted` scoring type. More information regarding the component-level `weighted` scoring type can be found [here](https://build.fhir.org/ig/HL7/cqf-measures/composite-measures.html#weighted-scoring).

## New behavior
Before, if you tried to build a measure report from a composite measure with scoring type `weighted`, you got the following error: `"Weighted scoring not implemented for composite measures"`. Now, you can build a measure report from a composite measure with scoring type `weighted`!

## Code changes
- `CompositeReportBuilder.ts` - new `addResults` function that handles `weighted` scoring as well as `weightedComponents` property that keeps track of each component in the composite measure and its associated weight. Per the spec, if a weight extension does not exist, then the weight is treated as 1.
- `AbstractMeasureReportBuilder.ts` - new `addResults` abstract function.
- `Calculator.ts` - `addResults` is now called so that `weighted` scoring composite measures can be handled.
- `CompositeReportBuilder.test.ts` / `simple-composite-measure-weighted-scoring.json` - added a quick unit test that checks the measure report for a composite measure with `weighted` scoring type in two examples: one where weight is defined through an extension and one where no weight is defined so it is assumed to be 1.
- `MeasureReportBuilder.addSDE.test.ts` / `MeasureReportBuilder.test.ts` - updated other unit tests to handle abstract class changes.

### Code change explanations
- The spec defines four Composite Measure scoring methods: `all-or-nothing`, `opportunity`, `linear`, and `weighted`. The first three are `individual-based` scoring methods that are already implemented in the `CompositeReportBuilder.ts` `addPatientResults` method. I could not add the `weighted` scoring calculation to that method because that method goes through individual patient results, but for `weighted` I need to keep track the results of each component. Therefore, I decided to add a method to the `AbstractMeasureReportBuilder` class `addResults` that goes through all of the patient results. In `MeasureReportBuilder.ts`, this method simply calls `addPatientResults` in a loop, but in `CompositeReportBuilder.ts`, it goes through all of the patient results and for each component increments the numerator or denominator count. 

# Testing guidance
- `npm run check` - make sure unit tests pass and there are no lint errors.
- `npm run test:integration` - make sure integration tests still pass.
- Another way I tested it was by editing the composite measure in `tests/integration/composite-all-or-nothing/composite-all-or-nothing-bundle.json` to have scoring type be `weighted` and also adding [cqfm-weight extension](https://build.fhir.org/ig/HL7/cqf-measures/StructureDefinition-cqfm-weight.html) to the related artifacts. Let me know if you want an example!
- With that new measure I ran the following:
`npm run cli -- reports -m test/integration/composite-all-or-nothing/composite-weighted.json --patients-directory test/integration/composite-all-or-nothing/patients -s 2023-01-01 -e 2023-12-31 --report-type "summary"`

### DIFF
--- a/src/calculation/CompositeReportBuilder.ts
+++ b/src/calculation/CompositeReportBuilder.ts
@@ -36,7 +36,7 @@ export class CompositeReportBuilder<T extends PopulationGroupResult> extends Abs
     compositeMeasure.relatedArtifact?.forEach(ra => {
       if (ra.resource && ra.type === 'composed-of') {
         let weight = 1;
-        ra.extension?.forEach(extension => {
+        ra.extension?.find(extension => {
           if (
             extension.url === 'http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqfm-weight' &&
             extension.valueDecimal
--- a/src/calculation/CompositeReportBuilder.ts
+++ b/src/calculation/CompositeReportBuilder.ts
@@ -90,10 +90,10 @@ export class CompositeReportBuilder<T extends PopulationGroupResult> extends Abs
     };
   }
 
-  public addResults(results: ExecutionResult<T>[]) {
+  public addAllResults(results: ExecutionResult<T>[]) {
     // the weighted composite scoring type is component-based, but the others are individual-based
     if (this.compositeScoringType === 'weighted') {
-      this.addComponentResults(results);
+      this.addWeightedResults(results);
     } else {
       results.forEach(result => {
         this.addPatientResults(result);
@@ -101,7 +101,7 @@ export class CompositeReportBuilder<T extends PopulationGroupResult> extends Abs
     }
   }
 
-  private addComponentResults(results: ExecutionResult<T>[]) {
+  private addWeightedResults(results: ExecutionResult<T>[]) {
     // https://build.fhir.org/ig/HL7/cqf-measures/composite-measures.html#weighted-scoring
     const componentPopulationResults: Record<string, { numerator: number; denominator: number; weight: number }> = {};
 
@@ -139,7 +139,7 @@ export class CompositeReportBuilder<T extends PopulationGroupResult> extends Abs
       this.compositeFraction.numerator += value.weight * (value.numerator / value.denominator);
     });
 
-    this.compositeFraction.denominator = Object.entries(componentPopulationResults).length;
+    this.compositeFraction.denominator = Object.keys(componentPopulationResults).length;
   }
 
   public addPatientResults(result: ExecutionResult<T>) {
--- a/src/calculation/CompositeReportBuilder.ts
+++ b/src/calculation/CompositeReportBuilder.ts
@@ -130,20 +130,7 @@ export class CompositeReportBuilder<T extends PopulationGroupResult> extends Abs
         }
       });
 
-      if (result.evaluatedResource) {
-        result.evaluatedResource.forEach(resource => {
-          const reference: fhir4.Reference = {
-            reference: `${resource.resourceType}/${resource.id}`
-          };
-          if (!this.report.evaluatedResource?.some(r => r.reference === reference.reference)) {
-            if (!this.report.evaluatedResource) {
-              this.report.evaluatedResource = [reference];
-            } else {
-              this.report.evaluatedResource.push(reference);
-            }
-          }
-        });
-      }
+      this.addEvaluatedResources(result);
     });
 
     // Multiply each of the component ratios by the weight associated with the component
--- a/src/calculation/CompositeReportBuilder.ts
+++ b/src/calculation/CompositeReportBuilder.ts
@@ -246,6 +246,11 @@ export class CompositeReportBuilder<T extends PopulationGroupResult> extends Abs
       if (patientDenomCount !== 0) {
         this.compositeFraction.numerator += (patientNumerCount * 1.0) / patientDenomCount;
       }
+    } else if (this.compositeScoringType === 'weighted') {
+      // https://build.fhir.org/ig/HL7/cqf-measures/composite-measures.html#weighted-scoring
+      throw new Error(
+        'addPatientResults cannot be used for weighted scoring since it is a component-based composite measure scoring method, addAllResults should be used instead'
+      );
     }
 
     this.addEvaluatedResources(result);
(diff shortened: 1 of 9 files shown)


src/calculation/CompositeReportBuilder.ts
```diff
@@ -22,13 +22,32 @@ export class CompositeReportBuilder<T extends PopulationGroupResult> extends Abs
...
+          if (
+            extension.url === 'http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqfm-weight' &&
+            extension.valueDecimal
+          ) {
+            weight = extension.valueDecimal;
+          }
+        });
+        this.weightedComponents[ra.resource] = weight;
+      }
+    });
```
sarahmcdougall, line 49: Not an official reviewer, but wanted to share something I noticed when reading through the [CQFM Composite Measure profile ](https://build.fhir.org/ig/HL7/cqf-measures/StructureDefinition-composite-measure-cqfm.html)for my task: looks like the weight extension can be defined on the `relatedArtifact` entries, which we do check for here, but the extension could also be defined on `group.component` of the composite measure. So we might also want to do a check to see if `group` is defined on the composite and if it contains the weight extension as well

sarahmcdougall: Screenshot 2023-05-17 at 2 58 47 PM

sarahmcdougall: Update - out of scope for this PR since we would need to add a check to our component extraction to also extract components from `group`.

Made [TACOSTRAT-1306](https://jira.mitre.org/browse/TACOSTRAT-1306) to address this

src/calculation/CompositeReportBuilder.ts
```diff
@@ -70,6 +89,71 @@ export class CompositeReportBuilder<T extends PopulationGroupResult> extends Abs
...
+    if (this.compositeScoringType === 'weighted') {
+      this.addComponentResults(results);
+    } else {
+      results.forEach(result => {
+        this.addPatientResults(result);
+      });
+    }
+  }
+
+  private addComponentResults(results: ExecutionResult<T>[]) {
```
mgramigna, line 103: maybe we should rename this to addWeightedResults or something since it's specific to that

src/calculation/CompositeReportBuilder.ts
```diff
@@ -70,6 +89,71 @@ export class CompositeReportBuilder<T extends PopulationGroupResult> extends Abs
...
+            }
+          }
+        });
+      }
+    });
+
+    // Multiply each of the component ratios by the weight associated with the component
+    // TODO: this may need to change if an extension is created for non-integer counts on the measure report
+    Object.values(componentPopulationResults).forEach(value => {
+      this.compositeFraction.numerator += value.weight * (value.numerator / value.denominator);
```
mgramigna, line 151: I think we should guard against divide by 0 here. So if `value.denominator` is 0 then we shouldn't increment the numerator, or ensure that it gets incremented by 0

test/unit/fixtures/measure/simple-composite-measure-weighted-scoring.json
```diff
@@ -0,0 +1,55 @@
...
+      }
+    ]
+  },
+  "relatedArtifact": [
+    {
+      "type": "composed-of",
+      "extension": [
+        {
+          "url": "http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqfm-weight",
+          "valueDecimal": "0.2"
```
mgramigna, line 37:
```suggestion
          "valueDecimal": 0.2
```

test/unit/fixtures/measure/simple-composite-measure-weighted-scoring.json
```diff
@@ -0,0 +1,55 @@
...
+      ],
+      "display": "Example Component 1",
+      "resource": "http://example.com/Measure/example-component-one|0.0.1"
+    },
+    {
+      "type": "composed-of",
+      "extension": [
+        {
+          "url": "http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqfm-weight",
+          "valueDecimal": "0.3"
```
mgramigna, line 48:
```suggestion
          "valueDecimal": 0.3
```

mgramigna (changes requested):

src/calculation/CompositeReportBuilder.ts
```diff
@@ -22,13 +22,32 @@ export class CompositeReportBuilder<T extends PopulationGroupResult> extends Abs
...
     this.compositeScoringType = getCompositeScoringFromMeasure(compositeMeasure) ?? 'all-or-nothing';
     this.compositeFraction = { numerator: 0, denominator: 0 };
     this.options = options;
+    this.weightedComponents = {};
+
+    // if the weight is not specified, then we assume the weight is 1
+    compositeMeasure.relatedArtifact?.forEach(ra => {
+      if (ra.resource && ra.type === 'composed-of') {
+        let weight = 1;
+        ra.extension?.find(extension => {
```
mgramigna, line 39: Sorry, I didn't mean just changing the forEach to a find, I meant use `find` to actually find the matching extension and only do the valueDecimal access if you get a result back:

```ts
const weightExtension = ra.extension?.find(extension => ...);

if (weightExtension) // do stuff

**Review by mgramigna — APPROVED** [review]

File: src/calculation/CompositeReportBuilder.ts [hunk 7]
```diff
@@ -173,9 +246,6 @@ export class CompositeReportBuilder<T extends PopulationGroupResult> extends Abs
       if (patientDenomCount !== 0) {
         this.compositeFraction.numerator += (patientNumerCount * 1.0) / patientDenomCount;
       }
-    } else if (this.compositeScoringType === 'weighted') {
```
hossenlopp, line 176: Bring back this error but with updated message to indicate that `addPatientResults` cannot be used for weighted scoring right now and `addAllResults` should be used instead.

src/calculation/MeasureReportBuilder.ts
hossenlopp, line 1: Lower down in this file in the static `buildMeasureReports` function, it should be updated to use the `addAllResults` to allow for individual measure reports for composite weighted measures to be calculated.

hossenlopp (changes requested):
Two changes to make sure we can still do individual measure reports and have descriptive error messaging for not supported.

src/calculation/MeasureReportBuilder.ts
hossenlopp: Ignore this comment, we do not support individual reports on composite measures.

hossenlopp (approved):

PR #242, opened 2023-05-22
Expression order

### LINKED ISSUE #238: Ordered CQL Expressions in Logic Highlighting HTML
**Which fqm-execution feature is this request related to?**
Logic highlighting and html generation

**Describe the solution you'd like**
Re-order the CQL expressions into a more intuitive format. Currently, the CQL expressions are semi-arbitrarily ordered based on the ingestion of libraries as read from the measure bundle. Suggest explicitly ordering the CQL expressions, likely putting population definitions at the top. Function definitions could be at the bottom or in some other consistent logical order.

**Additional context**
Example of current html:
Screen Shot 2023-05-17 at 10 29 44 AM
Population definitions are mixed in with other definitions, seemingly at random.

lmd59:
# Summary

Fixes #238 

Generated HTML sorts statements into population statements, non-functions, and function statements, where non-population statements are also in alphabetical order.

## New behavior
HTML output from calculation has appropriately ordered expressions.

## Code changes
- `Calculator.ts` - updates to pass the measure through to the HTMLBuilder
- `HTMLBuilder.ts` - adds `sortStatements` function which sorts the passed statements by population statements, non-functions, and function statements. Measure information is passed through to identify population statements.
- `DataRequirementsHelper.ts` - just a typo update
- `HTMLBuilder.test.ts` - test expression ordering and pass measure through
- Adds PopRetrieveFunc elm and expected HTML fixtures for testing: these fixtures include a population, simple retrieval, and function

# Testing guidance
- `npm run check`
- Testing with measure example input: `npm run cli -- -p "ecqm-content-r4-2021/bundles/measure/CervicalCancerScreeningFHIR/CervicalCancerScreeningFHIR-files/tests-denom-EXM124-bundle.json" -m "ecqm-content-r4-2021/bundles/measure/CervicalCancerScreeningFHIR/CervicalCancerScreeningFHIR-bundle.json" -o --debug` -> Search `output.json` file for "html" to find the html in situ or go to the `debug/html` folder and view any html there in browser to confirm expression order is correct


src/calculation/HTMLBuilder.ts
```diff
@@ -139,6 +145,40 @@ export function generateHTML(
   return result;
 }
 
+/**
+ * Destroys the input annotations and returns the sorted annotations in a logical order:
+ * First, population expressions, according to the order set in the measure
+ * Second, non-function expressions in alphabetical order
+ * Third, function expressions in alphabetical order
+ */
+export function sortAnnotations(measure: fhir4.Measure, groupId: string, annotations: StatementAnnotation[]) {
```
mgramigna, line 154: I personally think we can simply this code a lot by changing the approach. I think we are doing a lot of unnecessary array manipulation and memory allocation for this problem. My suggestion is to consider sorting the statementResults in the original function, before they are converted into annotations, and avoid any type of "removal" during the re-ordering. This should be achievable with just one `.sort`. Here's some pseudocode:

```
  relevantStatements.sort((s1, s2) => {
    if (one of them is a function) {
        // ensure the one that is the function comes second
    }
    
    if (s1 is a population expression) return -1;
    if (s2 is a population expression) return 1;

    return s1.statementName <= s2.statementName ? -1 : 1;
  });
```

A couple things to note:

1. Each statementResult has a handy `.isFunction` attribute
2. For the detection of populationExpressions, this function does a .findIndex on all the annotations within each measure population. The efficiency could be improved here by using a `Set` and assembling a unique set of strings that is all of the population expressions used in the measure group. Then we can do a `O(1)` lookup within the `sort` that I outlined to check the `if(__ is a population expression)`

I implemented a version of the above pseudocode locally and got the same results as this implementation. (sorry)

lmd59: So one thing I was trying to do was to ensure that the population expressions were in the same order expressed by the measure. Do you think that's valuable, or we just don't care as long as they're clustered at the top?

src/calculation/HTMLBuilder.ts
```diff
@@ -85,25 +87,55 @@ Handlebars.registerHelper('highlightCoverage', (localId, context) => {
...
 });
 
+/**
+ * Sort statements into population, then non-functions, then functions
+ */
+export function sortStatements(measure: fhir4.Measure, groupId: string, statements: StatementResult[]) {
+  const group = measure.group?.find(g => g.id === groupId) || measure.group?.[0];
+  const populationSet = new Set(group?.population?.map(p => p.criteria.expression));
+  function alphaCompare(a: StatementResult, b: StatementResult) {
+    return a.statementName > b.statementName ? 1 : b.statementName > a.statementName ? -1 : 0;
```
mgramigna, line 97:
```suggestion
    return a.statementName <= b.statementName ? -1 : 1;
```

I believe this is the same as what you have written. If they're equivalent, statement `a` comes first, which should be fine.

test/unit/HTMLBuilder.test.ts
```diff
@@ -28,7 +29,90 @@ describe('HTMLBuilder', () => {
...
+      raw: undefined,
+      statementName: 'ipp',
+      libraryName: 'Test',
+      localId: 'test-id-3',
+      final: FinalResult.TRUE
+    }
+  ];
+
   beforeEach(() => {
+    simpleMeasure = getJSONFixture('measure/simple-measure.json') as fhir4.Measure;
```
mgramigna, line 115: does this need to be in a beforeEach? If no properties are ever modified by the tests, we should just be able to load this in once before the `describe` as a const for efficiency

mgramigna (commented):
Looks good! Small comments

test/unit/HTMLBuilder.test.ts
lmd59: Agreed.

lmd59: Agreed!

test/unit/HTMLBuilder.test.ts
```diff
@@ -196,4 +297,12 @@ describe('HTMLBuilder', () => {
     const results = calculateClauseCoverage(statementResults, clauseResults);
     expect(results).toEqual('100');
   });
+
+  test('html generation orders population first, then other, then function', () => {
+    const expectedHTML = getHTMLFixture('PopRetrieveFunc.html').replace(/\s/g, '');
+    const res = generateHTML(singlePopMeasure, [popRetrieveFuncElm], prfStatementResults, prfClauseResults, 'test');
+    expect(res.replace(/\s/g, '')).toEqual(expectedHTML);
```
lmd59, line 304: @mgramigna thoughts on ditching this expect statement and the `PopRetrieveFunc.html` fixture or worth keeping?

mgramigna: yeet it outta there!

lmd59: yeeting...

mgramigna (approved):

hossenlopp (approved):
This looks good. It absolutely breaks in the regression script, as expected. The change in order is detected.

Issue #241, opened 2023-05-22
Calculations are not accurate to "seconds" in Condition.onsetDateTime


RohitKandimalla:
### Summary
When I have a patient whose "Condition.onSetDateTime" is less than the Measurement Period End date, the calculations don't seem to be accurate. 

The test patient should pass denominator exclusion when Condition.onSetDateTime = "2024-12-31T23:59:59.000Z" as that meets the timing criteria 'starts on or before end of measurement period'. The test case does not pass until I change the time to 23:59:00 or earlier on 12/31 - having the time of 23:59:01 fails. Screenshots of CQL logic and test case resource can be found in the attached external ticket.

MADiE is using CalculationOptions to override MP start and end dates. 

Measurement Period start date: 2024-01-01T00:00:00.000+00:00
Measurement Period End date: 2024-12-31T23:59:59.999+00:00

MeasureBundle: 
[measureBundle.txt](https://github.com/projecttacoma/fqm-execution/files/11531450/measureBundle.txt)
[patientBundle.txt](https://github.com/projecttacoma/fqm-execution/files/11531451/patientBundle.txt)

Also, the [documentation](https://projecttacoma.github.io/fqm-execution/#/?id=calculating-an-ecqm:~:text=Promise%3CCalculatorTypes.ValueSetOutput%3E-,Calculation%20Options,-The%20options%20that) says that the measurement period should be in YYYY-MM-DD format. but looks like it does calculations to the last minute but doesn't take seconds into consideration. 

#### External Tracking Ticket
https://oncprojectracking.healthit.gov/support/browse/BONNIEMAT-1481

RohitKandimalla:
Awesome, Thank you @mgramigna

mgramigna:
This issue has been resolved and released in [v1.2.0](https://github.com/projecttacoma/fqm-execution/releases/tag/v1.2.0)

Updated documentation for allowed formats of measurement period input to fqm-execution can be found [here](https://projecttacoma.github.io/fqm-execution/#/?id=calculation-options)

In short, any value adhering to the [FHIR datetime format](https://www.hl7.org/fhir/datatypes.html#dateTime) should now be supported.

PR #251, opened 2023-06-05
Update measurement period precision to millisecond

sarahmcdougall:
# Summary
Fixes https://github.com/projecttacoma/fqm-execution/issues/241
Updates measurement period precision support down to the millisecond.

## New behavior
Fqm-execution now supports precision of measurement period down to the millisecond rather than only down to the minute. This has potential impacts on the calculation results, depending on the measurement period precision being used (i.e. results may be different between the master branch and this branch when using millisecond precision in the `CalculationOptions` measurement period)

## Code changes
* `src/execution/ValueSetHelper.ts` —> update the `moment` format to `moment.defaultFormatUtc` in `parseTimeStringAsUTC` and `parseTimeStringAsUTCConvertingToEndOfYear` (see the [moment documentation](https://momentjs.com/docs/#/displaying/format/) for more info on formatting)
    * `parseTimeStringAsUTCConvertingToEndOfYear` does not get used anywhere in the code base, so open to thoughts on whether to keep it
* `README.md` —> documentation updates for measurement period precision
    * In terms of typing, the `defaultFormatUtc` format seems to align with the values allowed by a [FHIR dateTime](https://build.fhir.org/datatypes.html#dateTime), which I included in the documentation. Open to thoughts on whether to explicitly cite the type as a `dateTime` or not.

# Testing guidance
1. Test with the measure/patient bundle provided in #241 (attached below as a `.zip`) with the measurement period specified in the issue: 
* Measurement period start: `2024-01-01T00:00:00.000+00:00`
* Measurement period end: `2024-12-31T23:59:59.999+00:00`

Example CLI command (assuming the bundles are at the root of fqm-execution): 
```bash
npm run cli -- detailed -m measureBundle.json -p patientBundle.json -o --vs-api-key <your-api-key> --debug -s 2024-01-01T00:00:00.000+00:00 -e 2024-12-31T23:59:59.999+00:00
```

Pay attention to the “Narcolepsy Exclusion” statement. On the master branch, the statement will contain red clauses due to minute precision of the measurement period (`Condition.onsetDateTime` is `2024-12-31T23:59:01:000Z`):

Screenshot 2023-06-05 at 10 26 55 AM

On this branch, the entire statement should be green due to millisecond precision:
Screenshot 2023-06-05 at 10 26 29 AM

2. Test different measurement periods with different levels of precision against the example measure/patient bundle. Some examples include:
* YYYY-MM-DD format: when measurement period end is set to 2025-01-01, the statement should be green; when measurement period end is set to 2024-12-31, the statement should be red
* YYYY format: when measurement period end is set to 2025, the statement should be green; when measurement period end is set to 2024, the statement should be red
* YYYY-MM format: when measurement period end is set to 2025-01, the statement should be green; when measurement period end is set to 2024-12, the statement should be red

Bundles from GH issue:
[precisionBundles.zip](https://github.com/projecttacoma/fqm-execution/files/11653607/precisionBundles.zip)


mgramigna:
Do we actually want `moment.defaultFormat` instead of `moment.defaultFormatUtc`? 
image

`defaultFormatUtc` seems to actually ignore any timezone offsets present in the datetime and just always use `Z`, whereas `defaultFormat` accounts for the timezone offset when spitting out the UTC date:

image

I missed this when initially looking at the `moment` docs

mgramigna (approved):

lmd59 (approved):
Looks good to me!

PR #253, opened 2023-06-07
Update weighted composite measure calculation to divide by sum of all…

lmd59:
… weights

# Summary
Weighted composite measure calculation does a weighted average by setting the final denominator to be the some of all weights. 

Example calculation:

- component A: weight 2, component population score .6 (3/5 patients, 60%)
- component B: weight .5, component population score .8 (4/5 patients, 80%)

composite score = ((weight A)*(score A) + (weight B)*(score B))/(weight A + weight B) = ((2)*(.6)+(.5)*(.8))/(2+.5) = 1.6/2.5 = .64 (i.e. 64%)

## New behavior
Final denominator is the sum of weights, final score is a weighted average. This should not affect component-level linear combination scoring.

## Code changes

- `CompositeReportBuilder.ts` sums weights
- `CompositeReportBuilder.test.ts` has updated denominator and score as expected for the `compositeExecutionResults` and `simpleCompositeMeasureWeightedScoring` used in the test
- Added integration test in folder `test/integration/composite-weighted`

# Testing guidance
- `npm run check`
- `npm run test:integration` 
- Play around with different weights in the composite-weighted integration test:

Initially, the weights are 3 and 1 (resulting in a 62.5% overall score), but you can change around weights and run the following:
- `npm run cli -- reports -m test/integration/composite-weighted/composite-weighted-bundle.json --patients-directory test/integration/composite-weighted/patients -s 2023-01-01 -e 2023-12-31 --report-type "summary"`

Component results for the patients are:

- component 1: 0/1, 1/1 -> overall score 50%
- component 2: 1/1, 1/1 -> overall score 100%

Play with different weights to check for a weighted average between these two scores.


test/integration/composite-weighted/composite-weighted.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    //     expect.objectContaining({
+    //       componentCanonical: COMPONENT_TWO_CANONICAL,
+    //       populationResults: getPopulationResultAssertion({
+    //         [PopulationType.IPP]: true,
+    //         [PopulationType.DENOM]: true,
+    //         [PopulationType.NUMER]: true
+    //       })
+    //     })
+    //   ])
+    // );
```
sarahmcdougall, line 53:
```suggestion
```

test/integration/composite-weighted/composite-weighted.test.ts
```diff
@@ -0,0 +1,55 @@
+import { MeasureReport } from 'fhir/r4';
+import { calculateMeasureReports } from '../../../src/calculation/Calculator';
+import { getJSONFixture } from '../helpers/testHelpers';
+
+const MEASURE_BUNDLE: fhir4.Bundle = getJSONFixture('composite-weighted/composite-weighted-bundle.json');
+const PATIENT_COMP1_NUMER_COMP2_NUMER: fhir4.Bundle = getJSONFixture(
+  'composite-all-or-nothing/patients/patient-comp1-numer-comp2-numer-bundle.json'
```
sarahmcdougall, line 7:
```suggestion
  'composite-weighted/patients/patient-comp1-numer-comp2-numer-bundle.json'
```

test/integration/composite-weighted/composite-weighted.test.ts
```diff
@@ -0,0 +1,55 @@
...
+);
+const PATIENT_COMP1_DENOM_COMP2_NUMER: fhir4.Bundle = getJSONFixture(
+  'composite-all-or-nothing/patients/patient-comp1-denom-comp2-numer-bundle.json'
```
sarahmcdougall, line 10:
```suggestion
  'composite-weighted/patients/patient-comp1-denom-comp2-numer-bundle.json'
```

sarahmcdougall (changes requested):

test/integration/composite-weighted/composite-weighted.test.ts
```diff
@@ -0,0 +1,55 @@
...
+      [PATIENT_COMP1_NUMER_COMP2_NUMER, PATIENT_COMP1_DENOM_COMP2_NUMER],
+      {
+        measurementPeriodStart: '2023-01-01',
+        measurementPeriodEnd: '2023-12-31',
+        reportType: 'summary'
+      }
+    );
+
+    expect(results).toBeDefined();
+    const result = results as MeasureReport;
```
mgramigna, line 26: You could use the [CompositeMeasureReport](https://github.com/projecttacoma/fqm-execution/blob/e532aacf23f6bb1bd3afded4368a475173bb8a87/src/calculation/CompositeReportBuilder.ts#L12) type instead, which enforcese that there only be 1 group with 2 populations, avoiding the need for some question marks

test/integration/composite-weighted/composite-weighted.test.ts
```diff
@@ -0,0 +1,55 @@
+import { MeasureReport } from 'fhir/r4';
```
mgramigna, line 1: This is our first integration test with measure reports instead of detailed results, but I think it's really useful especially for these composites. We should think about a strategy to potentially include some measure report and detailed results testing across the integration tests.

test/integration/composite-weighted/cql/weighted-component-two.cql
```diff
@@ -0,0 +1,18 @@
+library AllOrNothingComponentTwo version '0.0.1'
```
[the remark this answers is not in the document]
mgramigna: Yes, generating all of these bundles should be repeatable

test/integration/composite-weighted/composite-weighted.test.ts
lmd59: Is this a good end of sprint activity (happy to update the list), or do we want to tackle it more formally?

mgramigna: Good end of sprint thing I think, just adding/expanding integration tests here in general

test/integration/composite-weighted/composite-weighted.test.ts
```diff
@@ -0,0 +1,30 @@
...
+        measurementPeriodStart: '2023-01-01',
+        measurementPeriodEnd: '2023-12-31',
+        reportType: 'summary'
+      }
+    );
+
+    expect(results).toBeDefined();
+    const result = results as CompositeMeasureReport;
+    expect(result.group[0].measureScore?.value).toBeDefined();
+    expect(result.group[0].measureScore?.value).toEqual(0.625);
```
sarahmcdougall, line 28: Thoughts on adding a check on the numerator/denominator counts as well? Would be something like the following:

```javascript
expect(result.group[0].population?.find(e => e.code?.coding?.[0].code === 'denominator')?.count).toEqual(4);
expect(result.group[0].population?.find(e => e.code?.coding?.[0].code === 'numerator')?.count).toEqual(2.5);
```

For weighted scoring, I feel like it's rare that the numer/denom counts are different from what's expected but still yield the correct measure score. But I have seen this happen a lot when testing all-or-nothing and opportunity scoring (ex. opportunity should give numer of 3, denom of 9, and measure score of 1/3, but instead it incorrectly outputs numer of 1, denom of 3, and measure score of 1/3. You wouldn't know that the numer/denom counts are incorrect just by checking the measure score). 

Might be good to have these additional checks in place so that when we expand upon our integration testing for other scoring types we can have a consistent approach across all scoring types (i.e. check that measure score is defined, measure score is correct, and numer/denom counts are both correct).

sarahmcdougall (approved):
Did some testing with edge cases and everything gets calculated as expected 👍 Integration test changes look good!

I left one in-line suggestion for integration testing but otherwise approving! feel free to either address the comment or we can table it for when we expand upon integration testing

test/integration/composite-weighted/composite-weighted.test.ts
lmd59: Oh sure- I like this idea. Thanks for the suggestion!

mgramigna: Since it's a composite report, do you even need to do a `.find`? We enforce that the populations is an ordered tuple of [denomPop, numerPop], so you _could_ just access those instead. The test will break if we ever change that, but that's expected.

mgramigna (approved):

PR #255, opened 2023-06-12
Add workaround for nested scopes within Property and FunctionRef

### LINKED ISSUE #254: Aliases don't highlight in some cases
### Summary

See the following example of a 100% coverage test case:

image

### Expected Behavior

`MyEncounter` should be highlighted since it is covered, but the coverage highlighting does not pick it up

### Version or Commit

Latest

[coverage-error.zip](https://github.com/projecttacoma/fqm-execution/files/11710023/coverage-error.zip)

2.7.0

### Any Additional Info

We believe there is a workaround that can be done in `fqm-execution` to map the alias onto the proper `localid`

mgramigna:
# Summary

Fixes #254 

Specifically, fixes the following cases:

1. Aliases sometimes don't highlight on subsequent uses within CQL clauses
2. _Calls_ to fluent functions (or regular functions) sometimes don't highlight the alias when it is part of an argument

These two cases are from the same root issue, where the ELM annotation is using a `localId` that never appears in the actual expression, but is 1 less than the `localId` that actually does appear in the expression. We already have _some_ handling for this in [ClauseResultsHelper](https://github.com/projecttacoma/fqm-execution/blob/e532aacf23f6bb1bd3afded4368a475173bb8a87/src/calculation/ClauseResultsHelpers.ts#L98), but this branch of logic only handles if the `.scope` attribute is present on the statement and is itself a reference to the original alias.

In QI-Core authored measures, the way that alias references show up can be different:

```js
// ELM generated from FHIR-based CQL
{
  "localId": "9",
  "path": "status",
  "scope": "MyEncounter",
  "type": "Property"
}
```

```js
// ELM generated from QICore-based CQL
{
  "localId": "9",
  "path": "value",
  "type": "Property",
  "source": {
    "path": "status",
    "scope": "MyEncounter",
    "type": "Property"
  }
}
```

Note the presence of `.source` which actually contains the scope. Importantly, for both of these examples, the `localId` used in the annotation for this alias is actually `8`:

```js
{
  "r": "9",
  "s": [
    {
      "r": "8", // localId 8 does not appear in the actual ELM.
      "s": [
        {
          "value": [
            "MyEncounter"
          ]
        }
      ]
    },
    {
      "value": [
        "."
      ]
    },
    {
      "r": "9",
      "s": [
        {
          "value": [
            "status"
          ]
        }
      ]
    }
  ]
}
```

This PR adds the same "subtract by one logic" to the recursive function that grabs all the localIds and ensures that ELM of the above format maps the use of that off-by-one `localId` on to the original alias in the CQL when defined.

## New behavior

Fixes highlighting of aliases in the above cases:

#### master

qicore-simple-alias-broken

#### This PR

qicore-simple-alias-fixed

#### master

functionref-alias-broken

#### This PR

functionref-alias-fixed

## Code changes

* Update docs with caveat about visual discrepancies in coverage
* Add workarounds to ClauseResultsHelper
* Add tests and cql/elm test fixtures used to check the workaround cases
* Also ensured that localIds show up as strings after using `parseInt`

# Testing guidance

Attached is the measure bundles used to generate the above screenshots, with the patients. Instructions to run it and replicate are in `DESCRIPTION.md`. In all cases, look at `debug/html/clause-coverage-*.html` to view the highlighting.

If you can think of any other alias highlighting cases that are broken on `master` that you believe aren't covered by this PR, we can work to create a test case for it. These are the only two I was able to find.

[coverage-aliases.tar.gz](https://github.com/projecttacoma/fqm-execution/files/11727004/coverage-aliases.tar.gz)

### DIFF
--- a/src/calculation/ClauseResultsHelpers.ts
+++ b/src/calculation/ClauseResultsHelpers.ts
@@ -192,7 +192,7 @@ export function findAllLocalIdsInStatement(
         emptyResultClauses.push({ lib: libraryName, aliasLocalId: alId, expressionLocalId: aliasMap[alias] });
       }
     } else if (k === 'localId') {
-      // else if they key is localId push the value
+      // else if the key is localId, push the value
       localIds[v] = { localId: v };
     } else if (Array.isArray(v) || typeof v === 'object') {
       // if the value is an array or object, recurse


hossenlopp (approved):

sarahmcdougall (changes requested):
Only one small suggested change from me. The additions to `ClauseResultsHelpers` look good, and the documentation additions read well (they are thorough enough without getting too technical). Also TIL what a fluent function is so that's cool 👍 Looks like there are a few uncovered statements/branches but they aren't directly related to the alias handling so I think the amount of unit testing we have right now should be fine

sarahmcdougall (approved):

PR #256, opened 2023-06-21
Add disableHTMLOrdering Calculation Option

sarahmcdougall:
# Summary
Adds a new calculation option `disableHTMLOrdering` that disables CQL statement sorting in HTML output.

## New behavior
When the `disableHTMLOrdering` Calculation Option is used, the CQL statements in the HTML output do not get sorted based on the type of statement (population statement, non-function, function). See https://github.com/projecttacoma/fqm-execution/pull/242 for more information on the sorting that occurs when this calculation option is absent. By default, `disableHTMLOrdering` is set to `false`.

## Code changes
* README - adds calculation option definition, adds section on CQL Statement Ordering
    * **Open to feedback on whether to add more content to the CQL Statement Ordering section - we could maybe make a simple example to show ordering v. no ordering? Not sure if that’s in scope of this task but might be good to add**
* `Calculator.ts` - Passes the disable option to the HTML generation functions for both logic highlighting and clause coverage highlighting
* `HTMLBuilder.ts` - Makes the sorting conditional depending on the value of `disableHTMLOrdering`
* `types/Calculator.ts` - Adds `disableHTMLOrdering` to the `CalculationOptions` interface

# Testing guidance
* Run unit tests
* Test with and without the `disableHTMLOrdering` calculation option and compare results. Here’s one approach for doing this:

1. Run calculation as is with the `CervicalCancerScreeningFHIR` measure.

Example CLI command:
```bash
npm run cli -- -p ecqm-content-r4-2021/bundles/measure/CervicalCancerScreeningFHIR/CervicalCancerScreeningFHIR-files/tests-denom-EXM124-bundle.json -m ecqm-content-r4-2021/bundles/measure/CervicalCancerScreeningFHIR/CervicalCancerScreeningFHIR-bundle.json -o ordered-output.json --debug
```
2. Examine the HTML output for logic highlighting and clause coverage. The HTML output should start with statements “Initial Population,” “Denominator,” “Denominator Exclusions,” etc.
3. Run calculation with HTML ordering disabled. This can be done by changing line 73 to the line `options.disableHTMLOrdering = true;` in`Calculator.ts`. Then, run calculation again with the same command as above, but change the output path to something like `unordered-output.json`.
4. Examine the HTML output for logic highlighting and clause coverage. The HTML output should now start with the functions “Normalize Interval,” “Normalize Abatement,” etc.
5. Compare the clause coverage percentage across the ordered and unordered HTML output - it should be the same. Check that the highlighting is also the same across the two sets of HTML output.
6. Compare the output of both calculations - the calculation output should be the same. An easy way to do this in VSCode is to use the “Select for Compare” feature with the two JSON files, which shows them side by side and highlights any diffs

### DIFF
--- a/README.md
+++ b/README.md
@@ -845,7 +845,7 @@ The following example of a proportion boolean measure shows how the logic highli
 
 ![Screenshot of Highlighting HTML Measure Logic](./static/logic-highlighting-example-2.png)
 
-### CQL Statement Ordering
+### CQL Statement Ordering in HTML
 By default, the statements in the generated HTML output are sorted into population statements, non-functions, and functions, respectively. Non-population statements appear in alphabetical order. To disable this behavior, use the `disableHTMLOrdering` calculation option.
 
 ## Group Clause Coverage Highlighting
--- a/README.md
+++ b/README.md
@@ -846,7 +846,7 @@ The following example of a proportion boolean measure shows how the logic highli
 ![Screenshot of Highlighting HTML Measure Logic](./static/logic-highlighting-example-2.png)
 
 ### CQL Statement Ordering
-By default, the statements in the generated HTML output are sorted into population statements, non-functions, and functions, respectively. To disable this behavior, use the `disableHTMLOrdering` calculation option.
+By default, the statements in the generated HTML output are sorted into population statements, non-functions, and functions, respectively. Non-population statements appear in alphabetical order. To disable this behavior, use the `disableHTMLOrdering` calculation option.
 
 ## Group Clause Coverage Highlighting
 
--- a/README.md
+++ b/README.md
@@ -432,7 +432,7 @@ The options that we support for calculation are as follows:
 - `[calculateHTML]`<[boolean](#calculation-options)>: Include HTML structure for highlighting (default: `true`)
 - `[calculateSDEs]`<[boolean](#calculation-options)>: Include Supplemental Data Elements (SDEs) in calculation (default: `true`)
 - `[clearElmJsonsCache]`<[boolean](#calculation-options)>: If `true`, clears ELM JSON cache before running calculation (default: `false`)
-- `[disabledHTMLOrdering]`<[boolean](#calculation-options)>: Disables custom ordering of CQL statements in HTML output (default: `false`)
+- `[disableHTMLOrdering]`<[boolean](#calculation-options)>: Disables custom ordering of CQL statements in HTML output (default: `false`)
 - `[enableDebugOutput]`<[boolean](#calculation-options)>: Enable debug output including CQL, ELM, results (default: `false`)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -844,6 +845,9 @@ The following example of a proportion boolean measure shows how the logic highli
 
 ![Screenshot of Highlighting HTML Measure Logic](./static/logic-highlighting-example-2.png)
 
+### CQL Statement Ordering
+By default, the statements in the generated HTML output are sorted into population statements, non-functions, and functions, respectively. To disable this behavior, use the `disableHTMLOrdering` calculation option.
```
elsaperelli, line 849: You can decide if you think this is necessary or not, but in the PR for this functionality, Lauren said: "Generated HTML sorts statements into population statements, non-functions, and function statements, where non-population statements are also in alphabetical order". Maybe we want to add the thing about non-population statements being in alphabetical order?

README.md
```diff
@@ -844,6 +845,9 @@ The following example of a proportion boolean measure shows how the logic highli
 
 ![Screenshot of Highlighting HTML Measure Logic](./static/logic-highlighting-example-2.png)
 
+### CQL Statement Ordering
```
elsaperelli, line 848:
```suggestion
### CQL Statement Ordering in HTML
```
Maybe this is unnecessary, but I am find the mixed use of "HTML Ordering" and "CQL Statement Ordering" a bit confusing and maybe this will clear it up a bit without losing any important info?

README.md
elsaperelli, line 849:
```suggestion
By default, the CQL statements in the generated HTML output are sorted into population statements, non-functions, and functions, respectively. To disable this behavior, use the `disableHTMLOrdering` calculation option.
```

src/calculation/HTMLBuilder.ts
```diff
@@ -181,13 +185,15 @@ export function generateHTML(
  * @param elmLibraries main ELM and dependencies to lookup statements
  * @param executionResults array of detailed population group results across
  * all patients
+ * @param disableHTMLOrdering disables statement sorting
```
elsaperelli, line 188:
```suggestion
 * @param disableHTMLOrdering disables CQL statement sorting
```

src/calculation/HTMLBuilder.ts
```diff
@@ -122,17 +122,21 @@ export function sortStatements(measure: fhir4.Measure, groupId: string, statemen
  * @param statementResults StatementResult array from calculation
  * @param clauseResults ClauseResult array from calculation
  * @param groupId ID of population group
+ * @param disableHTMLOrdering disables statement sorting
```
elsaperelli, line 125:
```suggestion
 * @param disableHTMLOrdering disables CQL statement sorting
```

elsaperelli (changes requested):
Nice! Just some small suggestions, but I trust whether you think they are needed or not.

PR #257, opened 2023-06-21
Human readable pretty statement results for FHIR

lmd59:
# Summary
Previously, statement results for FHIR would default to json-like appearances for objects. They now show the most relevant fields for FHIR objects and use mappings to identify more useful information and labels (such as displays for code systems and relevant path properties).

## New behavior
Statement results have improved `pretty` field for results that are objects or codes

## Code changes

- Adds `parse-code-systems` script to `package.json`, which uses the `parseCodeSystems.ts` script to create the code system URL to pretty code system mapping. This leverages the files in the `code-systems` folder as input and output to go from the `cts-metadata` file to the easily consumable `system-map` file.
- Adds `code-attributes/codePaths.ts` taken from fqm-testify for easy identification of primary code path
- Updates to `ClauseResultBuilder.ts`
  - Pulls in all valuable mappings
  - Updates the prettyResult function to more flexibly show type information, show a prettier version of the code system using the `system-map`, show array elements as singular if there isn't more than one, remove unused QDM logic, and handle FHIR objectlogic
  - Updates FHIR object logic to handle tuples without QDM logic and calls `prettyFHIRObject` for base FHIRObjects
  - Adds `prettyFHIRObject` to work with FHIRObjects that come from the calculation results to identify the resource type and component pieces in a human readable (similar to Bonnie) format
- Updates integration testing to add proportion-Encounter-tuple for tuple testing and makes sure other integration results (as in `ratio-Encounter-reuseObservationFunction`) have pretty results
- Updates unit tests to make sure codes and code systems are better displayed.

# Testing guidance
- `npm run check`
- `npm run test:integration` 

Play around with different calculations with interesting FHIR results. Some examples are included in the integration tests:

- `npm run cli -- -p "test/integration/proportion-Encounter-tuple/patients/patient-2Enc-1day-3day.json" -m "test/integration/proportion-Encounter-tuple/proportion-Encounter-tuple-bundle.json" -o --debug -s 2022-01-01 -e 2022-12-31` creates tuple outputs (look for "Enc with Durations" in the statement results), which you can compare to Bonnie tuple outputs (some differences exist in how intervals are rendered):

Screen Shot 2023-06-21 at 2 40 46 PM

- `npm run cli -- -p "test/integration/ratio-Encounter-reuseObservationFunction/patients/patient-3enc-1in-all-2in-denom.json" -m "test/integration/ratio-Encounter-reuseObservationFunction/ratio-Encounter-reuseObservationFunction-bundle.json" -o --debug -s 2022-01-01 -e 2022-12-31` creates results with multiple FHIRObjects (look for the "denom" statement) or results with a single FHIR object (look for the "numer" statement). This could be compared with a simple FHIR object in Bonnie:

Screen Shot 2023-06-21 at 2 45 54 PM

- You can also play with other measures that have object results for statements. For instance `ecqm-content-r4-2021/bundles/measure/CervicalCancerScreeningFHIR/CervicalCancerScreeningFHIR-bundle.json` may have an encounter result for the statement `Qualifying Encounters` or an observation result for the statement `Cervical Cytology Within 3 Years`. As a note, this measure uses outdated gender administration code system URLs, so the human-readable code system lookup does not work (and the URL is shown instead).


mgramigna:
This fails when running with an old connectathon bundle:

image

In this case, `resourceType` is actually `Coding`. Are we assuming that we always have a FHIR resource by the time we access the `primaryCodePath`? Or is this just identifying an edge-case that we aren't handling?

src/scripts/parseCodeSystems.ts
```diff
@@ -0,0 +1,35 @@
+// Code systems source: https://cts.nlm.nih.gov/fhir/metadata?_format=json
```
mgramigna, line 1: We should prob add this file to `coveragePathIgnorePatterns` in `jest.config.js`

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -18,6 +18,15 @@ import {
 } from '../types/Calculator';
 import { UnexpectedProperty } from '../types/errors/CustomErrors';
 
+// code system, search parameters, and code paths for pulling pretty information from resources
+import _systemMap from '../code-system/system-map.json';
+const systemMap = _systemMap as Record<string, string>;
```
mgramigna, line 23: We import the JSON file directly and then cast it to `Record<string, string>`. Would it be easier to just make it `systemMap.ts` or something and export a `const` that is the proper `Record` type?

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -363,100 +372,110 @@ export function prettyResult(result: any | null, indentLevel?: number, keyIndent
...
   } else if (result instanceof Interval) {
-    return `INTERVAL: ${prettyResult(result.low)} - ${prettyResult(result.high)}`;
+    const typeStr = includeType ? 'INTERVAL: ' : '';
+    return `${typeStr}${prettyResult(result.low)} - ${prettyResult(result.high)}`;
   } else if (result instanceof Code) {
-    // TODO: Sort out a better way to have a friendly system display for FHIR codes
-    return `CODE: ${result.system} ${result.code}`;
+    const typeStr = includeType ? 'CODE: ' : '';
+    if (result.system) {
+      if (systemMap[result.system]) {
```
mgramigna, line 380: Slightly more idiomatic to check if the key exists in the object rather than checking if its value is truthy

```suggestion
      if (result.system in systemMap) {
```

src/code-system/cts-metadata.json
```diff
@@ -0,0 +1,810 @@
+{
```
mgramigna, line 1: The files in this directory are just used in the script, right? If so, we should make sure these JSON files don't get included in the `build` directory when doing `npm run build` which will increase the size of the fqm-execution package despite not being relevant for the source code

src/scripts/parseCodeSystems.ts
```diff
@@ -0,0 +1,35 @@
...
+
+const inputPath = path.resolve(path.join(__dirname, '../code-system/cts-metadata.json'));
+const outputPath = path.resolve(path.join(__dirname, '../code-system/system-map.json'));
+const ctsMetadata = readFileSync(inputPath, 'utf8');
+
+/**
+ * Parse cts-metadata and transform into code system url->name mapping
+ */
+async function parse(codeSystemJSON: string) {
+  const codeSystemData = (await JSON.parse(codeSystemJSON)) as fhir4.CapabilityStatement;
```
mgramigna, line 14: This function is not async https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#return_value

```suggestion
  const codeSystemData = JSON.parse(codeSystemJSON) as fhir4.CapabilityStatement;
```

src/scripts/parseCodeSystems.ts
```diff
@@ -0,0 +1,35 @@
...
+import path from 'path';
+
+const inputPath = path.resolve(path.join(__dirname, '../code-system/cts-metadata.json'));
+const outputPath = path.resolve(path.join(__dirname, '../code-system/system-map.json'));
+const ctsMetadata = readFileSync(inputPath, 'utf8');
+
+/**
+ * Parse cts-metadata and transform into code system url->name mapping
+ */
+async function parse(codeSystemJSON: string) {
```
mgramigna, line 13: Slightly confusing to have this be named `codeStystemJSON` when it is a JSON string. Suggest either doing the JSON.parse outside of this function and change this from a `string` to a `fhir4.CapabilityStatement`, or just rename the variable

mgramigna, line 13: If you get rid of the async JSON.parse call, does this function need to be async at all?

test/integration/proportion-Encounter-tuple/proportion-Encounter-tuple.test.ts
```diff
@@ -0,0 +1,43 @@
+import { calculate } from '../../../src/calculation/Calculator';
```
mgramigna, line 1: I'm a bit confused as to why we're testing the pretty results via integration testing. Is there something specific here that isn't replicatable via hardcoded statement results or something that would work within a unit test instead? This feels more like a use case for a unit test as opposed to creating a whole measure bundle just to test pretty results

test/unit/ClauseResultsBuilder.test.ts
```diff
@@ -349,6 +350,21 @@ describe('ClauseResultsBuilder', () => {
...
+      const code = new Code('testCode', 'http://snomed.info/sct');
+      expect(ClauseResultsBuilder.prettyResult(code)).toEqual('CODE: SNOMEDCT testCode');
+    });
+
+    test('should create pretty code for unmapped system URL', () => {
+      const code = new Code('testCode', 'www.example.com');
+      expect(ClauseResultsBuilder.prettyResult(code)).toEqual('CODE: www.example.com testCode');
+    });
+
+    test('should create pretty code for undefined system URL', () => {
```
mgramigna, line 364: Along with my above comment on integration tests, is there any way we can expand these unit tests more to cover what that integration test is covering instead?

mgramigna (changes requested):
These are my requested changes outside of `ClauseResultsBuilder.ts`. I'm holding off on doing a detailed review of that file until the fixes for the other bundles I talked about are in place

lmd59:
> This fails when running with an old connectathon bundle:
> 
> image
> 
> In this case, `resourceType` is actually `Coding`. Are we assuming that we always have a FHIR resource by the time we access the `primaryCodePath`? Or is this just identifying an edge-case that we aren't handling?
> 
> EDIT: This also happens with the ColorectalCancerScreeningsFHIR bundle in ecqm-content-r4-2021:
> 
> image

I have this fixed, but I'm looking into some other code-like types handling based on some additional discussion with @hossenlopp .

test/unit/ClauseResultsBuilder.test.ts
lmd59: I played around with trying to do unit testing with FHIRObjects a couple of times, but because that class is not actually exported, we run into a lot of trouble, and I think would have to mock out everything to the point of non-usefulness.

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -363,100 +371,149 @@ export function prettyResult(result: any | null, indentLevel?: number, keyIndent
...
   } else if (result instanceof Interval) {
-    return `INTERVAL: ${prettyResult(result.low)} - ${prettyResult(result.high)}`;
+    const typeStr = includeType ? 'INTERVAL: ' : '';
+    return `${typeStr}${prettyResult(result.low)} - ${prettyResult(result.high)}`;
   } else if (result instanceof Code) {
-    // TODO: Sort out a better way to have a friendly system display for FHIR codes
-    return `CODE: ${result.system} ${result.code}`;
+    return prettyCode(result.code, result.system, result.display, includeType);
+  } else if (result instanceof Concept) {
+    return prettyConcept(result.display, result.codes, includeType, indentLevel, keyIndent);
```
elsaperelli, line 379: One thing I noticed that I am not even sure if it's a problem or not is that the pretty result for SDE Sex and SDE Race are different (one is in square brackets and the other is not). I believe they are both Coding types, so should they be the same? I did some digging but I got stuck but it looks like the SDE race statement goes through the `prettyFHIRObject` function where the SDE sex statement only goes through the `prettyCode` function.

Screenshot 2023-06-27 at 11 38 38 AM

elsaperelli (changes requested):
Great stuff! Some small inline comments and some open questions. I tested this on all the ecqm-content-r4-2021 and connectathon measures and found two things I wasn't completely sure about:

1. It looks like medication request looks doesn't have a space between the two words:

image

Not sure if this even matters but I thought I would point it out. Relevant measure is EXM105-8.2.000 & the numerator patient.

2. Sometimes the observation statement pretty result isn't in square brackets. An example of it without square brackets can be seen in the DiabetesHemoglobinA1cHbA1cPoorControl9FHIR measure & denom patient (with measurementPeriodStart: 2019-01-01 and measurementPeriodEnd: 2019-12-31):

Screenshot 2023-06-27 at 12 18 33 PM

src/calculation/ClauseResultsBuilder.ts
mgramigna: I believe these statements return different data types. SDE Race expression (in the base FHIR version for colorectal cancer screening) returns a list of FHIR Codings. SDE Sex returns a singular CQL system Code

package.json
```diff
@@ -51,6 +51,7 @@
     "docs": "docsify serve . ",
     "lint": "tsc && eslint \"**/*.{js,ts}\"",
     "lint:fix": "tsc --noEmit && eslint \"**/*.{js,ts}\" --quiet --fix",
+    "parse-code-systems": "ts-node src/scripts/parseCodeSystems.ts",
```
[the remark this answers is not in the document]
lmd59: I like this! I worry that "systems" is a bit vague, but the script itself is called "parseCodeSystems" so any needed clarification is not too far from hand!

test/unit/ClauseResultsBuilder.test.ts
```diff
@@ -349,6 +350,37 @@ describe('ClauseResultsBuilder', () => {
...
+      const code = new Code('testCode');
+      expect(ClauseResultsBuilder.prettyResult(code)).toEqual('CODE: UNDEFINED_SYSTEM testCode');
+    });
+
+    test('should create pretty concept', () => {
+      const concept = new Concept([new Code('testCode1'), new Code('testCode2')], 'testConcept');
+      const forTest = ClauseResultsBuilder.prettyResult(concept);
+      console.log(forTest);
```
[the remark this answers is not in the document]
lmd59: eep- sorry!

lmd59:
> Great stuff! Some small inline comments and some open questions. I tested this on all the ecqm-content-r4-2021 and connectathon measures and found two things I wasn't completely sure about:
> 
> 1. It looks like medication request looks doesn't have a space between the two words:
> 
> image
> 
> Not sure if this even matters but I thought I would point it out. Relevant measure is EXM105-8.2.000 & the numerator patient.
> 
> 2. Sometimes the observation statement pretty result isn't in square brackets. An example of it without square brackets can be seen in the DiabetesHemoglobinA1cHbA1cPoorControl9FHIR measure & denom patient (with measurementPeriodStart: 2019-01-01 and measurementPeriodEnd: 2019-12-31):
> 
> Screenshot 2023-06-27 at 12 18 33 PM

1. MedicationRequest is the name of the resource type, which I think we receive as camel case - maybe gets a bit confusing when we all-caps it. I could potentially not capitalize the whole type, leaving it camel case. I don't think that would be too different from Bonnie, though not what I initially discussed with @hossenlopp. Would that work better?
2. The square brackets are to indicate that it's an array, so if there are square brackets, that just means the type of the result is an array (even if it just has one item). This depends on the statement being made and what kind or results are expected from that statement calculation. We experimented a bit with getting rid of the brackets when there's just one item, but it might still be valuable for the user to know whether the type of the result is an array or not, so we left it with the brackets.

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -363,100 +372,162 @@ export function prettyResult(result: any | null, indentLevel?: number, keyIndent
...
+      result.get('code')?.['value'],
+      result.get('system')?.['value'],
+      result.get('display')?.['value'],
+      includeType
+    );
+  } else if (resourceType === 'CodeableConcept') {
+    return prettyConcept(result.get('text')?.['value'], result.get('coding'), includeType, indentLevel, keyIndent);
+  }
+
+  // Handle resources
```
mgramigna, line 509: I think there are a lot of data types missing from this handling. Some more realistic than others, but I think the assumption that we are dealing with FHIR resources after this if/else branch is not quite right. What if we have a FHIR Duration, string, Identifier, Reference, code, id, etc.? I tested with a big example and got a couple of errors where the code assumes we are dealing with a FHIR resource type. I'll include the bundle and patient I was testing with here, I think it covers a lot more cases than the integration test

[pretty-kitchen-sink.zip](https://github.com/projecttacoma/fqm-execution/files/11885689/pretty-kitchen-sink.zip)

I don't necessarily think we need to add handling for each one of these, but I would definitely prefer if this could could ensure that `resourceType` is a valid key of `parsedCodePaths` before trying to access it so we avoid runtime errors.

mgramigna (changes requested):

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -363,100 +372,215 @@ export function prettyResult(result: any | null, indentLevel?: number, keyIndent
   if (result instanceof DateTime) {
     return moment.utc(result.toString()).format('MM/DD/YYYY h:mm A');
   } else if (result instanceof Interval) {
-    return `INTERVAL: ${prettyResult(result.low)} - ${prettyResult(result.high)}`;
+    const typeStr = includeType ? 'INTERVAL: ' : '';
```
mgramigna, line 375: Unrelated to this line, but for `DateTime`s we are using `MM/DD/YYYY h:mm A` and chopping off the seconds. Maybe we should just make it `MM/DD/YYYY h:mm:ss A` instead to show all the time values

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -363,100 +372,215 @@ export function prettyResult(result: any | null, indentLevel?: number, keyIndent
...
+  indentLevel?: number,
+  keyIndent?: number
+): string {
+  const resourceType = result.getTypeInfo()._name;
+  const keyIndentation = Array(keyIndent).join(' ');
+  const typeStr = includeType ? `${resourceType.toUpperCase()}: ` : '';
+  if (resourceType === 'Period') {
+    return `${typeStr}${prettyResult(result.get('start'))} - ${prettyResult(result.get('end'))}`;
+  } else if (resourceType === 'dateTime') {
+    return `${moment.utc(result.get('value').toString()).format('MM/DD/YYYY h:mm A')}`;
```
mgramigna, line 513: Same here

```suggestion
    return `${moment.utc(result.get('value').toString()).format('MM/DD/YYYY h:mm:ss A')}`;
```

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -363,100 +372,215 @@ export function prettyResult(result: any | null, indentLevel?: number, keyIndent
...
+  } else if (resourceType === 'date') {
+    return `${moment.utc(result.get('value').toString()).format('MM/DD/YYYY')}`;
+  } else if (resourceType === 'Coding') {
+    return prettyCode(
+      result.get('code')?.['value'],
+      result.get('system')?.['value'],
+      result.get('display')?.['value'],
```
mgramigna, line 520:
```suggestion
      result.get('code')?.value,
      result.get('system')?.value,
      result.get('display')?.value,
```

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -363,100 +372,215 @@ export function prettyResult(result: any | null, indentLevel?: number, keyIndent
...
+      includeType
+    );
+  } else if (resourceType === 'CodeableConcept') {
+    return prettyConcept(result.get('text')?.['value'], result.get('coding'), includeType, indentLevel, keyIndent);
```
mgramigna, line 524:
```suggestion
    return prettyConcept(result.get('text')?.value, result.get('coding'), includeType, indentLevel, keyIndent);
```

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -363,100 +372,215 @@ export function prettyResult(result: any | null, indentLevel?: number, keyIndent
...
+  } else if (resourceType === 'Duration') {
+    return `${typeStr}${result.get('value')?.['value']} ${result.get('unit')?.['value']}`;
```
mgramigna, line 526:
```suggestion
    return `${typeStr}${result.get('value')?.value} ${result.get('unit')?.value}`;
```

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -363,100 +372,215 @@ export function prettyResult(result: any | null, indentLevel?: number, keyIndent
...
+  } else if (resourceType === 'Ratio') {
+    const numer = result.get('numerator')?.['value']?.['value'];
```
mgramigna, line 528:
```suggestion
    const numer = result.get('numerator')?.value?.value;
```

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -363,100 +372,215 @@ export function prettyResult(result: any | null, indentLevel?: number, keyIndent
...
+    const numerUnit = result.get('numerator')?.['unit']?.['value'];
```
mgramigna, line 529:
```suggestion
    const numerUnit = result.get('numerator')?.unit?.value;
```

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -363,100 +372,215 @@ export function prettyResult(result: any | null, indentLevel?: number, keyIndent
...
+    const denom = result.get('denominator')?.['value']?.['value'];
```
mgramigna, line 530:
```suggestion
    const denom = result.get('denominator')?.value?.value;
```

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -363,100 +372,215 @@ export function prettyResult(result: any | null, indentLevel?: number, keyIndent
...
+    const denomUnit = result.get('denominator')?.['unit']?.['value'];
```
mgramigna, line 531:
```suggestion
    const denomUnit = result.get('denominator')?.unit?.value;
```

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -363,100 +372,215 @@ export function prettyResult(result: any | null, indentLevel?: number, keyIndent
...
+  if (result.getTypeInfo().findElement('value', true)) {
+    const value = result.get('value');
+
+    if (typeof value === 'string' || typeof value === 'boolean' || typeof value === 'number') {
+      // Try simple value
+      return `${typeStr}${value}`;
+    } else if (
+      typeof value?.['value'] === 'string' ||
+      typeof value?.['value'] === 'boolean' ||
+      typeof value?.['value'] === 'number'
```
mgramigna, line 573:
```suggestion
      typeof value?.value === 'string' ||
      typeof value?.value === 'boolean' ||
      typeof value?.value === 'number'
```

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -363,100 +372,215 @@ export function prettyResult(result: any | null, indentLevel?: number, keyIndent
...
+    ) {
+      // Try nested value
+      return `${typeStr}${value?.['value']}`;
```
mgramigna, line 576:
```suggestion
      return `${typeStr}${value?.value}`;
```

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -363,100 +372,215 @@ export function prettyResult(result: any | null, indentLevel?: number, keyIndent
...
-  } else if (result instanceof String || typeof result === 'string') {
+  } else if (typeof result === 'string') {
     return `"${result}"`;
-  } else if (result instanceof Array) {
-    prettyResultReturn = result.map((value: any) => prettyResult(value, indentLevel, keyIndent));
+  } else if (Array.isArray(result)) {
+    prettyResultReturn = result.map((value: any) => prettyResult(value, includeType, indentLevel, keyIndent));
     return `[${prettyResultReturn.join(`,\n${currentIndentation}${keyIndentation}`)}]`;
-  } else if (result instanceof Object) {
+  } else if (typeof result === 'object') {
```
mgramigna, line 392: Omg. Today I learned that `typeof null === 'object'` which makes me very sad. Let's add another guard I guess. This throws an error if a test patient has an Encounter or something that is missing the queried for info in the kitchen sink example (like an encounter without a period)

```suggestion
  } else if (result != null && typeof result === 'object') {
```

mgramigna (changes requested):
Small inline changes.

I also think this PR should include some documentation of the pretty results in the readme, explaining how each statement result for each patient has a `.pretty` attribute that is a human-readable version of what data was returned from that statement.

src/calculation/ClauseResultsBuilder.ts
lmd59: oh - I didn't realize this would work since it doesn't work one level up, but this is much better!

src/calculation/ClauseResultsBuilder.ts
```diff
@@ -504,28 +522,63 @@ export function prettyFHIRObject(
...
+
+  // Handle values
+  if (result.getTypeInfo().findElement('value', true)) {
+    const value = result.get('value');
+
+    if (typeof value === 'string' || typeof value === 'boolean' || typeof value === 'number') {
+      // Try simple value
+      return `${typeStr}${value}`;
+    } else if (
+      typeof value?.['value'] === 'string' ||
```
elsaperelli: Not sure if this is a big deal but I thought I would point it out to get thoughts on it...but in Matt's bundle, one of the statement results is of resourceType `EncounterStatus` and its value is a string. Since `EncounterStatus` isn't in `parsedCodePaths`, then the pretty result is `ENCOUNTERSTATUS: final`. Do we want the typeStr to be all uppercase in this scenario?

elsaperelli (changes requested):
Nice! Great additions. Just one inline suggestion. Also, I noticed that `buildStatementAndClauseResults` has `doPretty` as a parameter, even though I think it is always set to true. Do we still want this?

lmd59:
> buildStatementAndClauseResults

Hmm- yeah looks like buildStatementAndClauseResults is only ever called with `doPretty: true`. It certainly isn't hurting anything. Maybe we were thinking of making this configurable at a higher level at some point?

lmd59:
> Nice! Great additions. Just one inline suggestion. Also, I noticed that `buildStatementAndClauseResults` has `doPretty` as a parameter, even though I think it is always set to true. Do we still want this?

Hmm- yeah looks like `buildStatementAndClauseResults()` is only ever called with doPretty: true. It certainly isn't hurting anything. Maybe we were thinking of making this configurable at a higher level at some point?

src/calculation/ClauseResultsBuilder.ts
lmd59: Yeaaaah- I did notice this. I think there's going to be a lot of potential variability in what kind of thing ends up being printed from this statement. I would lean toward keeping this all caps to match the rest of the output... or changing everything to not be all caps. The resource type being Pascal case kind of makes sense as a special case, but I don't think this one really does, IMO.

elsaperelli (approved):

test/integration/pretty-various-types/pretty-various-types.test.ts
```diff
@@ -0,0 +1,115 @@
+import { calculate } from '../../../src/calculation/Calculator';
+import { CalculationOptions, DetailedPopulationGroupResult } from '../../../src/types/Calculator';
+import { getJSONFixture, getGroupByIndex } from '../helpers/testHelpers';
+
+const CALCULATION_OPTIONS: CalculationOptions = {
+  measurementPeriodStart: '2022-01-01',
+  measurementPeriodEnd: '2022-12-31',
+  enableDebugOutput: true
```
mgramigna, line 8: do we need this option?
```suggestion
  enableDebugOutput: true
```

lmd59: I guess not- it can be helpful when you're working with the feature directly, but probably doesn't need to be in the integration tests long term.

lmd59: We have it in a few of integration tests. I'll remove it from all of them!

mgramigna (approved):

PR #260, opened 2023-07-06
Order population expressions for HTML output

elsaperelli:
# Summary
This is similar to Lauren's PR [here](https://github.com/projecttacoma/fqm-execution/pull/242) where we sorted the statements into population statements, non-functions, and then function statements; however, now we also order the population statements in the following order:

1. initial-population
2. denominator
3. denominator-exclusion
4. denominator-exception
5. numerator
6. numerator-exclusion
7. measure-population
8. measure-population-exclusion
9. measure-observation (alphabetically sorted if multiple)

...and then non-functions and function statements.

## New behavior
Population expression statements are now sorted in the above order. This order is based on the following "egg":
image

## Code changes
- `HTMLBuilder.ts` - I just added the sorting logic for the population expressions directly to the `sortStatements` function. Rather than using the `populationSet`, we now use `populationIdentifiers` because it includes the `code` (which is always the same) and the criteria expression (which can be different) for each population. That way we can sort statements because on what code their statement name links to.
- `HTMLBuilder.test.ts` - two new unit tests for `sortStatements()`- one for a proportion boolean measure (checks ordering of ipp, denom, numer) and one for a cv boolean measure (checks for ordering of ipp, measure population, measure population exclusion, measure observation function).

# Testing guidance
- `npm run check`
- I tested this in two ways:
1. I just ran calculation on as many measure bundles as I could find with the `--debug` flag and inspected the html output in the `debug/html` directory.
2. If you want to test certain cases, you can write your own CQL, bundle it with Matt's ecqm-bundler, and then run calculation on that to ensure the ordering is correct.


lmd59 (commented):
Still doing testing, but something in my environment got a bit wonky, so leaving these code comments while I figure that one out!

README.md
```diff
@@ -865,7 +865,24 @@ The following example of a proportion boolean measure shows how the logic highli
...
+4. denominator-exception
+5. numerator
+6. numerator-exclusion
+7. measure-population
+8. measure-population-exclusion
+9. measure-observation (alphabetically sorted if multiple)
+
+The order of these populations is determined by most inclusive to least inclusive populations as shown in the following diagram:
+
+![Screenshot of Population Venn Diagram](./static/population-diagram.png)
```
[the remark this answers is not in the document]
elsaperelli: True! Do you remember where we originally found this lol

src/calculation/HTMLBuilder.ts
```diff
@@ -92,23 +92,49 @@ Handlebars.registerHelper('highlightCoverage', (localId, context) => {
...
+    'denominator-exclusion',
+    'denominator-exception',
+    'numerator',
+    'numerator-exclusion',
+    'measure-population',
+    'measure-population-exclusion',
+    'measure-observation'
+  ];
+
+  const populationIdentifiers: Record<string, string> = {};
```
mgramigna, line 107: Quick comment saying that this is a lookup of cql expression identifier -> population type might be useful

src/calculation/HTMLBuilder.ts
```diff
@@ -92,23 +92,49 @@ Handlebars.registerHelper('highlightCoverage', (localId, context) => {
...
   function alphaCompare(a: StatementResult, b: StatementResult) {
     return a.statementName <= b.statementName ? -1 : 1;
   }
+
   statements.sort((a, b) => {
+    // if population statement, leave order or send to beginning
+    if (Object.keys(populationIdentifiers).includes(a.statementName)) {
+      return Object.keys(populationIdentifiers).includes(b.statementName) ? populationCompare(a, b) : -1;
+    }
+    if (Object.keys(populationIdentifiers).includes(b.statementName)) return 1;
```
mgramigna, line 130: `if(Object.keys(something).includes(k))` can almost always be replaced by `if(k in something)`

mgramigna (changes requested):

src/calculation/HTMLBuilder.ts
```diff
@@ -92,23 +92,50 @@ Handlebars.registerHelper('highlightCoverage', (localId, context) => {
  */
 export function sortStatements(measure: fhir4.Measure, groupId: string, statements: StatementResult[]) {
   const group = measure.group?.find(g => g.id === groupId) || measure.group?.[0];
-  const populationSet = new Set(group?.population?.map(p => p.criteria.expression));
+  const populationOrder = [
```
lmd59, line 95: This is the same order as the enum itself, so I think you can do
```suggestion
  const populationOrder = Object.values(PopulationType);
```
but if this throws off other things, feel free to ignore.

lmd59 (approved):

src/calculation/HTMLBuilder.ts
```diff
@@ -92,23 +92,40 @@ Handlebars.registerHelper('highlightCoverage', (localId, context) => {
  */
 export function sortStatements(measure: fhir4.Measure, groupId: string, statements: StatementResult[]) {
   const group = measure.group?.find(g => g.id === groupId) || measure.group?.[0];
-  const populationSet = new Set(group?.population?.map(p => p.criteria.expression));
+  const populationOrder = Object.values(PopulationType);
```
mgramigna, line 95: hmmmm. Not sure I like relying on the iteration order of `Object.values` for a TypeScript enum. I think we should bring back the explicit ordering in a constant array.

mgramigna (approved):

PR #264, opened 2023-07-17
ts-node upgrade

elsaperelli:
# Summary
Upgrades the `ts-node` version to `10.4.0`.

## New behavior
There should be no new behavior.

## Code changes
- `package.json` / `package-lock.json`

# Testing guidance
- `npm run check`

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "opener": "^1.5.2",
     "prettier": "^2.0.5",
     "ts-jest": "^29.0.3",
-    "ts-node": "^10.4.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -39,7 +39,7 @@
     "opener": "^1.5.2",
     "prettier": "^2.0.5",
     "ts-jest": "^29.0.3",
-    "ts-node": "^9.1.1",
+    "ts-node": "^10.4.0",
```
mgramigna, line 42: latest is 10.9.1, might as well try that and confirm that the behavior is still good?

PR #266, opened 2023-07-25
Update clause coverage percentage calculation to capture all relevant clauses

### LINKED ISSUE #262: Madie - Bonnie: Discrepancy between code highlighting and coverage
### Summary
  The included measure is displaying 100% code coverage, however the highlighting is not covering all union clauses.

#### External Tracking Ticket
  https://oncprojectracking.healthit.gov/support/browse/BONNIEMAT-1511

### Expected Behavior
  Highlighting in both Madie and Bonnie would be handled the same.

[CMS133FHIR-v0.0.007-FHIR4.zip](https://github.com/projecttacoma/fqm-execution/files/12029705/CMS133FHIR-v0.0.007-FHIR4.zip)
image

elsaperelli:
# Summary
Fixes #262 

## New behavior
- Fixes a bug in the clause coverage percentage calculation. 
- Clause coverage percentage calculation now captures all relevant clauses and therefore matches the clause coverage highlighting.

## Code changes
- Before, we were getting `allRelevantClauses` by finding the clauseResults whose `localID` and `libraryName` matched a statement in `relevantStatements`. This does not capture all of the relevant clauses. We want to find the clauseResults whose `statementName` and `libraryName` match statements in `relevantStatements`. This way all clauses are included in the calculation, not just the root statement clauses like it was doing before.
- Additionally, combines some code in `generateClauseCoverageHTML` that didn't need to be separate.

# Testing guidance
- `npm run check`
- `npm run test:integration`
- Make sure nothing breaks in the testing.
- Test with the bundle and test cases provided in the issue. The clause coverage percentage for this should now be 65.6% and not 100%. 
- Test some other cases to make sure that the highlighting matches the percentage. Specifically, check that the percentage changes when additional clauses end up being highlighted.

### DIFF
--- a/src/calculation/HTMLBuilder.ts
+++ b/src/calculation/HTMLBuilder.ts
@@ -240,16 +240,13 @@ export function generateClauseCoverageHTML(
   // go through the lookup object of each of the groups with their total
   // detailedResults and calculate the clause coverage html for each group
   Object.entries(groupResultLookup).forEach(([groupId, detailedResults]) => {
-    const flattenedStatementResults = detailedResults.flatMap(s => s.statementResults);
+    // Grab the statement results from just the first patient since just only need the names
+    // of the statements and whether or not their relevance is NA
+    const flattenedStatementResults = detailedResults[0].statementResults;
     const flattenedClauseResults = detailedResults.flatMap(c => (c.clauseResults ? c.clauseResults : []));
 
-    // Filter out any statement results where the statement relevance is not NA and filter out duplicates from that
-    // uniqWith appears to pick the first element it encounters that matches the uniqueness condition
-    // when iterating, which is fine because the relevance not being N/A is the only thing that matters now
-    const uniqueRelevantStatements = uniqWith(
-      flattenedStatementResults.filter(s => s.relevance !== Relevance.NA),
-      (s1, s2) => s1.libraryName === s2.libraryName && s1.localId === s2.localId
-    );
+    // Filter out any statement results where the statement relevance is NA
+    const uniqueRelevantStatements = flattenedStatementResults.filter(s => s.relevance !== Relevance.NA);
 
     if (!disableHTMLOrdering) {
       sortStatements(measure, groupId, uniqueRelevantStatements);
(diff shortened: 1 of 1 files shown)


sarahmcdougall (approved):
Tested via 

- unit/integration tests
- tested the provided measure bundle and patient bundle(s)
- tested the provided bundles by covering additional clauses and checking the percentage change
- ran calculation against additional bundles and population types
- logged the numerator/denominator of the clause coverage percentage calculation to check that they were representative of clauses and not top-level statements

Only suggestion I have is to make the PR title slightly more descriptive so that we can easily refer back to this PR in the release notes

src/calculation/HTMLBuilder.ts
```diff
@@ -243,19 +243,11 @@ export function generateClauseCoverageHTML(
     const flattenedStatementResults = detailedResults.flatMap(s => s.statementResults);
```
hossenlopp, line 243:
```suggestion
    const flattenedStatementResults = detailedResults[0].statementResults;
```

We could work from the statementResults from just the first patient. Since we just need the names of the statements, and wether or not relevance is `NA`. If things are NA then the statement is not use at all by the measure. This information is on all statementResults for all patients in all groups (unique results to the group of course).

src/calculation/HTMLBuilder.ts
```diff
@@ -243,19 +243,11 @@ export function generateClauseCoverageHTML(
...
-    // Grab every statement with any relevance other than N/A
-    // There may be multiple entries for a given statement across the results,
-    // but we know that non of them can be irrelevant
-    const relevantStatements = uniqWith(
-      flattenedStatementResults,
-      (s1, s2) => s1.libraryName === s2.libraryName && s1.localId === s2.localId && s1.relevance === s2.relevance
-    ).filter(s => s.relevance !== Relevance.NA);
-
-    // From all the relevant ones, filter out any duplicate statements
+    // Filter out any statement results where the statement relevance is not NA and filter out duplicates from that
```
hossenlopp, line 246:
```suggestion
    // Filter out any statement results where the statement relevance is NA and filter out duplicates from that
```

We are filtering out the `NA`s keeping everything else.

hossenlopp (commented):
Suggested potential changes as per discussion slack.

sarahmcdougall (approved):
No changes in results from testing, so looks good to me!

hossenlopp (approved):

Issue #268, opened 2023-07-27
Add statement-level HTML generation


sarahmcdougall:
**Which fqm-execution feature is this request related to?**
HTML statement generation (used for logic highlighting and clause coverage)

**Describe the solution you'd like**
A new boolean calculation option `buildStatementLevelHTML` will be added, that is true by default.

When **true**, the statement results returned from measure calculation will have a new populated field `statementLevelHTML` on each statement result. The statement-level HTML will be accessible via the `statementResults` array that exists on the `detailedResults` that are returned upon successful measure calculation. Note that the statement-level HTML will only exist for statements with relevance not equal to “NA.”

The structure of each `statementResult` will be as follows:

```
{
   "libraryName": "library name",
   "statementName": "statement name",
   ...
   "statementLevelHTML": "statement-level HTML"
}
``` 

When **false**, the statement results returned from measure calculation will not have a populated field for statement-level HTML. If the `calculateHTML` calculation option is set to true, then the `detailedResults` for a single population group will still contain the HTML markup across all the CQL statements for the given measure. In other words, when false, the functionality is the same as the current functionality on the master branch.

For backwards compatibility, the `html` field that currently exists on the `detailedResults` will not be affected, and the debug option will still generate the debug HTML with the entire HTML output.

**Additional context**
This new feature will allow the user to build their HTML alongside the “pretty” formatted statements without having to do so on the entire HTML output. This gives additional flexibility, as the user will not need to parse through the HTML and inject content at each statement result.

sarahmcdougall:
Added in [v1.3.0](https://github.com/projecttacoma/fqm-execution/releases/tag/v1.3.0)

PR #270, opened 2023-08-07
Add statement-level HTML to statement results

sarahmcdougall:
# Summary
Adds statement-level HTML as a field on each statement result returned from measure calculation.

See https://github.com/projecttacoma/fqm-execution/issues/268 for more details on this feature.

## New behavior
The user can now specify a new calculation option: `buildStatementLevelHTML`. When set to `true`, the statement results returned from the detailed results will contain a new element `statementLevelHTML` that maps to a string of the HTML markup for the given statement. Thus, the structure of each `statementResult` will be as follows:

```
{
   "libraryName": "library name",
   "statementName": "statement name",
   ...
   "statementLevelHTML": "statement-level HTML"
}
```

For backwards compatibility, the overall `html` element and corresponding debug output are still available.

## Code changes
* Documentation updates to reflect the new calculation option and updated output
* Calculator type updates to reflect the new calculation option
* HTML Builder changes
    * Removed the `statementAnnotations` array, which stores all the statement annotations and then loops over each one for generating HTML clauses. Instead generates the statement-level HTML when looping over the relevant statements, and appends the statement-level HTML to an overall HTML structure.
* HTML Builder unit tests

# Testing guidance
* Run unit tests and integration tests (`npm run test:plus`)
* Run measure calculation against a measure with the `--debug` flag enabled. After calculation finishes, compare the logic highlighting `html` to the `statementResults` available on the `detailedResults`.
    * For statements that are not highlighted in the debug html (that have relevance equal to `NA`), the `statementLevelHTML` element should *not* be present on the statement’s `statementResults` object.
    * For statements that do not have relevance equal to `NA`, the `statementLevelHTML` element should be present on the statement’s `statementResults` object, and the statement-level HTML should match the HTML markup for the given statement in the debug HTML file. You can search by the `data-statement-name`/`data-library-name` in both the `statementResults` and debug html to compare the markup. The statement-level HTML markup should include the `<pre>` tag with the appropriate styling.

Note - open to thoughts on integration testing for this feature. A backlog task exists for adding more thorough testing for HTML building since the unit tests are very small in scope, but we could incorporate the beginnings of integration testing into this PR if desired. I did some brainstorming but am not 100% sure of a path forward for the integration testing (ex. Do we want to add HTML building tests for each measure type? Do we want to add tests to just the proportion boolean integration test? Do we want to make a whole new directory of integration tests specifically for HTML building?)

### DIFF
--- a/test/unit/HTMLBuilder.test.ts
+++ b/test/unit/HTMLBuilder.test.ts
@@ -176,6 +176,34 @@ describe('HTMLBuilder', () => {
     }
   });
 
+  test('statement-level HTML is not added to statement results when statement relevance is NA', () => {
+    const statementResultsNA = [
+      {
+        statementName: 'statementName',
+        libraryName: 'libraryName',
+        final: FinalResult.NA,
+        relevance: Relevance.NA,
+        localId: 'localId'
+      }
+    ];
+
+    const clauseResultsNA = [
+      {
+        statementName: 'statementName',
+        libraryName: 'libraryName',
+        localId: 'localId',
+        final: FinalResult.NA,
+        raw: true
+      }
+    ];
+
+    generateHTML(simpleMeasure, [elm], statementResultsNA, clauseResultsNA, 'test', {
+      buildStatementLevelHTML: true
+    });
+
+    expect(statementResults[0].statementLevelHTML).toBeUndefined();
+  });
+
   test('simple HTML with generation with clause coverage styling', () => {
     // Ignore tabs and new lines
     const expectedHTML = getHTMLFixture('simpleCoverageAnnotation.html').replace(/\s/g, '');
--- a/test/unit/HTMLBuilder.test.ts
+++ b/test/unit/HTMLBuilder.test.ts
@@ -165,6 +165,17 @@ describe('HTMLBuilder', () => {
     expect(res.includes(falseStyleString)).toBeTruthy();
   });
 
+  test('statement-level HTML is added to statement results when calculation option is specified', () => {
+    const res = generateHTML(simpleMeasure, [elm], statementResults, trueClauseResults, 'test', {
+      buildStatementLevelHTML: true
+    });
+
+    expect(statementResults[0].statementLevelHTML).toBeDefined();
+    if (statementResults[0].statementLevelHTML) {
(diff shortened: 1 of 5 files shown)


test/unit/HTMLBuilder.test.ts
```diff
@@ -165,6 +165,17 @@ describe('HTMLBuilder', () => {
     expect(res.includes(falseStyleString)).toBeTruthy();
   });
 
+  test('statement-level HTML is added to statement results when calculation option is specified', () => {
+    const res = generateHTML(simpleMeasure, [elm], statementResults, trueClauseResults, 'test', {
+      buildStatementLevelHTML: true
+    });
+
+    expect(statementResults[0].statementLevelHTML).toBeDefined();
```
elsaperelli, line 173: The only other case I can think of for now is maybe one that makes sure there is no `statementLevelHTML` when the relevance is NA?

elsaperelli (changes requested):
Great job with this!! I went to test out implementing it with the prettyResults in fqm-testify and it took me 5 minutes rather than the 1000 it took me when having to parse through the whole HTML :) Some small stuff, but I did some testing and the html looks great and it is backwards compatible.

In terms of more testing, I agree that we should add more HTML testing but in my opinion that is out of scope for this task and that we can add it whenever we do that integration testing task.

I also ran the regression script which was successful since all the tests failed (they should with the added `statementLevelHTML` but the `html` object did not change.

elsaperelli (approved):
Lgtm!

hossenlopp (changes requested):
This looks good, one really small request/question, should/is this option be defaulted to `false`?

Running the CLI with CMS1028 and test-cases generates a 350MB detailed results file. The CLI should minimally default to false to speed up runs.

sarahmcdougall:
> This looks good, one really small request/question, should/is this option be defaulted to `false`?
> 
> Running the CLI with CMS1028 and test-cases generates a 350MB detailed results file. The CLI should minimally default to false to speed up runs.

Ah good point. The option is defaulted to true, but I agree it should default to false to speed up runs. I will make that change this morning 👍

hossenlopp (approved):
Now we can see a clean regression run for this.

PR #276, opened 2023-08-23
Additional alias highlighting

### LINKED ISSUE #275: Encounter.reasonCode is not recognized
### Summary
For the attached MeasureBundle and TestCaseBundle, looks like highlighting is not as expected. 

ValidEncounter.reasonCode is not recognized even though the attached TestCaseBundle has Encounter resource with reasonCode

I double checked and `Chronic recurrent major depressive disorder (disorder) - 2618002` is a valid code in `Major Depressive Disorder Active` Please Advice

image

#### External Tracking Ticket
[BONNIEMAT-1567](https://oncprojectracking.healthit.gov/support/browse/BONNIEMAT-1567)

### Expected Behavior
Expecting the Encounter.ReasonCode to be identified and used for execution.

[CMS161FHIR-v0.0.004-DENOMPass-DxDuringEncOV.txt](https://github.com/projecttacoma/fqm-execution/files/12345522/CMS161FHIR-v0.0.004-DENOM
[MeasureBundle.txt](https://github.com/projecttacoma/fqm-execution/files/12345524/MeasureBundle.txt)
Pass-DxDuringEncOV.txt)

v2.11.0

elsaperelli:
# Summary
Fixes #275 
This PR fixes the highlighting issue with the `ValidEncounter` alias in the above issue. This PR does not fix the clause coverage calculation issue and highlighting for `reasonCode`, that is being done in [this](https://github.com/cqframework/cql-execution/pull/314) PR in cql-execution.

## New behavior
Added new handling to `findAllLocalIdsInStatement` to account for aliases that are nested within queries from CQL that was authored using QICore. 

## Code changes
- `ClauseResultsHelpers.ts` - added branch to `findAllLocalIdsInStatement`  to handle query expressions that have a source that is an array with a single value but that value does not have a localId. In this case (that seems to occur when the CQL is authored using QICore), the alias is within this `source[0].expression.scope`. The alias localId is then going to be one less than the localId of the query statement. 
- `ClauseResultsHelpers.test.ts` - added a unit test for this new handling that uses CQL query expression authored using QICore.
- `QICoreQuery.cql` / `QICoreQuery.json` - test fixtures for the above unit test

# Testing guidance
- `npm run check`, `npm run test:integration`, run regression tests (all should pass)
- Run `detailed` calculation on the above issue with a measurement period start of 2024-01-01 and a measurement period end of 2024-12-31 and the debug option off.
- Look at the overall clause coverage html (`cd debug/html`). `ValidEncounter` should be highlighted (again, `reasonCode` will not be highlighted)
- Look at the measure logic highlighting. `ValidEncounter` should be highlighted green (`reasonCode` will be highlighted red)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2398,9 +2398,9 @@
       }
     },
     "node_modules/cql-execution": {
-      "version": "3.0.0",
-      "resolved": "https://registry.npmjs.org/cql-execution/-/cql-execution-3.0.0.tgz",
-      "integrity": "sha512-5nVt7XExWkyUbb76l+xlCnDd9c3cCuTdKgr88kkPnCy5SeFC+TYSo2pVb1zGVppWCOqVw2+ju8nQyuYX5xaYzA==",
+      "version": "3.0.1",
+      "resolved": "https://registry.npmjs.org/cql-execution/-/cql-execution-3.0.1.tgz",
+      "integrity": "sha512-RC6uxrzvrJImFvyKvYDNLdumCS7nufJobhz5abfV3ZeFxbPR6E2gqJcn0T2KxVh4y40+BqbSGABwgiFTEITCdQ==",
       "dependencies": {
         "@lhncbc/ucum-lhc": "^4.1.3",
         "immutable": "^4.1.0",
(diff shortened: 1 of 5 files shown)


sarahmcdougall (changes requested):
Great work on this - tested via several different methods (highlighting inspection, integration tests, unit tests, etc.) and everything looks good. Just have one small suggested change

Something we could consider in the future is refactoring the `test/unit/fixtures/elm/queries` directory since we have been adding a lot of files to it recently. Could put each test case in its own subdirectory (ex. a `QICoreQuery` subdirectory for the CQL and ELM used for the test). This is something we can keep in mind for end-of-sprint work, not something that needs to be addressed here

sarahmcdougall (approved):
The changes look good, and good call with adding the check for length 1 on the array 👍

PR #284, opened 2023-10-13
Data requirement profiles

elsaperelli:
# Summary
This PR adds the `meta.profile` found on the `templateId` of an ELMRetrieve (if it exists) to the `profile` attribute of a data requirement.

## New behavior
When the user runs dataRequirements on a measure that was authored using QI-Core, the data requirements output will now contain the `profile` attribute which is populated with the `meta.profile`, found in the `templateId` of the ELMRetrieve.

## Code changes
- `src/gaps/RetrievesFinder.ts` - gets the `templateId` from the ELMRetrieve and adds it to the `results` so it can eventually get added to the data requirement.
- `src/helpers/DataRequirementHelpers.ts` - function `addProfileToDataRequirement` adds the profile to the data requirement if a `templateId` exists on the ELMRetrieve.
- `src/types/Calculator.ts` - adds optional `templateId` to the `DataTypeQuery` interface.
- `test/unit/DataRequirements.test.ts` - adds a unit test for `addProfileToDataRequirement` and fixes some typos.
- `test/unit/RetrievesFinder.test.ts` - adds templateId to results.

# Testing guidance
- `npm run check`
- `npm run test:integration`
- Run data-requirements on a qi-core measure and a non qi-core measure and check that `profile` exists on the data requirements output and is correct. Question: do we want this to be populated for non qi-core measures as well? That is the current functionality.

### DIFF
--- a/src/helpers/DataRequirementHelpers.ts
+++ b/src/helpers/DataRequirementHelpers.ts
@@ -105,7 +105,6 @@ export async function getDataRequirements(
     allRetrieves.map(retrieve => {
       const dr = generateDataRequirement(retrieve);
       GapsInCareHelpers.addFiltersToDataRequirement(retrieve, dr, withErrors);
-      addProfileToDataRequirement(retrieve, dr);
       addFhirQueryPatternToDataRequirements(dr);
       return dr;
     }),
@@ -123,15 +122,6 @@ export async function getDataRequirements(
     },
     withErrors
   };
-}
-
-/**
- * Adds the meta.profile to the data requirement profile attribute
- */
-export function addProfileToDataRequirement(retrieve: DataTypeQuery, dataRequirement: fhir4.DataRequirement) {
-  if (retrieve.templateId) {
-    dataRequirement.profile = [retrieve.templateId];
-  }
 }
 
 /**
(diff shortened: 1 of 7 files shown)


src/helpers/DataRequirementHelpers.ts
```diff
@@ -124,6 +125,15 @@ export async function getDataRequirements(
   };
 }
 
+/**
+ * Adds the meta.profile to the data requirement profile attribute
+ */
+export function addProfileToDataRequirement(retrieve: DataTypeQuery, dataRequirement: fhir4.DataRequirement) {
```
hossenlopp, line 131: I think it would make sense to have this logic in the `generateDataRequirement` function since it is core information to all `dataRequirements`. Yes that might require a bit of a rework in that function but it could make it a bit cleaner in doing so. Create it with the type, add the profile if there is one, figure out wether or not to add the code or valueSet based filter, then return it.

hossenlopp (changes requested):
Suggestion for a small bit of rework that would make the existing code cleaner.

test/unit/RetrievesFinder.test.ts
```diff
@@ -17,6 +17,7 @@ const EXPECTED_VS_RETRIEVE_RESULTS: { results: DataTypeQuery[]; withErrors: Grac
       retrieveLocalId: '14',
       queryLocalId: undefined,
       retrieveLibraryName: 'SimpleQueries',
+      templateId: 'http://hl7.org/fhir/StructureDefinition/Condition',
```
p9g, line 20: Should these templateId's be using QI-Core profiles?
Note that CQL can have multiple "using" lines, e.g., using FHIR version '4.0.1' and using QICore version '4.1.1'
Would that result in multiple templateIds?

elsaperelli: This is seemingly where they are on the ELMRetrieves, and they seemingly only have one (the type of `templateId` on the `ELMRetrieve` interface in `ELMTypes.ts` is `string | undefined`). Unless you think that ELMRetrieve interface definition needs to be reworked. @hossenlopp maybe you have some more insight on this as well?

sarahmcdougall (commented):
Only one small suggestion, otherwise works well!

test/unit/RetrievesFinder.test.ts
hossenlopp: The `templateId` in the ELM Retrieve may or may not be populated by just a single profile. So this is correct.

The test data here really can be anything, we are just making sure we collect the profile from our scan of the ELM wether it is a base FHIR profile or a QICore one. Nothing is being done here to interpret it.

sarahmcdougall (approved):
Did some re-testing/sanity checking to make sure everything is still intact after the rebase, and it is 👍 Good work on this! And good work incorporating this new feature into unit tests!

hossenlopp (approved):

PR #286, opened 2023-10-17
Fix ValueSet reference finder to use proper resource field in relatedArtifacts

hossenlopp:
# Summary
Looks for the ValueSet references in the RelatedArtifact `resource` field. Previously this only looked for the references in the non-existent `url` field which many measures were using.

## New behavior
Looks at `resource` in addition to `url` when looking for referenced ValueSet URLs

## Code changes
`ValueSetHelper.ts` - Looks in `resource` as well.

# Testing guidance
Run unit tests. 
Add ValueSets to a recent measure package that has the fixed references, ensure ValueSets that are referenced in `relatedArtifacts` only are being pulled in. For example: `http://cts.nlm.nih.gov/fhir/ValueSet/2.16.840.1.113762.1.4.1029.206` for the Intensive Care Unit location is not something that will be in the `dataRequirements` and is only in the `relatedArtifacts` in CMS190.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,7 +32,7 @@
         "@typescript-eslint/parser": "^5.45.0",
         "axios-mock-adapter": "^1.19.0",
         "cql-translation-service-client": "^0.6.1",
-        "docsify": "^4.13.0",
+        "docsify": "^4.13.1",
         "ecqm-bundler": "^0.4.1",
         "eslint": "^7.5.0",
         "eslint-config-prettier": "^6.11.0",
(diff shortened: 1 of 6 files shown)


elsaperelli (approved):
Just one suggestion but other than that looks good to me!

I tested with adding valuesets to CMS190, CMS71, and CMS104.

I also tested that adding valuesets to ColorectalCancerScreeningFHIR and CMS108 did not change.

elsaperelli (approved):
Lgtm!

Issue #287, opened 2023-10-19
Coverage for Negation Aliases


jkotanchik-SB:
### Summary

Another round of Measures experiencing issues reaching 100% coverage. It may be an extension of #285 and I move these materials to that issue if it makes sense. 

The potential difference being the problematic aliases deal with negation, if that's a difference at all.

Impacted Measures:
1. **CMS 2 Preventive Care and Screening: Screening for Depression and Follow-Up PlanFHIR**

2. **CMS 22 Preventive Care and Screening: Screening for High Blood Pressure and Follow-Up DocumentedFHIR**

3. **CMS69 Preventive Care and Screening: Body Mass Index (BMI) Screening and Follow-Up PlanFHIR**

4. **CMS347 Statin Therapy for the Prevention and Treatment of Cardiovascular DiseaseFHIR**

#### External Tracking Ticket
https://oncprojectracking.healthit.gov/support/browse/BONNIEMAT-1631

### Expected Behavior
100% coverage

### Version or Commit
v1.3.1

### Inputs (e.g. Measure Bundle, Patient Bundle, CQL Library)
Measure Bundles:
1. [CMS2FHIR-v0.2.000-FHIR4.zip](https://github.com/projecttacoma/fqm-execution/files/13044977/CMS2FHIR-v0.2.000-FHIR4.zip)
2. [CMS22FHIR-v0.2.000-FHIR4.zip](https://github.com/projecttacoma/fqm-execution/files/13044976/CMS22FHIR-v0.2.000-FHIR4.zip)
3. [CMS69FHIR-v0.2.000-FHIR4.zip](https://github.com/projecttacoma/fqm-execution/files/13044975/CMS69FHIR-v0.2.000-FHIR4.zip)
4. [CMS347FHIR-v0.2.000-FHIR4.zip](https://github.com/projecttacoma/fqm-execution/files/13044972/CMS347FHIR-v0.2.000-FHIR4.zip)

MADiE Patient Exports:
1. [CMS2FHIR-v0.2.000-FHIR4-TestCases.zip](https://github.com/projecttacoma/fqm-execution/files/13044988/CMS2FHIR-v0.2.000-FHIR4-TestCases.zip)
2. [CMS22FHIR-v0.2.000-FHIR4-TestCases.zip](https://github.com/projecttacoma/fqm-execution/files/13044987/CMS22FHIR-v0.2.000-FHIR4-TestCases.zip)
[...]

PR #291, opened 2023-12-05
Not Equal and Not Equivalent clause coverage calculation handling

elsaperelli:
# Summary
#287 contains CMS69 which includes a `!~` (not equivalent) operation in the CQL that has some undesirable coverage calculation behavior due to the way Not Equivalent and Not Equal operations are translated from CQL to ELM json. This PR provides a workaround so that Not Equivalent and Not Equal clauses behave like Equal and Equivalent clauses. 

## New behavior
Before, in order to reach 100% clause coverage with CMS69, test cases where `BMIEncounter.class` WAS equivalent to 'virtual' AND where `BMIEncounter.class` was NOT equivalent to 'virtual' needed to be provided. This behavior did not match the clause coverage highlighting where only the not equivalent test case was needed for the clause to be highlighted as covered. Now, the coverage calculation matches the coverage highlighting. For CMS69 to reach 100% clause coverage, ONLY a test case where `BMIEncounter.class` is NOT equivalent (`!~`) to 'virtual' is needed. 

## Code changes
- `ClauseResultsHelpers.ts` - a workaround was added to for this specific case in `findAllLocalIdsInStatement` so that the result of the equal or equivalent is now the result of the not. 

# Testing guidance
- `npm run check`
- `npm run cli -- detailed -m <path-to-CMS69> --patients-directory <path-to-CMS69-patients-directory> --trust-meta-profile true --debug` on the master branch with the test cases provided in the should result in 96% clause coverage.
- Switch to this branch and run the same command. This should result in 96.2% clause coverage, showing that the `!~` clause is now included in the calculation.
- Another good way to test this would to be to `npm link` this branch to fqm-testify to play around with different test cases.

### DIFF
--- a/src/calculation/ClauseResultsHelpers.ts
+++ b/src/calculation/ClauseResultsHelpers.ts
@@ -214,6 +214,20 @@ export function findAllLocalIdsInStatement(
     } else if (k === 'type' && v === 'Literal' && statement.localId && statement.value === 'false') {
       // If this is a "Literal" expression whose value is false, mark that it `isFalsyLiteral` so we can interpret final results differently
       localIds[statement.localId] = { localId: statement.localId, isFalsyLiteral: true };
+    } else if (k === 'type' && v === 'Not') {
+      // If this is a "Not" expression, we will want to check if it's operand type is 'Equivalent' or 'Equal'
+      // If so, we will want to treat this as a 'Not Equivalent' or 'Not Equal' expression which we can do so
+      // by mapping the 'Equal' or 'Equivalent' clause localId to that of the 'Not' clause
+      if (statement.operand) {
+        if (statement.operand.type === 'Equivalent' || statement.operand.type === 'Equal') {
+          emptyResultClauses.push({
+            lib: libraryName,
+            aliasLocalId: statement.operand.localId,
+            expressionLocalId: statement.localId
+          });
+        }
+      }
+      localIds[statement.localId] = { localId: v };
     } else if (k === 'localId') {
       // else if the key is localId, push the value
       localIds[v] = { localId: v };
--- a/src/calculation/ClauseResultsHelpers.ts
+++ b/src/calculation/ClauseResultsHelpers.ts
@@ -230,7 +230,6 @@ export function findAllLocalIdsInStatement(
           });
         }
       }
-      localIds[statement.localId] = { localId: v };
     } else if (k === 'localId') {
       // else if the key is localId, push the value
       localIds[v] = { localId: v };
--- a/src/calculation/ClauseResultsHelpers.ts
+++ b/src/calculation/ClauseResultsHelpers.ts
@@ -219,7 +219,10 @@ export function findAllLocalIdsInStatement(
       // If so, we will want to treat this as a 'Not Equivalent' or 'Not Equal' expression which we can do so
       // by mapping the 'Equal' or 'Equivalent' clause localId to that of the 'Not' clause
       if (statement.operand) {
-        if (statement.operand.type === 'Equivalent' || statement.operand.type === 'Equal') {
+        if (
+          statement.operand.type === 'Equivalent' ||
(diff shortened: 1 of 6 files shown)


src/calculation/ClauseResultsHelpers.ts
```diff
@@ -214,6 +214,20 @@ export function findAllLocalIdsInStatement(
     } else if (k === 'type' && v === 'Literal' && statement.localId && statement.value === 'false') {
       // If this is a "Literal" expression whose value is false, mark that it `isFalsyLiteral` so we can interpret final results differently
       localIds[statement.localId] = { localId: statement.localId, isFalsyLiteral: true };
+    } else if (k === 'type' && v === 'Not') {
+      // If this is a "Not" expression, we will want to check if it's operand type is 'Equivalent' or 'Equal'
+      // If so, we will want to treat this as a 'Not Equivalent' or 'Not Equal' expression which we can do so
+      // by mapping the 'Equal' or 'Equivalent' clause localId to that of the 'Not' clause
+      if (statement.operand) {
+        if (statement.operand.type === 'Equivalent' || statement.operand.type === 'Equal') {
```
hossenlopp, line 222: This whole set of logic could use a test case. It's one of the few things in here that isn't tested.

hossenlopp (changes requested):
Just needs some unit testing.

src/calculation/ClauseResultsHelpers.ts
elsaperelli: Good call! I added a unit test for the `!~`...however, I do not think I am able to add a unit test for the `!=` logic at the moment since an update to the translator will be coming soon that puts a localId on the `Equal` expression (but it doesn't right now). Am I correct in this? @birick1 may have additional insight as well

hossenlopp (approved):
Looks good and added test covers most of this.

src/calculation/ClauseResultsHelpers.ts
hossenlopp, line 222: We should also make sure that `operand.localId` exists before attempting anything here.

hossenlopp (changes requested):
Regression test shows some changes that were unexpected.

src/calculation/ClauseResultsHelpers.ts
```diff
@@ -214,6 +214,22 @@ export function findAllLocalIdsInStatement(
     } else if (k === 'type' && v === 'Literal' && statement.localId && statement.value === 'false') {
       // If this is a "Literal" expression whose value is false, mark that it `isFalsyLiteral` so we can interpret final results differently
       localIds[statement.localId] = { localId: statement.localId, isFalsyLiteral: true };
+    } else if (k === 'type' && v === 'Not') {
+      // If this is a "Not" expression, we will want to check if it's operand type is 'Equivalent' or 'Equal'
+      // If so, we will want to treat this as a 'Not Equivalent' or 'Not Equal' expression which we can do so
+      // by mapping the 'Equal' or 'Equivalent' clause localId to that of the 'Not' clause
+      if (statement.operand) {
```
hossenlopp, line 221: The `statement.operand.localId` check should be moved here. Additionally we should ensure that this `statement.localId` also exists. We can only do this mapping if BOTH have localIds.

src/calculation/ClauseResultsHelpers.ts
```diff
@@ -214,6 +214,22 @@ export function findAllLocalIdsInStatement(
...
+        if (
+          statement.operand.type === 'Equivalent' ||
+          (statement.operand.type === 'Equal' && statement.operand.localId)
```
hossenlopp, line 224: Move this `localId` check above.

hossenlopp (changes requested):

hossenlopp (approved):

hossenlopp (approved):
New tests are looking good

lmd59 (approved):

PR #293, opened 2024-01-29
Refactor to separate gaps and data requirements

lmd59:
# Summary
Consolidate helpers, rename files as helpful, and copy functionality or move to shared helpers so that gaps and dataRequirements can function independently without cross referencing each other.

## New behavior
None

## Code changes
- Moved QueryFilterParser and RetrievesHelper to elm helpers
- Moved ClauseResultsHelper, rename ReportBuilderFactory
- Moved GapsReportBuilder to calculation and rename import to match file name, rename test to GapsReportBuilder.test
- Remove GapsReportBuilder dependency from DataRequirementHelpers and replicate dependeny function
- Moved detailedFilter functions and codeLookup and flattenFilters to QueryFilterParser
- Test and other import updates for all of these

# Testing guidance

- `npm run check`
- Run integration tests
- Test other basic functionality by running gaps and dataRequirements on your favorite test measure, test patient, and test library


elsaperelli (commented):
I am still in the process of reviewing this PR, but I did notice that the regression script (command: `./regression/run-regression.sh -b master -v`) fails and I don't think it should. At first glance, I suspect it has to do with simply the movement of files, but I will look into it more.

src/helpers/elm/QueryFilterParser.ts
```diff
@@ -1204,3 +1205,219 @@ export function interpretComparator(
...
+    const a: AnyFilter[] = [];
+    (filter as AndFilter).children.forEach(c => {
+      a.push(...flattenFilters(c));
+    });
+
+    return a;
+  }
+}
+
+describe('Flatten Filters', () => {
```
elsaperelli, line 1367: Ah yes, I believe we found the culprit! Looks like this snuck in here :)

src/cli.ts
```diff
@@ -8,7 +8,7 @@ import {
   calculateMeasureReports,
   calculateGapsInCare,
   calculateRaw,
-  calculateDataRequirements,
+  calculateMeasureDataRequirements,
```
elsaperelli, line 11: Are we sure we want to change the name of this right now? That would make this a breaking change potentially or am I wrong? If we do indeed want to change it, we should probably update the README.

src/helpers/DataRequirementHelpers.ts
```diff
@@ -511,3 +347,67 @@ export function getFlattenedRelatedArtifacts(
...
 }
+
+/**
+ *
+ * @param q The query which contains the filters to add to the data requirement
+ * @param dataRequirement Data requirement to add date filters to
+ * @param withErrors Errors object which will eventually be returned to the user if populated
+ * @returns void, but populated the dataRequirement filters
+ */
+export function addFiltersToDataRequirement(
```
elsaperelli, line 358: What's the difference between this function and the one that was renamed to `addGapFiltersToDataRequirement` in `GapsReportBuilder`?

elsaperelli (changes requested):
Woo refactoring! A tough thing to review but very necessary changes. I am hesitant about changing `calculateDataRequirements` name at the moment so I want to see what you think. I also think it's possible that there's a duplicate function but I am not sure.

src/cli.ts
lmd59: Oh- good call. I will revert this.

src/helpers/DataRequirementHelpers.ts
lmd59: There is no difference. This is copied so that we can change things around on data requirements without interfering with gaps. Will DM with more info.

elsaperelli (approved):
Nice!

hossenlopp (approved):
Lots of moving things around. But this does look good.

PR #299, opened 2024-03-14
Uncoverage

hossenlopp:
# Summary
New option to show which clauses are uncovered as a new set of output. Also adds details on coverage including the list of clauses that were not coverage.

Additionally coverage html calculation was made considerably faster (~200x). This should be VERY noticeable when run with larger sets of patients. 

## New behavior
New options in calculator that are false by default:
 - `calculateClauseUncoverage`: Calculates HTML that highlights clauses that are not covered by the set of patients.
 - `calculateCoverageDetails`: Calculates the total count of covered and uncovered clauses and lists the location of the uncovered clauses.
 
These new options will run in the CLI and the coverageDetails will be output as it's own debug output file. This is intended to help with pinpointing coverage issues as it can tell you exactly where in the ELM there are gaps in the coverage.

## Code changes
`HTMLBuilder.ts` - Reworked logic for calculating coverage to collect unique lists of covered and uncovered clause results. Reworked coverage highlighting to work off these smaller lists to vastly speed up HTML generation. Added a helper to highlight specifically for uncoverage.
`Calculator.ts` - Now can handle the two new options and writes out the coverageDetails to a new debug file.
`types/Calculator.ts` - New types for the new options and output.
`cli.ts` - Defaults `calculateClauseUncoverage` and `calculateCoverageDetails` to true.

# Testing guidance
Test with any measures that feel appropriate and look for any differences in coverage highlighting or percentage. Check to make sure that uncoverage highlighting looks sane.

### DIFF
--- a/README.md
+++ b/README.md
@@ -933,6 +933,8 @@ have a blue background and dashed underline in the highlighted HTML, indicating 
 
 The highlighted HTML also provides an approximate percentage for what percentage of the measure logic is covered by the patients that were passed in to execution.
 
+This option, `calculateClauseCoverage`, defaults to enabled. When running the CLI in `--debug` mode this option will be enabled and output to the debug directory.
+
 ![Screenshot of Highlighted Clause Coverage HTML](./static/coverage-highlighting-example.png)
 
 HTML strings are returned for each group defined in the `Measure` resource in the lookup object, `groupClauseCoverageHTML`, which maps `group ID -> HTML string`
@@ -959,6 +961,8 @@ const { results, groupClauseCoverageHTML } = await Calculator.calculate(measureB
 
 `fqm-execution` can also generate highlighted HTML that indicate which pieces of the measure logic did NOT have a "truthy" value during calculation. This is the complement to coverage. Clauses that are not covered will be highlighted red.
 
+This option, `calculateClauseUncoverage`, defaults to disabled. When running the CLI in `--debug` mode this option will be enabled and output to the debug directory.
+
 HTML strings are returned for each group defined in the `Measure` resource in the lookup object,`groupClauseUncoverageHTML`, which is structured similarly to Group Clause Coverage.
 
 ```typescript
@@ -982,6 +986,8 @@ const { results, groupClauseUncoverageHTML } = await Calculator.calculate(measur
 ### Coverage Details
 
 Details on clause coverage can also be returned. This includes a count of how many clauses there are, how many are covered and uncovered, and information about which clauses are uncovered.
+
+This option, `calculateCoverageDetails`, defaults to disabled. When running the CLI in `--debug` mode this option will be enabled and output to the debug directory.
 
 This information is returned for each group defined in the `Measure` resource in the lookup object,`groupClauseCoverageDetails`, which maps `group ID -> coverageDetails`. See example below for structure of this object.
 
--- a/README.md
+++ b/README.md
@@ -971,7 +971,7 @@ const { results, groupClauseUncoverageHTML } = await Calculator.calculate(measur
 // groupClauseUncoverageHTML
 /*      ^?
         {
-          'population-group-1': '<div><h2> population-group-1 Clause Uncoverage: X clauses</h2> ...'
+          'population-group-1': '<div><h2> population-group-1 Clause Uncoverage: X of Y clauses</h2> ...'
           ...
         }
 
--- a/README.md
+++ b/README.md
@@ -935,7 +935,7 @@ The highlighted HTML also provides an approximate percentage for what percentage
 
 ![Screenshot of Highlighted Clause Coverage HTML](./static/coverage-highlighting-example.png)
 
-HTML strings are returned for each group defined in the `Measure` resource as a lookup object `groupClauseCoverageHTML` which maps `group ID -> HTML string`
+HTML strings are returned for each group defined in the `Measure` resource in the lookup object, `groupClauseCoverageHTML`, which maps `group ID -> HTML string`
 
 ```typescript
 import { Calculator } from 'fqm-execution';
@@ -957,9 +957,9 @@ const { results, groupClauseCoverageHTML } = await Calculator.calculate(measureB
 
 ### Uncoverage Highlighting
 
-`fqm-execution` can also generate highlighted HTML that indicate which pieces of the measure logic which did NOT have a "truthy" value during calculation. This is the complement to coverage. Clauses that are not covered will be highlighted red.
+`fqm-execution` can also generate highlighted HTML that indicate which pieces of the measure logic did NOT have a "truthy" value during calculation. This is the complement to coverage. Clauses that are not covered will be highlighted red.
 
-HTML strings are returned for each group defined in the `Measure` resource as a lookup object `groupClauseUncoverageHTML` similarly to Group Clause Coverage.
+HTML strings are returned for each group defined in the `Measure` resource in the lookup object,`groupClauseUncoverageHTML`, which is structured similarly to Group Clause Coverage.
 
 ```typescript
 import { Calculator } from 'fqm-execution';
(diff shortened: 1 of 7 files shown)


src/calculation/Calculator.ts
```diff
@@ -41,6 +41,7 @@ import { clearElmInfoCache } from '../helpers/elm/ELMInfoCache';
 import _, { omit } from 'lodash';
 import { ELM } from '../types/ELMTypes';
 import { getReportBuilder } from '../helpers/ReportBuilderFactory';
+import { option } from 'commander';
```
lmd59, line 44: Looks like an accidental import

src/types/Calculator.ts
```diff
@@ -395,6 +400,18 @@ export interface CalculationOutput<T extends CalculationOptions> extends Calcula
...
   coverageHTML?: string;
   groupClauseCoverageHTML?: Record<string, string>;
+  groupClauseUncoverageHTML?: Record<string, string>;
+  groupClauseCoverageDetails?: Record<string, ClauseCoverageDetails>;
+}
+
+export interface ClauseCoverageDetails {
+  totalClauses: number;
+  coveredClauses: number;
+  uncoveredClauses: {
```
lmd59, line 410: Because the other two clauses details are numbers, this naming seems a a bit incongruous. It would like me to expect all of them to be the same type. Maybe we change the other two to `totalClausCount` and `coveredClauseCount` or something like that so that it doesn't lead to the expectation that `uncoveredClauses` also has a count?

src/calculation/HTMLBuilder.ts
```diff
@@ -317,5 +379,18 @@ export function calculateClauseCoverage(relevantStatements: StatementResult[], c
     allRelevantClauses.filter(clause => clause.final === FinalResult.TRUE),
     (c1, c2) => c1.libraryName === c2.libraryName && c1.localId === c2.localId
   );
-  return ((coveredClauses.length / allUniqueClauses.length) * 100).toPrecision(3);
+
+  const uncoveredClauses = allUniqueClauses.filter(c => {
+    if (coveredClauses.find(coveredC => c.libraryName === coveredC.libraryName && c.localId === coveredC.localId)) {
```
lmd59, line 384: suggest simplify this if/else boolean
```suggestion
    return !(coveredClauses.find(coveredC => c.libraryName === coveredC.libraryName && c.localId === coveredC.localId)) {
```

src/calculation/Calculator.ts
hossenlopp: blargh it is!

hossenlopp: Resolved in [f132302](https://github.com/projecttacoma/fqm-execution/pull/299/commits/f1323025f5f095fbc6f26127fe85d26b19b7ee6d)

src/types/Calculator.ts
hossenlopp: Renamed in ca98110 and added uncovered count

src/calculation/HTMLBuilder.ts
hossenlopp: Good call! Implemented this in f4b34e6.

elsaperelli (changes requested):
Woo uncoverage!!! Super excited for this to be a new part of fqm-e :) Code all looked good to me, I would just suggest adding a section about this to the README. 

Another suggestion although I don't know how doable/worth it it would be is to include some sort of "hidden" uncovered clause count? For example, when I run this branch on the MAT6725 measure bundle that was created with the new translator, the uncoverage HTML says 2 uncovered clauses but only one clause is highlighted in red. This is because the other uncovered clause is not a part of the HTML and therefore "hidden". Is there a way we could make this known in our debug output somehow? I think it would be extremely helpful in our future work with coverage highlighting and calculation with the new translator. Also happy to make this an additional task if out of scope.

lmd59:
We're seeing some differences in the "complementary" nature of the coverage highlighting and the uncoverage highlighting. Example w/ all available test patients in ecqm-content-r4-2021/bundles/measure/CervicalCancerScreeningFHIR...
Coverage:
Screenshot 2024-03-19 at 3 06 29 PM
Uncoverage:
Screenshot 2024-03-19 at 3 06 55 PM

Might be nice to do a quick pass at re-unhighlighting the first part of the clause, but is not critical to this PR (can be tackled later if needed).

elsaperelli (changes requested):
I still want to do some testing and actually look at the new complementary uncoverage HTML, but I figured I would put up the few comments I have on README wording :) 

Yay uncoverage!

README.md
```diff
@@ -953,6 +955,64 @@ const { results, groupClauseCoverageHTML } = await Calculator.calculate(measureB
...
+
+
+*/
+```
+
+### Coverage Details
+
+Details on clause coverage can also be returned. This includes a count of how many clauses there are, how many are covered and uncovered, and information about which clauses are uncovered.
+
+This information is returned for each group defined in the `Measure` resource as a lookup object `groupClauseCoverageDetails` similarly to Group Clause Coverage.
```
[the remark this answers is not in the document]
hossenlopp: Addressed in 1e744b9.

README.md
```diff
@@ -953,6 +955,64 @@ const { results, groupClauseCoverageHTML } = await Calculator.calculate(measureB
 */
 ```
 
+### Uncoverage Highlighting
+
+`fqm-execution` can also generate highlighted HTML that indicate which pieces of the measure logic which did NOT have a "truthy" value during calculation. This is the complement to coverage. Clauses that are not covered will be highlighted red.
+
+HTML strings are returned for each group defined in the `Measure` resource as a lookup object `groupClauseUncoverageHTML` similarly to Group Clause Coverage.
```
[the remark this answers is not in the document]
hossenlopp: Addressed in 1e744b9.

README.md
```diff
@@ -953,6 +955,64 @@ const { results, groupClauseCoverageHTML } = await Calculator.calculate(measureB
 */
 ```
 
+### Uncoverage Highlighting
+
+`fqm-execution` can also generate highlighted HTML that indicate which pieces of the measure logic which did NOT have a "truthy" value during calculation. This is the complement to coverage. Clauses that are not covered will be highlighted red.
```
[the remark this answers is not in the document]
hossenlopp: Addressed in 1e744b9.

README.md
```diff
@@ -953,6 +955,64 @@ const { results, groupClauseCoverageHTML } = await Calculator.calculate(measureB
...
+import { Calculator } from 'fqm-execution';
+
+const { results, groupClauseUncoverageHTML } = await Calculator.calculate(measureBundle, patientBundles, {
+  calculateClauseUncoverage: true /* false by default */
+});
+
+// groupClauseUncoverageHTML
+/*      ^?
+        {
+          'population-group-1': '<div><h2> population-group-1 Clause Uncoverage: X clauses</h2> ...'
```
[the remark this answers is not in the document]
hossenlopp: Good call! Added this in 1f25221.

elsaperelli (approved):
This is an incredible addition to fqm-execution. I did a LOT of testing to look for any discrepancies in the complementary highlighting and to make sure the regular clause coverage highlighting and calculation continued to work the same way. Noting the measures that I tested here for future reference (mostly taken from fqm-execution coverage issues since those measures are the ones that would have benefitted from this feature): CMS104, CMS108, CMS135, CMS144, CMS145, CMS159, CMS165, CMS190, CMS22, CMS334, CMS645, CMS646, CMS69, CMS72, MAT6725.

The only one in which I found an issue with the complementary highlighting was CMS144 and it was the `NoBetaBlockerOrdered.medication` piece of CQL which we currently have a backlog task to address and therefore I do not think it is an issue in the scope of this PR (just noting for completeness).

Everything makes sense to me! 👏 🥳

hossenlopp:
> Does the CLI have an option for uncoverage? Can README say something about that?

This does run in the CLI when you use `--debug`. I added info to the README to clarify this in 3c91aa0.

lmd59 (approved):

PR #300, opened 2024-03-29
Sort mustSupports, compare DRs with direct reference codes

elsaperelli:
# Summary
While working on data requirements tasking, Hoss and Lauren noticed some issues with the data requirements comparison script now that mustSupports are being incorporated. There were three issues and this addresses two of them: the mustSupport arrays were not being sorted before being compared (i.e. ['code', 'status'] did not equal ['status', 'code'] and data requirements that are not referenced by valueSets but rather through direct reference codes were not being analyzed for mustSupport comparison. 

## New behavior
- Sorts the mustSupports arrays before doing a comparison.
- Compares mustSupports of data requirements that are referenced by a direct reference code rather than a ValueSet. 

## Code changes
- `data-requirements/summary-compare.ts` - implemented above changes 

# Testing guidance
- `npm run check`
- Run the summary-compare script: `./summary-compare.sh`
- Make sure mustSupports pass when the arrays have the same contents 
- Make sure data requirements using direct reference codes are included in the must supports comparison


data-requirements/summary-compare.ts
```diff
@@ -197,6 +216,23 @@ async function main() {
...
+          const equalMustSupportsByCode = _.isEqual(elmParserMustSupports?.sort(), fqmEMatchMustSupportsByCode?.sort());
+
+          if (!equalMustSupportsByCode) {
+            console.log(`${FG_RED}%s${RESET}:`, `MUST SUPPORTS FAIL (${key}-${elmParserDirectCode})`);
+            console.log(`fqm-execution has the following mustSupports: ${fqmEMatchMustSupportsByCode ?? ''}`);
+            console.log(`elm-parser-for-ecqms has the following mustSupports: ${elmParserMustSupports}`);
+          } else {
+            console.log(
+              `${FG_GREEN}%s${RESET}: matching mustSupports`,
+              `MUST SUPPORTS PASS (${key}-${elmParserValueSet})`
```
lmd59, line 233:
```suggestion
              `MUST SUPPORTS PASS (${key}-${elmParserDirectCode})`
```

data-requirements/summary-compare.ts
```diff
@@ -179,13 +197,14 @@ async function main() {
       elmParserDRByKey.forEach(dr => {
         const elmParserMustSupports = dr.mustSupport;
         const elmParserValueSet = dr.codeFilter?.find(cf => cf.valueSet)?.valueSet;
```
lmd59, line 199:
```suggestion
        // ignore valueset codefilter if it's a "value" type, TODO: get rid of explicit value check and
        // use primary code path from proposed extension added to dr, once implemented/available
        const elmParserValueSet = dr.codeFilter?.find(cf => cf.valueSet && cf.path !== 'value')?.valueSet;
```

data-requirements/summary-compare.ts
```diff
@@ -179,13 +197,14 @@ async function main() {
       elmParserDRByKey.forEach(dr => {
         const elmParserMustSupports = dr.mustSupport;
```
lmd59, line 198:
```suggestion
        const elmParserMustSupports = dr.mustSupport?.filter(ms => !ms.includes('.'));
```
I'm currently filtering out secondary properties when doing the comparison. I think we could technically go either way on this one, but thought I'd note it here for consideration.

elsaperelli: Since the goal of the data requirements comparison script is to see the mustSupports output of both `fqm-execution` and `elm-parser-for-ecqms`, I am a little hesitant to do this filtering because then we are not seeing the full output for the `elm-parser-for-ecqms`? I think I am potentially missing some context here though, so I am happy to discuss more!

data-requirements/summary-compare.ts
elsaperelli: Sorry, for some reason I thought I added this but I guess I didn't!

lmd59 (approved):

PR #305, opened 2024-05-22
Fix QueryFilter oarsing on optional Query relationship field

hossenlopp:
# Summary

Fixes an issue found in CMS816 bundle from May 2024 connectathon. The ELM in one of the libraries has a Query expression that has the `relationship` field not existent instead of as an empty array. This is allowed by the CQL logical specification. This issue can cause the `dataRequirements` operation to fail.

## New behavior

Only copies the relationship info if it actually exists.

## Code changes
- `src/helpers/elm/QueryFilterParser.ts` - Check of existence of the `relationship` field before adding it the list to parse.
- `src/types/ELMTypes.ts` - Update the ELM type for Query to make `relationship` optional.

# Testing guidance
Ensure CMS861 does not error out when doing `dataRequirement`.


elsaperelli (approved):
Lgtm!

lmd59 (approved):

Issue #306, opened 2024-07-02
Calculation for Stratification does not consider associated population


mcmcphillips:
### Summary
Calculations for stratifications do not take into account their associated populations.
Given a measure with a stratification that evaluates to true, and an associated population that evaluates to false, the stratification result is still true. 

### Expected Behavior
Given a stratification that evaluates true, with it's associated population that does not evaluate true, the stratification result should return false. 

### Version or Commit
    "fqm-execution": "^1.3.3"
### Inputs (e.g. Measure Bundle, Patient Bundle, CQL Library)

options: 
 {
[measureBundle.json](https://github.com/user-attachments/files/16074142/measureBundle.json)
[patientBundle.json](https://github.com/user-attachments/files/16074150/patientBundle.json)
[valuesets.txt](https://github.com/user-attachments/files/16074151/valuesets.txt)

          includeClauseResults: false,
          trustMetaProfile: true,
          buildStatementLevelHTML: true,
          measurementPeriodStart: 2025-01-01T00:00:00.000+00:00
          measurementPeriodEnd: 2025-12-31T23:59:59.999+00:00,
        },

### cql-to-elm Version Used for Measure Logic Translation (if known)
3.10.0

hossenlopp:
Hi @mcmcphillips,

Thanks for submitting this issue with most of the data we need. Just have one question, are you using the "detailed" output of fqm-execution or the MeasureReport output?

In general in the detailed results the stratifications are treated as a "filter" to be applied to the population(s) they belong to. So they will be evaluated and be true or false regardless of results of the population(s) results. When creating a MeasureReport from the detailed results, the stratified results will have stratifier applied to the population(s) and if the patient isn't in a population, they will not count in that stratification. This may be what your implementation should look to doing if you are working from the detailed results.

The detailed results always evaluate the stratifications since in a measure testing environment it may be useful to see the results of the stratifications separately from the population(s).

In a future sprint we will prioritize looking deeper into this issue and how the usage of cqfm-appliesTo may affect fqm-execution.

-Chris

mcmcphillips:
Hi @hossenlopp 

We've been using fqm-execution's Calculator.calculate method with the above provided inputs / options that returns a measureReport. 

The results we've been accessing have been located from the result of Calulator.calculate as
result.results.detailedResults.stratifierResults which come in the shape of 

`{
    "strataCode": "strata-1",
    "result": true,
[...]

PR #308, opened 2024-08-05
appliesResult to StratifierResult in detailed results output

elsaperelli:
# Summary
This PR adds an optional attribute to the `StratifierResult` interface so that we can report the result of a stratifier that includes an `appliesTo` extension for Patient-based Measures only. 

## New behavior
The [`appliesTo`](https://hl7.org/fhir/us/cqfmeasures/STU4/StructureDefinition-cqfm-appliesTo.html) extension indicates which population a stratifier should apply to. If it is included on a Measure.group.stratifier, we now report the result of the stratifier AND the result of the population it applies to on the `appliesResult` attribute on the StratifierResult. The previous `result` attribute remains as it was: simply the result of the stratifier. The DetailedResults now contain the raw stratifier result in result (like before) as well as the result of the stratifier with the population result it applies to taken into consideration in `appliesResult`. 

## Code changes
- `src/calculation/DetailedResultsBuilder.ts` - add AND logic of population result and stratifier result to appliesResult attribute of StratifierResults for patient based measures
- `src/types/Calculator.ts` - add optional appliesResult attribute to `StratifierResult` interface
- `test/unit/DetailedResultsBuilder.test.ts` - unit tests for described changes

# Testing guidance
- Run detailed results with the measure and patient bundles provided in #306 . This can be done in a variety of ways but I personally just the cli: `npm run cli -- detailed -m <path to mb> --p <path to patient bundle (I think the file they provide is in an array so just make sure to get rid of that)> -s 2025-01-01 -e 2025-12-31 --trust-meta-profile true -o --debug`
- Inspect the detailedResults.json. Make sure the stratifierResults.result are the same as before and the stratifierResults.appliesResult are correct based on the population result that stratifier applies to.

### DIFF
--- a/src/calculation/DetailedResultsBuilder.ts
+++ b/src/calculation/DetailedResultsBuilder.ts
@@ -33,6 +33,9 @@ export function createPopulationValues(
     populationResults = popAndStratResults.populationResults;
     stratifierResults = popAndStratResults.stratifierResults;
     populationResults = handlePopulationValues(populationResults, populationGroup, measureScoringCode);
+    if (stratifierResults) {
+      stratifierResults = handleStratificationValues(populationGroup, populationResults, stratifierResults);
+    }
   } else {
     // episode of care based measure
     // collect results per episode
@@ -45,6 +48,9 @@ export function createPopulationValues(
       const popAndStratResults = createPatientPopulationValues(populationGroup, patientResults);
       populationResults = popAndStratResults.populationResults;
       stratifierResults = popAndStratResults.stratifierResults;
+      if (stratifierResults) {
+        stratifierResults = handleStratificationValues(populationGroup, populationResults, stratifierResults);
+      }
     } else {
       populationResults = [];
       stratifierResults = [];
@@ -97,6 +103,9 @@ export function createPopulationValues(
           }
         });
       });
+      if (stratifierResults) {
+        stratifierResults = handleStratificationValues(populationGroup, populationResults, stratifierResults);
+      }
     }
   }
   const detailedResult: DetailedPopulationGroupResult = {
@@ -304,27 +313,11 @@ export function createPatientPopulationValues(
     populationGroup.stratifier.forEach(strata => {
       if (strata.criteria?.expression) {
         const value = patientResults[strata.criteria?.expression];
-
-        // if the cqfm-appliesTo extension is present, then we want to consider the result of that
-        // population in our stratifier result
-        const appliesToExtension = strata.extension?.find(
-          e => e.url === 'http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqfm-appliesTo'
-        );
-
-        let popValue = true;
-        if (appliesToExtension) {
-          const popCode = appliesToExtension.valueCodeableConcept?.coding?.[0].code;
-          if (popCode) {
-            popValue = patientResults[popCode];
-          }
-        }
         const result = isStatementValueTruthy(value);
-        const appliesResult = isStatementValueTruthy(value && popValue);
 
         stratifierResults?.push({
           strataCode: strata.code?.text ?? strata.id ?? `strata-${strataIndex++}`,
           result,
-          appliesResult,
           ...(strata.id ? { strataId: strata.id } : {})
         });
       }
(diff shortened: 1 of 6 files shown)


lmd59 (approved):
Looks good to me!

src/calculation/DetailedResultsBuilder.ts
```diff
@@ -304,10 +304,27 @@ export function createPatientPopulationValues(
...
+        // population in our stratifier result
+        const appliesToExtension = strata.extension?.find(
+          e => e.url === 'http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqfm-appliesTo'
+        );
+
+        let popValue = true;
+        if (appliesToExtension) {
+          const popCode = appliesToExtension.valueCodeableConcept?.coding?.[0].code;
+          if (popCode) {
+            popValue = patientResults[popCode];
```
hossenlopp, line 318: This isn't the right place to get the results for population as the `patientResults` is the raw data from the engine. The define statement that provides the results might not match the code. Should look back at the `populationResults` that have already been collected and find by `populationType`.

src/calculation/DetailedResultsBuilder.ts
```diff
@@ -304,10 +304,27 @@ export function createPatientPopulationValues(
...
+          }
+        }
         const result = isStatementValueTruthy(value);
+        const appliesResult = isStatementValueTruthy(value && popValue);
```
hossenlopp, line 322: This call to `isStatementValueTruthy` shouldn't be needed as the `&&` should be done on the `result` and the boolean result found in the `populationResults`.

src/calculation/DetailedResultsBuilder.ts
hossenlopp: Additionally the populationResults will not be accurate at this point and should happen after the call to `handlePoulationValues` on line 35.

hossenlopp (changes requested):
The place where the appliedResult is created needs to be moved to have accurate results.

src/calculation/DetailedResultsBuilder.ts
```diff
@@ -320,6 +330,39 @@ export function createPatientPopulationValues(
   };
 }
 
+export function handleStratificationValues(
+  populationGroup: fhir4.MeasureGroup,
+  populationResults: PopulationResult[],
+  stratifierResults: StratifierResult[]
+): StratifierResult[] {
+  if (populationGroup.stratifier) {
+    populationGroup.stratifier.forEach(strata => {
```
hossenlopp, line 339: You may want to iterate over the stratifierResults instead of the group's stratifiers and do the lookup the opposite way around similar to how it is done in the MeasureReport builder. https://github.com/projecttacoma/fqm-execution/blob/0b752a1a5da6a3aa5ee4d1c5a5df50b292ff5fb5/src/calculation/MeasureReportBuilder.ts#L237
The code there also handles to two different ways the stratifiers are referenced via id or codes.

hossenlopp (changes requested):
A handful of things in the new function.

It should also be run on the individual episode results after `handlePopulationValues` at https://github.com/projecttacoma/fqm-execution/blob/master/src/calculation/DetailedResultsBuilder.ts#L439.

The code in the measure report builder that does appliesTo logic could be removed and the `appliedResult` could be used instead of the `result`.

hossenlopp (approved):
Looks good to me.

hossenlopp (approved):
I like the Stratification addition to the README. This looks ready to go!

test/unit/MeasureReportBuilder.test.ts
```diff
@@ -475,10 +483,10 @@ describe('MeasureReportBuilder Static', () => {
       });
     });
 
-    test('should generate a summary MeasureReport whose stratifierResults only contain one population in the stratum', () => {
+    test('should generate a summary MeasureReport whose stratifierResults only contain all populations', () => {
```
lmd59, line 486:
```suggestion
    test('should generate a summary MeasureReport whose stratifierResults contain all populations', () => {
```

lmd59 (approved):
Made one optional test description clarification

Issue #313, opened 2024-09-20
Coverage Drops After Incorporating Translator v3.14.0


jkotanchik-SB:
### Summary

After updating to the 3.14.0 cql-to-elm Translator, our QA team identified reduced Coverage values on some hand-picked measures:

| # | Measure | 3.14.0 | 3.3.2 |
| - | --------- |-----| -------|
|1| CMS135FHIR | 99% | 100% |
|2| CMS144FHIR | 96% | 100% |
|3| CMS145FHIR | 99% | 100% |
|4| CMS159FHIR | 99% | 100% |
|5| CMS645FHIR | 94% | 99% |
|6| CMS646FHIR | 95% | 99% |
|7| CMS771FHIR | 94% | 100% |

#### External Tracking Ticket
N/A

### Expected Behavior

Coverage matches when using older and current cql-to-elm version.

### Version or Commit
`v1.5.0`

### Inputs (e.g. Measure Bundle, Patient Bundle, CQL Library)
[MeasureBundles.zip](https://github.com/user-attachments/files/17077786/MeasureBundles.zip)

Measure Exports:
1. [CMS145FHIR-v0.2.000-FHIR4.zip](https://github.com/user-attachments/files/17561456/CMS145FHIR-v0.2.000-FHIR4.zip)
2. [CMS72FHIR-v0.7.000-FHIR4.zip](https://github.com/user-attachments/files/17561459/CMS72FHIR-v0.7.000-FHIR4.zip)
3. [CMS1028FHIR-v0.1.000-FHIR4.zip](https://github.com/user-attachments/files/17561460/CMS1028FHIR-v0.1.000-FHIR4.zip)

Patient Bundles:
1. [CMS135FHIR-v0.2.000-FHIR4-TestCases.zip](https://github.com/user-attachments/files/17077736/CMS135FHIR-v0.2.000-FHIR4-TestCases.zip)
3. [CMS144FHIR-v1.4.000-FHIR4-TestCases.zip](https://github.com/user-attachments/files/17077743/CMS144FHIR-v1.4.000-FHIR4-TestCases.zip)
4. [CMS145FHIR-v0.2.000-FHIR4-TestCases.zip](https://github.com/user-attachments/files/17077739/CMS145FHIR-v0.2.000-FHIR4-TestCases.zip)
[...]

Issue #285, opened 2023-10-17
Coverage Calculation Discounting Aliases and/or Union Operators


jkotanchik-SB:
### Summary
These measures (bundles attached) are supposed to have 100% coverage but are showing between 81% and 99% even though only a few aliases and/or union operators are unhighlighted:

1. **CMS 135 - Heart Failure (HF): Angiotensin-Converting Enzyme (ACE) Inhibitor or Angiotensin Receptor Blocker (ARB) or Angiotensin Receptor-Neprilysin Inhibitor (ARNI) Therapy for Left Ventricular Systolic Dysfunction (LVSD)FHIR** 
Coverage: 98%

2. **CMS 145 - Coronary Artery Disease (CAD): Beta-Blocker Therapy-Prior Myocardial Infarction (MI) or Left Ventricular Systolic Dysfunction (LVSD)FHIR**
Coverage: 99%

3. **CMS 159 - Depression Remission at Twelve MonthsFHIR**
Coverage: 81%

4. **CMS 645 - Bone density evaluation for patients with prostate cancer and receiving androgen deprivation therapyFHIR**
Coverage: 99%

5. **CMS 646 - Intravesical Bacillus-Calmette-Guerin for non-muscle invasive bladder cancerFHIR**
Coverage: 92%

### External Tracking Ticket
https://oncprojectracking.healthit.gov/support/browse/BONNIEMAT-1627

### Expected Behavior
100% coverage

### Version or Commit
v1.3.1

### Inputs (e.g. Measure Bundle, Patient Bundle, CQL Library)
Measure bundles:

1. [CMS135FHIR-v0.1.000-FHIR.json.zip](https://github.com/projecttacoma/fqm-execution/files/12950461/CMS135FHIR-v0.1.000-FHIR.json.zip)
2. [CMS145FHIR-v0.1.000-FHIR.json.zip](https://github.com/projecttacoma/fqm-execution/files/12950532/CMS145FHIR-v0.1.000-FHIR.json.zip)
3. [CMS159FHIR-v0.1.000-FHIR.json.zip](https://github.com/projecttacoma/fqm-execution/files/12950590/CMS159FHIR-v0.1.000-FHIR.json.zip)
4. [CMS645FHIR-v1.3.000-FHIR.json.zip](https://github.com/projecttacoma/fqm-execution/files/12950632/CMS645FHIR-v1.3.000-FHIR.json.zip)
[...]

PR #314, opened 2024-10-08
Coverage 3.14.0

lmd59:
# Summary
Implements an annotation-based coverage approach that only uses local ids that are also in the annotation (and therefore can be seen in the cql coverage highlighting). This mitigates updates in version 3.14.0 of the translator which added many additional local ids that could otherwise impact coverage.

## New behavior
Coverage calculations will be approximately the same as for translator 3.3.2. There may be some very small differences in in coverage percentages since this is a new coverage approach, but any coverage gaps should be identifiable for the end user.

## Code changes

- Small README updates
- `ClauseResultsHelpers.ts` Update to add `findAnnotationLocalIds` function that identifies all "r"-keyed local ids within an annotation structure. Uses this to limit the final localIds to the ones that can be found in the annotation.
- `ClauseResultsHelpers.test.ts` Updates to "Not Equivalent" and "Not Equal" and addition of a "Not Null" unit test to align with new localid approach
- Adds `3.15.0` folder for new translations of existing easily translatable cql unit test fixtures (some existing cql fixtures do not have the right dependency library versions to be easily re-translated)

# Testing guidance
- Test with the 3.14.0 measures provided in https://github.com/projecttacoma/fqm-execution/issues/313 and confirm coverage percentages are restored. Maybe test with other measures with known past coverage issues.

### DIFF
--- a/src/helpers/ClauseResultsHelpers.ts
+++ b/src/helpers/ClauseResultsHelpers.ts
@@ -68,13 +68,13 @@ export function findAllLocalIdsInStatementByName(libraryElm: ELM, statementName:
  * @param {object} annotation - all or a subset of the annotation structure to search
  * @return {Array} List of local ids in the annotation.
  */
-function findAnnotationLocalIds(annotation: any): any[] {
+function findAnnotationLocalIds(annotation: any): string[] {
   if (Array.isArray(annotation)) {
     return annotation.flatMap(ent => findAnnotationLocalIds(ent));
   } else if (typeof annotation === 'object') {
     return Object.entries(annotation).flatMap(ent => {
       // if key is r, return value, else recurse
-      if (ent[0] === 'r') return ent[1];
+      if (ent[0] === 'r') return ent[1] as string;
       return findAnnotationLocalIds(ent[1]);
     });
   }
--- a/src/helpers/ClauseResultsHelpers.ts
+++ b/src/helpers/ClauseResultsHelpers.ts
@@ -3,7 +3,7 @@ import { ELMFunctionRef } from '../types/ELMTypes';
 import { ELM, ELMBinaryExpression, ELMStatement } from '../types/ELMTypes';
 
 /**
- * Finds all localIds in a statement by it's library and statement name.
+ * Finds all localIds in a statement by its library and statement name.
  * @public
  * @param {ELM} libraryElm - The library the statement belongs to.
  * @param {string} statementName - The statement name to search for.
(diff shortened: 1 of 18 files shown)


src/helpers/ClauseResultsHelpers.ts
```diff
@@ -49,7 +49,37 @@ export function findAllLocalIdsInStatementByName(libraryElm: ELM, statementName:
...
+  return annotatedLocalIds;
+}
+
+/**
+ * Recursively finds just localIds that are in an annotation structure by pulling out all "r:"-keyed values
+ * @public
+ * @param {object} annotation - all or a subset of the annotation structure to search
+ * @return {Array} List of local ids in the annotation.
+ */
+function findAnnotationLocalIds(annotation: any): any[] {
```
hossenlopp, line 71: This is pretty clever even if a bit hard to read. I think it could be specific about the return type being `string[]` though. I realize the existing code around this doesn't have proper types, but I think this new code could start to use something other than `any`.

hossenlopp (commented):
Small thought on typescript types.

src/helpers/ClauseResultsHelpers.ts
lmd59: Legit- I'll give it a shot!

hossenlopp (approved):

elsaperelli (commented):
I am not sure how much we care about this, but we should at the very least note it in the release notes. For CMS646, I noticed that the highlighting for v3.3.2 and v3.14.0 are slightly different. While the percentage is the same, looks like some aliases no longer get highlighted. Could be worth investigating if there is a fix and if not, at least noting it.

lmd59:
> I am not sure how much we care about this, but we should at the very least note it in the release notes. For CMS646, I noticed that the highlighting for v3.3.2 and v3.14.0 are slightly different. While the percentage is the same, looks like some aliases no longer get highlighted. Could be worth investigating if there is a fix and if not, at least noting it.
> 
> For example: image

So it looks like you're seeing a change that was caused by the translator update and not by this PR as 3.3.2-translated measure results before and after this PR change have the same highlighting (alias highlighted), and 3.14.0-translated measure results before and after this PR change have the same highlighting (alias  un-highlighted).
Since this PR's task aims to resolve coverage and address the linked ticket, addressing highlighting may be out of scope (though the ticket does reference https://github.com/projecttacoma/fqm-execution/issues/285 as well, which is a highlighting issue).
I think we may also want to address highlighting separately because the fix is more likely to break backwards compatibility. Essentially, we have workarounds for highlighting aliases that use a plus 1 increment to identify the correct localid association (since aliases are identified within the annotation but not within the ELM). Initial testing shows that the relationships between the correct localids are at least sometimes a +2. Especially if we are aiming for backwards compatibility, this may take a new approach.

test/unit/ClauseResultsHelper.test.ts
```diff
@@ -28,10 +28,24 @@ describe('ClauseResultsHelpers', () => {
...
-      // the result of the Not expression
-      expect(localIds[100]).toBeDefined();
-      expect(localIds[100]).toEqual({ localId: '100', sourceLocalId: '23' });
+      // is 100, and the localId for the Not expression is 23. We want 100 to not be included because 23 is
+      // the correct id from the annotation.
+      expect(localIds[23]).toBeDefined();
+      expect(localIds[100]).not.toBeDefined();
+    });
+
+    test('finds localIds for an not null operator', () => {
```
elsaperelli, line 37:
```suggestion
    test('finds localIds for a not null operator', () => {
```

elsaperelli (changes requested):
Just a few small comments but functionality works great! Definitely want to wait to get the correct version of CMS145 to test.

src/helpers/ClauseResultsHelpers.ts
```diff
@@ -3,7 +3,7 @@ import { ELMFunctionRef } from '../types/ELMTypes';
 import { ELM, ELMBinaryExpression, ELMStatement } from '../types/ELMTypes';
 
 /**
- * Finds all localIds in a statement by it's library and statement name.
+ * Finds all localIds in a statement by its library and statement name.
```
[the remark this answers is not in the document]
lmd59: Actually, let's change it to `findLocalIdsInStatementByName` too to help clear it up a bit.

src/helpers/ClauseResultsHelpers.ts
```diff
@@ -49,7 +49,37 @@ export function findAllLocalIdsInStatementByName(libraryElm: ELM, statementName:
       clause.isUnsupported = true;
     }
   }
-  return localIds;
+
+  // find all localids in the annotation
+  const allAnnotatedIds = findAnnotationLocalIds(statement?.annotation);
+  // filter out local ids that aren't in the annotation
+  const annotatedLocalIds: { [key: string]: any } = {};
```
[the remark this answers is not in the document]
lmd59: I like that- good call!

elsaperelli (approved):
Did a lot of testing on Friday (separate from the task that we have to do additional testing on this) and didn't find any inconsistencies. 

Lgtm! 🥳

hossenlopp (approved):

Issue #311, opened 2024-09-13
UnexpectedProperty: Stratum strata-2 in group 


JSRankins:
### Summary
In using the global CLI (v 1.5.0), ran into an issue with the associated measure and patient bundles for the reports method using something similar to following command:

`fqm-execution reports -m "path\to\measures\EpisodeMeasure.json" -p "path\to\patients\FMUSC_Patient.json" -o "path\to\output\Individal_Report.json" --report-type individual --vs-api-key {apiKey}`

The following error was received:
```
UnexpectedProperty: Stratum strata-2 in group 66dfc5b649447a49beca9124 not found in measure reports
    at C:\Users\srankins\AppData\Roaming\npm\node_modules\fqm-execution\build\calculation\MeasureReportBuilder.js:205:43
    at Array.forEach (<anonymous>)
    at C:\Users\srankins\AppData\Roaming\npm\node_modules\fqm-execution\build\calculation\MeasureReportBuilder.js:186:93
    at Array.forEach (<anonymous>)
    at C:\Users\srankins\AppData\Roaming\npm\node_modules\fqm-execution\build\calculation\MeasureReportBuilder.js:173:45
    at Array.forEach (<anonymous>)
    at MeasureReportBuilder.addPatientResults (C:\Users\srankins\AppData\Roaming\npm\node_modules\fqm-execution\build\calculation\MeasureReportBuilder.js:148:32)
    at C:\Users\srankins\AppData\Roaming\npm\node_modules\fqm-execution\build\calculation\MeasureReportBuilder.js:586:21
    at Array.forEach (<anonymous>)
    at MeasureReportBuilder.buildMeasureReports (C:\Users\srankins\AppData\Roaming\npm\node_modules\fqm-execution\build\calculation\MeasureReportBuilder.js:583:26) {
  debugOutput: undefined
}
```

The details method appears to work for the measure and patient bundles (response attached: [Episode_Details.json](https://github.com/user-attachments/files/16999600/Episode_Details.json)):
`fqm-execution details -m "path\to\measures\EpisodeMeasure.json" -p "path\to\patients\Episode_Patient.json" -o "path\to\output\Episode_Details.json" --vs-api-key {apiKey}`

### Expected Behavior
Calculation and generation of individual report with the relevant results

### Version or Commit
1.5.0

### Inputs (e.g. Measure Bundle, Patient Bundle, CQL Library)
Measure Bundle: 
[...]

PR #315, opened 2024-10-10
Stratified report fix

lmd59:
# Summary
Addresses UnexpectedProperty issue in https://github.com/projecttacoma/fqm-execution/issues/311. 

## New behavior
For episode-based measures with stratifiers, (as was already done for patient-based), we now default the `strataCode` to be the id if the `strata.code?.text` does not exist

## Code changes

- `DetailedResultsBuilder.ts` - defaults `strataCode` to `strata.id` as secondary preferred value for episode results
- `MeasureReportBuilder.ts` - slight efficiency improvement and added clarifying comment
- Adds tests addressing episode+stratifier cases in `DetailedResultsBuilder.test.ts` an `MeasureReportBuilder.test.ts`

# Testing guidance

- `npm run check`
- Pull measure/patient from ticket and test using `fqm-execution details -m "path\to\measures\EpisodeMeasure.json" -p "path\to\patients\Episode_Patient.json" -o "path\to\output\Episode_Details.json" --vs-api-key {apiKey}`


src/calculation/MeasureReportBuilder.ts
```diff
@@ -647,9 +647,9 @@ export default class MeasureReportBuilder<T extends PopulationGroupResult> exten
     options: CalculationOptions
   ): fhir4.MeasureReport[] {
     const reports: fhir4.MeasureReport[] = [];
+    const measure = extractMeasureFromBundle(measureBundle);
+    const builder = new MeasureReportBuilder(measure, options);
```
hossenlopp, line 651: This line needs to be inside the `forEach` loop below as it needs to be creating a new MeasureReport for each patient.

Good call to move the `extractMeasureFromBundle` outside the loop.

hossenlopp (changes requested):

src/calculation/MeasureReportBuilder.ts
lmd59: whoops! Thank you for the catch!

test/unit/MeasureReportBuilder.test.ts
```diff
@@ -490,6 +606,49 @@ describe('MeasureReportBuilder Static', () => {
...
+    let measureReports: fhir4.MeasureReport[];
+    beforeAll(() => {
+      measureReports = MeasureReportBuilder.buildMeasureReports(
+        episodeWithStratMeasureBundle,
+        episodeWithStratExecutionResults,
+        calculationOptions
+      );
+    });
+
+    test('should generate 1 result', () => {
```
elsaperelli, line 619: Maybe be a bit more specific here like "Generates one measure report"

elsaperelli (approved):
🥳

PR #317, opened 2024-10-28
Relationship alias coverage fixes

hossenlopp:
# Summary

Fixes highlighting of aliases in with/without relationships.

## New behavior

Looks in the annotation for the localId that includes the relationship alias so it can be highlighted to match the relationship source.

Also uses the uncoverage work to properly 'not' change the highlighting in coverage when there is a node of the annotation that does not have a result due to its localId not existing in the logic.

## Code changes

- `ClauseResultsHelpers.ts` - Added new functionality to find the localId in the annotation that contains the source and alias of relationship clauses. This is then assigned as the alias for the source of the relationship so the alias can be highlighted as covered when the source is covered.
- `HTMLBuilder.ts` - Now using the uniquified clauseReuslts results from uncoverage to highlight coverage and allow for annotation nodes that we do not care about (i.e. the ones with localId's only in annotations which we cannot account for in expressions) to not have any styling so they can inherit their parent styling.

# Testing guidance

Run unit tests and regression scripts to look for changes in coverage. Some changes in coverage and highlighting are to be expected.


src/helpers/ClauseResultsHelpers.ts
```diff
@@ -132,11 +132,16 @@ export function findAllLocalIdsInStatement(
         aliasMap[v] = statement.expression.localId;
         // Determine the localId for this alias.
         if (statement.localId) {
-          // Older translator versions require with statements to use the statement.expression.localId + 1 as the alias Id
-          // even if the statement already has a localId. There is not a clear mapping for alias with statements in the new
-          // translator, so they will go un highlighted but this will not affect coverage calculation
+          // There is not a clear mapping for `With` and `Without` relationship alias with statements in newer
+          // translator versions. The node in the annotation that contains the alias does has a local id that doesn't
```
lmd59, line 136:
```suggestion
          // translator versions. The node in the annotation that contains the alias has a local id that doesn't
```

src/helpers/ClauseResultsHelpers.ts
```diff
@@ -132,11 +132,16 @@ export function findAllLocalIdsInStatement(
         aliasMap[v] = statement.expression.localId;
         // Determine the localId for this alias.
         if (statement.localId) {
-          // Older translator versions require with statements to use the statement.expression.localId + 1 as the alias Id
-          // even if the statement already has a localId. There is not a clear mapping for alias with statements in the new
-          // translator, so they will go un highlighted but this will not affect coverage calculation
+          // There is not a clear mapping for `With` and `Without` relationship alias with statements in newer
```
lmd59, line 135:
```suggestion
          // There is not a clear mapping for `With` and `Without` relationship aliases with statements in newer
```

src/helpers/ClauseResultsHelpers.ts
```diff
@@ -132,11 +132,16 @@ export function findAllLocalIdsInStatement(
...
+          // translator versions. The node in the annotation that contains the alias does has a local id that doesn't
+          // exist in the elm. We have to find this localId by looking for it in the annotation structure.
           if (statement.type === 'With' || statement.type === 'Without') {
-            alId = (parseInt(statement.expression.localId, 10) + 1).toString();
+            if (annotation) {
```
lmd59, line 139: Since we've added a nested `if` here, we should probably make sure we have a default value for `alId` when there isn't an annotation or when we otherwise can't find the relationship alias, otherwise `alId` will be undefined. On line 36 of this file, we have a check against null, but we don't seem to have a check against undefined.

src/helpers/ClauseResultsHelpers.ts
```diff
@@ -347,6 +352,78 @@ export function findLocalIdsForComparisonOperators(
   }
 }
 
+/**
+ * Helper function to kick off the recursive search for the relationship (With, Without) source's localId in annotation
```
lmd59, line 356:
```suggestion
 * Helper function to kick off the recursive search for the relationship (With, Without) source's localId in the annotation
```

src/helpers/ClauseResultsHelpers.ts
```diff
@@ -347,6 +352,78 @@ export function findLocalIdsForComparisonOperators(
...
+    const id = findRelationshipAliasNodeAnnotationId([a.s], sourceLocalId);
+    if (id) {
+      return id;
+    }
+  }
+  return null;
+}
+
+/**
+ * Recursively looks through the annotation structure for the source localId and grab its parent id which also contains
```
lmd59, line 371:
```suggestion
 * Recursively looks through the annotation structure for the source localId and grabs its parent id which also contains
```

src/helpers/ClauseResultsHelpers.ts
```diff
@@ -347,6 +352,78 @@ export function findLocalIdsForComparisonOperators(
...
+ * the alias. This search is valid to be used with relationship (With, Without) clauses and  allows us to tag the found
```
lmd59, line 372:
```suggestion
 * the alias. This search is valid to be used with relationship (With, Without) clauses and allows us to tag the found
```

src/helpers/ClauseResultsHelpers.ts
```diff
@@ -347,6 +352,78 @@ export function findLocalIdsForComparisonOperators(
...
+ *    ]
+ *  }
+ *
+ * @returns id of the node that is parent to the source localId if found
+ */
+function findRelationshipAliasNodeAnnotationId(
+  annotation: AnnotationStatement[],
+  sourceLocalId: string
+): string | null {
+  for (const as of annotation) {
```
lmd59, line 411: Suggest a different iterator variable than `as` which may be confusing since "as" can be a key word in some contexts.

lmd59 (commented):
Love that this fix is working out for us! I added some little comment cleanup suggestions so that you can easily batch most of them, and then a couple of other small suggestions.

We still have a couple of other `- 1` and `+ 1`'s in the ClauseResultsHelpers file. Do we know if we will have any similar issues with these other alias id's? Might be out of scope for this task, but wanted to bring it up in case we need to create a follow on task.

src/helpers/ClauseResultsHelpers.ts
lmd59: Recommended change: Update check on line 36 to also encompass the undefined case.

lmd59 (approved):
Nice updates!

elsaperelli (approved):
Lgtm!

PR #321, opened 2024-11-07
Regression cql

lmd59:
# Summary
Updates regression scripts to do an npm install before running so that cql-execution (and other dependencies) can be changed between runs based on the package file.
Adds scripting for pulling from the gitlab coverage-script-bundles and test regression against those formats of packages.

## New behavior
No changes to fqm-execution functionality

## Code changes

# Testing guidance
Run the regression tests using `./regression/run-regression.sh` on this branch (regression tests should pass). Create a local branch off of this branch and commit any changes, including a change to the cql-execution version (`"cql-execution": "3.0.1",`), then run regression scripts again, checking for expected output.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ cache
 test/integration/.start-translator
 regression/output/*-*
 regression/connectathon
+regression/coverage-script-bundles
 regression/ecqm-content-r4-2021
 regression/ecqm-content-qicore-2022
 data-requirements/fqm-e-dr/*
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
         "axios": "^0.21.1",
         "commander": "^6.1.0",
         "cql-exec-fhir": "^2.1.3",
-        "cql-execution": "3.0.1",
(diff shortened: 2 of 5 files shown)


cmoesel (commented):
Thank you, @lmd59! No one asked me to review this, but I needed to run a regression anyway, so I took it for a spin.

Everything seems to be working well! The addition of the coverage script bundles is a nice feature, as it appears to add quite a few additional test cases. Excellent!

The one thing I would suggest you consider is making it _optional_ to include the coverage script bundles (maybe enabled by a command line flag). The reasons:
1. The coverage script bundles take a long time. Sometimes a quick regression might be preferred.
2. Anyone using fqm-execution outside of MITRE will not have access to the MITRE GitLab -- so the coverage script bundles will never work for them.

Thanks for this. It will be helpful in the future for feeling more confident about changes to cql-execution.

cmoesel (approved):
Looks good to me! Thank you, @lmd59! That said, I'm new here and wasn't around for discussions about this feature -- so you probably will want a review and approval from someone on the team as well.

PR #322, opened 2024-12-06
Fix handling of function without library for Equal

lmd59:
# Summary
CMS334 https://github.com/cqframework/ecqm-content-qicore-2024/tree/main/bundles/measure/CesareanBirthFHIR has an `ELMFunctionRef` without a libraryName that gets hit by `interpretEqual`. This makes `interpretFunctionRef` return an undefined which makes `isOfTypeGracefulError` choke.

I would prefer to fix this by making `interpretFunctionRef` return definitive types, but it looks like everywhere else (and now in `interpretEqual`) we do a check for undefined via a lovely `== null` check after it's called. So that's what we're doing here for now, but it would be nice to clean some of this up further in the future... should we add a ticket?
This PR also makes `isOfTypeGracefulError` a bit more robust.

## New behavior
No longer breaks on CMS334.

## Code changes

- See summary

# Testing guidance

- npm run check
- npm run cli -- dataRequirements -m {path2CMS334-VS.json} (I added valuesets to the measure locally, but you can also use the command line option to fetch valuesets during calculation)

### DIFF
--- a/src/helpers/elm/QueryFilterParser.ts
+++ b/src/helpers/elm/QueryFilterParser.ts
@@ -720,6 +720,11 @@ export function interpretEqual(equal: ELMEqual, library: ELM): EqualsFilter | Un
     propRef = equal.operand[0] as ELMProperty;
   }
 
+  if (propRef == null) {
+    withError.message = `could not resolve property ref for Equal:${equal.localId}. first operand is a ${equal.operand[0].type}`;
+    return { type: 'unknown', withError };
+  }
+
   if (isOfTypeGracefulError(propRef)) {
     return { type: 'unknown', withError: propRef };
   }
--- a/src/types/errors/GracefulError.ts
+++ b/src/types/errors/GracefulError.ts
@@ -6,7 +6,7 @@ export interface GracefulError {
 }
 
 export const isOfTypeGracefulError = (tbd: any): tbd is GracefulError => {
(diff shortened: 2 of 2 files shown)


elsaperelli (approved):
I think these changes make sense for now in order to successfully run CMS334. I think making a ticket to clean up the any return type and the `propRef == null` statements is a good idea.

Looks like it just needs a rebase!

hossenlopp (changes requested):
This is a partial fix. The actual issue that causes the crash is in the `interpretFunctionRef` function. It could use an `else` part to handle the situation when it is a reference to a function in the same library. It should be returning a `GracefulError` like it does when the libraryName does exist and can't be determined what it does

lmd59:
> This is a partial fix. The actual issue that causes the crash is in the `interpretFunctionRef` function. It could use an `else` part to handle the situation when it is a reference to a function in the same library. It should be returning a `GracefulError` like it does when the libraryName does exist and can't be determined what it does

Yeah- that was what I was trying to say about `interpretFunctionRef` in the summary above. But this `== null` check is the way that we handle it for every other interpret* function. I think we should be consistent for now but make a ticket to overhaul `interpretFunctionRef` and the interpret* functions that call it so they're *all* handling it correctly (Note: this will probably involve a fair bit of unit test updates as well)

hossenlopp (approved):

PR #334, opened 2025-05-02
Gaps groupscore

lmd59:
# Summary
A [cqf computable measure](https://hl7.org/fhir/us/cqfmeasures/StructureDefinition-computable-measure-cqfm.html) can define the measure scoring and improvement notation on a specific group rather than a measure as a whole. This PR updates gaps in care to be able to pull this information from the group as a primary source, with the measure as a backup source.

## New behavior
Does not error when scoring and group notation are on the group definition rather than the top measure definition.

## Code changes
- Updates to MeasureBundleHelper (and tests) to provide a helper function for getting the group improvement notation
- Updates to the `calculateGapsInCare` function in `Calculator.ts` to identify the current group earlier, and use that as the primary source for the scoring and improvement notation with a fallback to the measure's main scoring and improvement notation. Throws an error if neither are available.

# Testing guidance
Note: Coverage script bundles will still fail to calculate for gaps because they have no improvement notation (at least the ones I spot checked). Improvement notation is 0..1, but I could not find anything in the spec about what the assumed value should be, so I believe we should continue to reject measure bundles without an improvement notation.
Latest bundles, on the other hand, appear to appropriately have a group improvement notation, so I have updated the code to handle the group-based improvement notation, and I will provide a test bundle in slack.

- `npm run check`
- `npm run cli -- gaps -m [path2]/CMS130FHIR-v0.3.000-FHIR-with-vs.json -p [path2]/CMS130FHIR-v0.3.000-FHIR-with-vs-test-cases/DENOM\ Patient-denom-EXM130.json`
- Should return a guidance response (which should be interesting since this is a denominator patient).

### DIFF
--- a/src/calculation/Calculator.ts
+++ b/src/calculation/Calculator.ts
@@ -519,7 +519,7 @@ export async function calculateGapsInCare<T extends OneOrMultiPatient>(
         measureResource.improvementNotation?.coding?.[0].code;
 
       if (!improvementNotation) {
-        throw new UnexpectedProperty('Could not find improvement notation code for this group or measure');
+        throw new UnexpectedProperty('Unable to find improvement notation code for this group or measure');
       }
 
       // If positive improvement measure, consider patients in denominator but not numerator for gaps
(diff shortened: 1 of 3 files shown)


elsaperelli (approved):
Tested with provided CMS130 bundle and test case and CMS108 in the coverage script bundles. Tested coverage regression, regression, integration, and unit tests. Lgtm!

hossenlopp (approved):

Issue #337, opened 2025-05-13
Numerator Exclusion result is returned as True when Numerator itself is False


RohitKandimalla:
### Summary

Attached the QiCore V6 Measure Bundle and the Test case.

In the CQL, Population Result of Numerator is returned as False which is excpeted
{
    "populationType": "numerator",
    "criteriaExpression": "Numerator",
    "result": false,
    "populationId": "340EA45E-2411-4192-9C9D-3DF8D89A1D97"
}
, but Numerator Exception is returning True.
{
    "populationType": "numerator-exclusion",
    "criteriaExpression": "Numerator Exclusions",
    "result": true,
    "populationId": "a74233ca-247b-4beb-949c-47db0bc2bcaa"
}

CQLIT team verified the CQL and they see no issue with it, Although they said "The DENEX and NUMEX are defined exactly the same so not surprising that it would count for both populations but it is up to the evaluation engine to know not to count the NUMEX if NUMER is 0."

So the question is: when the numerator is false, should the Numerator Exception even be triggered?

#### External Tracking Ticket
https://oncprojectracking.healthit.gov/support/browse/MADIE-2210

### Expected Behavior
Expecting Numerator Exception to be False
Note: Population relevance of Numex is false.

### Version or Commit
v4.3.1

### Inputs (e.g. Measure Bundle, Patient Bundle, CQL Library)

Measure Bundle
[CMS871FHIR-v0.2.001-FHIR6 (1).zip](https://github.com/user-attachments/files/20191477/CMS871FHIR-v0.2.001-FHIR6.1.zip)

Test Cases
[CMS871FHIR-v0.2.001-FHIR6-TestCases.zip](https://github.com/user-attachments/files/20191478/CMS871FHIR-v0.2.001-FHIR6-TestCases.zip)

### Relevant Calculation Options (e.g. Measurement Period, meta.profile Validation)
MP : 1/1/2026 - 12/31/2026

### cql-to-elm Version Used for Measure Logic Translation (if known)
v3.22.0

lmd59:
Thank you for providing the test data. We believe this is an issue with ratio measure population exclusion in particular. We are looking into addressing this soon.

PR #339, opened 2025-05-21
Fix Numex for ratio measures

lmd59:
# Summary
Reorganizes population nulling logic by measure score calculation type to provide better readability and fix issue https://github.com/projecttacoma/fqm-execution/issues/337

## New behavior
When a ratio measure has a false numerator, its numex (numerator exclusions) should be false as well.

## Code changes

- In `DetailedResultsBuilder.ts`, update comments with logic source references and addition descriptions, reorganize population nulling logic by measure score type with a different function for each score type. Reduce out unnecessary logic and short cut logic as able based on the score type. This reorganization lets us especially makes sure `// Cannot be in the NUMEX if not in the NUMER` is enforced for all cases of relevance for ratio measures.
- Add convenience method `isCVMeasure` in `MeasureBundleHelpers.ts`
- Add and update tests in `DetailedResultsBuilder.test.ts`

# Testing guidance

- `npm run check`
-  With the test information provided in the issue linked above, `npm run cli -- detailed -m {path to CMS871FHIR-v0.2.001-FHIR.json} --patients-directory {path to CMS871FHIR-v0.2.001-FHIR6-TestCases } -o --debug --vs-api-key {api key}` - all `numerator-exclusion` in output should be false
- Run `./regression/run-regression.sh` to check regression tests!

### DIFF
--- a/src/calculation/DetailedResultsBuilder.ts
+++ b/src/calculation/DetailedResultsBuilder.ts
@@ -244,9 +244,9 @@ export function handleCVPopulationValues(populationResults: PopulationResult[]):
     // Short-circuit return since no more processing needs to be done if IPP is false with only one IPP
     return populationResultsHandled;
   }
-  // Cannot be in most populations if not in DENOM or MSRPOPL
+  // Cannot be in most populations if not in MSRPOPL
   if (hasResult(PopulationType.MSRPOPL, populationResults) && !getResult(PopulationType.MSRPOPL, populationResults)) {
-    // If there is a MSRPOPL, all observations point to it, so null them out
+    // If a result exists for the MSRPOPL, all observations point to it, so null them out
     const popResult = populationResults.find(result => result.populationType === PopulationType.OBSERV);
     if (popResult) {
       popResult.result = false;
--- a/src/calculation/DetailedResultsBuilder.ts
+++ b/src/calculation/DetailedResultsBuilder.ts
@@ -133,6 +133,8 @@ export function handlePopulationValues(
   measureScoringCode?: string
 ): PopulationResult[] {
   /* Population logic guidance: https://mmshub.cms.gov/sites/default/files/Measure-Calculations.pdf
+   * Follows population requirements in table 3-1 here: https://hl7.org/fhir/us/cqfmeasures/measure-conformance.html#data-criteria
+   *
    * For proportional: Setting values of populations if the correct populations are not set based on the following logic guidelines
    * Initial Population (IPP): The set of patients or episodes of care to be evaluated by the measure.
    * Denominator (DENOM): A subset of the IPP.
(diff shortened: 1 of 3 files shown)


src/calculation/DetailedResultsBuilder.ts
```diff
@@ -178,44 +210,49 @@ export function handlePopulationValues(
...
+        result.observations = null;
+      }
+      result.result = false;
+    });
+    // Short-circuit return since no more processing needs to be done if IPP is false with only one IPP
+    return populationResultsHandled;
+  }
+  // Cannot be in most populations if not in DENOM or MSRPOPL
+  if (hasResult(PopulationType.MSRPOPL, populationResults) && !getResult(PopulationType.MSRPOPL, populationResults)) {
+    // If there is a MSRPOPL, all observations point to it, so null them out
```
elsaperelli, line 249: I am a little confused about this logic, could you maybe walk me through it?

To my understanding, this is looking if the MSRPOPL has a result (result is not null) AND that that result is false ?

src/calculation/DetailedResultsBuilder.ts
```diff
@@ -224,6 +261,48 @@ export function handlePopulationValues(
...
+}
+
+export function handleStandardPopulationValues(
+  populationResults: PopulationResult[],
+  group: fhir4.MeasureGroup
+): PopulationResult[] {
+  // proportional/cohort handling
+  const populationResultsHandled = populationResults;
+  // Cannot be in all populations if not in IPP.
+  if (!getResult(PopulationType.IPP, populationResults)) {
```
elsaperelli, line 275: Yeah having a hard time with this. So if getResult() returns true, then the ! makes that false, right?

elsaperelli (changes requested):
Just a few suggestions and some questions about some of the logic I would love to discuss! Otherwise functionality looks good!

src/calculation/DetailedResultsBuilder.ts
```diff
@@ -141,9 +144,29 @@ export function handlePopulationValues(
    * Numerator (NUMER): A subset of the Denominator. The Numerator criteria are the processes or outcomes expected for each patient,
    * procedure, or other unit of measurement defined in the Denominator.
    * Numerator Exclusions (NUMEX): A subset of the Numerator that should not be considered for calculation.
-   * Measure Population Exclusions (MSRPOPLEX): Identify that subset of the MSRPOPL that meet the MSRPOPLEX criteria.
+   *
+   * For Ratio: the NUMER and DENOM are independent, and there may be multiple IPPs.
+   *
+   * For CV:
+   * Measure Population Exclusions (MSRPOPL): A subset of the IPP for continuous variable measures.
```
lmd59, line 151:
```suggestion
   * Measure Population (MSRPOPL): A subset of the IPP for continuous variable measures.
```

src/calculation/DetailedResultsBuilder.ts
```diff
@@ -141,9 +144,29 @@ export function handlePopulationValues(
...
+   * Measure Population Exclusions (MSRPOPLEX): A subset of the MSRPOPL that should not be considered for calculation.
    */
+
+  if (MeasureBundleHelpers.isRatioMeasure(group, measureScoringCode)) {
+    return handleRatioPopulationValues(populationResults, group);
+  } else if (MeasureBundleHelpers.isCVMeasure(group, measureScoringCode)) {
+    return handleCVPopulationValues(populationResults);
+  } else {
+    return handleStandardPopulationValues(populationResults, group);
```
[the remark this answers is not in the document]
lmd59: It's proportion and cohort... and maybe composite would end up getting to this logic branch too, but I don't think we would let it get this far if it were composite.

src/calculation/DetailedResultsBuilder.ts
lmd59: Yes- I believe so. Happy to look at it together.

src/calculation/DetailedResultsBuilder.ts
lmd59: Yes- this checks if the IPP result is false

src/calculation/DetailedResultsBuilder.ts
```diff
@@ -178,44 +210,49 @@ export function handlePopulationValues(
...
+    populationResults.forEach(result => {
+      if (result.populationType === PopulationType.OBSERV) {
+        result.observations = null;
+      }
+      result.result = false;
+    });
+    // Short-circuit return since no more processing needs to be done if IPP is false with only one IPP
+    return populationResultsHandled;
+  }
+  // Cannot be in most populations if not in DENOM or MSRPOPL
```
lmd59, line 247:
```suggestion
  // Cannot be in most populations if not in MSRPOPL
```

elsaperelli (approved):
🥳

PR #363, opened 2026-03-17
Default value constants

elsaperelli:
# Summary
This PR uses constants for default values in the code as suggested by @cmoesel in his [fqm-analysis](https://github.com/projecttacoma/fqm-execution/commit/a89e1d208d0527e4813bfc4420fd09542c017902).

## New behavior
There should be no new behavior.

## Code changes
- `src/calculation/Calculator.ts` - use default constants from constants.ts for calculation option defaults
- `src/calculation/CompositeReportBuilder.ts` - use constant for 'UnknownMeasure' default value
- `src/calculation/MeasureReportBuilder.ts` - use constant for isIndividual true default value
- `src/cli.ts` - use constant for 'output.json' default value
- `src/constants.ts` - file for constants
- `src/execution/Execution.ts` / `src/helpers/MeasureBundleHelpers.ts` - use constants for default measurement period start and end date strings

# Testing guidance
- Look through the project to see if there are any other default values that could be defined as constants for clarity.
- `npm run check` sanity check
- `npm run test:integration` sanity check

### DIFF
--- a/src/calculation/Calculator.ts
+++ b/src/calculation/Calculator.ts
@@ -42,6 +42,18 @@ import { clearElmInfoCache } from '../helpers/elm/ELMInfoCache';
 import _, { omit } from 'lodash';
 import { ELM } from '../types/ELMTypes';
 import { getReportBuilder } from '../helpers/ReportBuilderFactory';
+import {
+  DEFAULT_BUILD_STATEMENT_LEVEL_HTML,
+  DEFAULT_CALCULATE_CLAUSE_COVERAGE,
+  DEFAULT_CALCULATE_CLAUSE_UNCOVERAGE,
+  DEFAULT_CALCULATE_COVERAGE_DETAILS,
+  DEFAULT_CALCULATE_HTML,
+  DEFAULT_CALCULATE_RAVS,
+  DEFAULT_CALCULATE_SDES,
+  DEFAULT_DISABLE_HTML_ORDERING,
+  DEFAULT_TRUST_META_PROFILE,
+  DEFAULT_VERBOSE_CALCULATION_RESULTS
+} from '../constants';
 
 /**
  * Calculate measure against a set of patients. Returning detailed results for each patient and population group.
@@ -59,7 +71,7 @@ export async function calculate<T extends CalculationOptions>(
   valueSetCache: fhir4.ValueSet[] = []
 ): Promise<CalculationOutput<T>> {
   // Ensure verboseCalculationResults defaults to true when not provided in options
-  options.verboseCalculationResults = options.verboseCalculationResults ?? true;
+  options.verboseCalculationResults = options.verboseCalculationResults ?? DEFAULT_VERBOSE_CALCULATION_RESULTS;
 
   const debugObject: DebugOutput | undefined = options.enableDebugOutput ? <DebugOutput>{} : undefined;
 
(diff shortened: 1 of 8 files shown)


src/calculation/MeasureReportBuilder.ts
```diff
@@ -41,7 +42,7 @@ export default class MeasureReportBuilder<T extends PopulationGroupResult> exten
     if (this.options.reportType) {
       this.isIndividual = this.options.reportType === 'individual';
     } else {
-      this.isIndividual = true;
+      this.isIndividual = DEFAULT_IS_INDIVIDUAL_REPORT;
```
hossenlopp: I think this one is not necessary. The code here was pretty clear that it defaults to `true` and no longer is clear thats the case.

hossenlopp (changes requested):

src/calculation/MeasureReportBuilder.ts
lmd59: Are you saying to remove this constant entirely?

elsaperelli: Oh I assumed that- removed in [Revert one of the default constants](https://github.com/projecttacoma/fqm-execution/pull/363/commits/7a109511110bc21c49e36026c9c4d216c6d7c3c6). Happy to change back though if we want to discuss further.

lmd59: Thanks! When I looked into it more, I realized it only made sense if it was suggesting moving the constant entirely 👍

hossenlopp: Sorry about not being clear earlier. This looks good now.

hossenlopp (approved):

src/cli.ts
lmd59: const for DEFAULT_REPORT_TYPE here?

src/helpers/MeasureBundleHelpers.ts
lmd59: DEFAULT_CQL_NAME for 'unknown library' here?

lmd59 (commented):
Not sure the boundary for what we're setting a default constant for right now and what we're not, but noted a couple of other things in line, and also `DetailedResultsBuilder`, line 113 has default groupId 'unknown'

We also have a lot of hard-coded ELM types (i.e. 'FunctionRef' or 'ExpressionRef'), but those are likely better addressed with improved ELM typing.

src/calculation/Calculator.ts
lmd59: Will we ever hit these two falses since we default to true above?

lmd59 (commented):
One more thing I noticed.

src/cli.ts
elsaperelli: Sounds good! Addressed in [Add default for reportType](https://github.com/projecttacoma/fqm-execution/pull/363/commits/a634a129051b5bd1b4da765d72406cb63b1cf191)

src/helpers/MeasureBundleHelpers.ts
elsaperelli: Addressed in [Add default for unknown library](https://github.com/projecttacoma/fqm-execution/pull/363/commits/38b323b8948e9106fbf94d82f34a1dfb09257881)

src/calculation/Calculator.ts
elsaperelli: I think you are right but the typing here is still that the values are boolean | undefined so we may want to keep this as is? Happy to discuss more.

hossenlopp (approved):

src/calculation/Calculator.ts
lmd59: I would either add `as boolean` on these or have them default to the existing constants (i.e. `?? DEFAULT_CALCULATE_SDES`). Even though it won't be hit, the additional false backup option here is a bit confusing especially since the constant defaults are true.

lmd59 (commented):
Any thoughts on doing a constant for DetailedResultsBuilder, line 113 with default groupId 'unknown'?

src/calculation/Calculator.ts
elsaperelli: Addressed in [Constant for unknown group id, default for calulate SDEs and RAV](https://github.com/projecttacoma/fqm-execution/pull/363/commits/ce5c0e6d54dd82225d07ea2555676c68e615e73b)

lmd59 (approved):
Yay constants!!

hossenlopp (approved):

PR #365, opened 2026-05-14
Update TS, ESLint, Prettier

hossenlopp:
# Summary

Update TypeScript, ESLint, Prettier. Update ESLint and TS configurations appropriately. ESLint and Prettier had a few new/changed rules that were resolved.

This update also made it so the Node Buffer class can no longer be used for base64 operations. TypeScript no longer replaces it during build as ES now has built in Uint8Array functions.

## New behavior

None.

## Code changes

- `src/helpers/MeasureBundleHelpers.ts` - Switched to using polyfilled Uint8Array and TextDecoder functions for base64 decode.
- `src/execution/ValueSetResolver.ts` - Switched to using polyfilled Uint8Array functions for base64 encode.

# Testing guidance

Run unit tests. Run regression test.

Build and use `npm link ../fqm-execution` in fqm-testify. Ensure built library still functions.


src/execution/ValueSetResolver.ts
```diff
@@ -19,7 +20,7 @@ export class ValueSetResolver {
   }
 
   private authHeaderValue(): string {
-    return Buffer.from(`apikey:${this.apiKey}`).toString('base64');
```
elsaperelli, line 22: Hmm I do not think this is working. Getting a 401 when I try to do a `npm run cli --valueSets`

test/unit/fixtures/elm/cql-translator.ts
```diff
@@ -1,5 +1,3 @@
-/* eslint-disable @typescript-eslint/ban-types */
-
 import fs from 'fs';
```
elsaperelli, line 1: I am not seeing an error, but do we need to update these like you did in `src/cli.ts`?

elsaperelli (changes requested):
Thinking are looking better! Two main things:

1. Looks like we need to include `test/` in the tsconfig.json or make a separate tsconfig.json for that directory.
2. I am not sure if the Buffer replacement is working correctly, at least for ValueSetResolver.ts, MeasureBundleHelpers.ts might be okay, I am getting 401 errors when I try to do `npm run cli --valuesets`.

test/unit/fixtures/elm/cql-translator.ts
hossenlopp: Eslint was complaining that this rule wasn't getting tripped in this file with this disable command in place. I needed to remove it to just resolve the warning. It like still is needed elsewhere.

src/execution/ValueSetResolver.ts
hossenlopp: Turns out I wasn't turning the string into the Uint8Array properly. A TextEncoder is needed. Resolved in 6dc1d32.

elsaperelli (approved):
Sweet!!! Working great!

eslint.config.mjs
```diff
@@ -0,0 +1,19 @@
+// @ts-check
+
+import js from '@eslint/js';
+import { defineConfig, globalIgnores } from 'eslint/config';
+import tseslint from 'typescript-eslint';
+import eslintConfigPrettier from "eslint-config-prettier/flat";
+
+export default defineConfig(
+  globalIgnores(["**/build", "**/connectathon", "**/coverage-percent-script"]),
```
lmd59:
```suggestion
  globalIgnores(["**/build", "**/connectathon", "**/coverage-percent-script", "**/regression/default-bundles", "**/regression/bundles"]),
```
Thoughts on adding this in? These folders sometimes end up with annoying errors for me.

lmd59 (commented):
Looks great! One other ignore that I would be interested in adding if it sounds good to you. Thanks for working through the headaches on this one!

eslint.config.mjs
hossenlopp: Good point! added in 70e1e20

lmd59 (approved):

PR #366, opened 2026-06-01
cql-execution 3.3.1 bump

elsaperelli:
# Summary
Upgrades to [cql-execution 3.3.1](https://github.com/cqframework/cql-execution/releases/tag/v3.3.1).

## New behavior
- There should be no new behavior (except for highlighting of boundless intervals).

## Code changes
- Upgrades cql-execution to 3.3.1
- Adds `npx` to `ts-node` line in unit test Makefile

# Testing guidance
- `npm run check`
- `npm run test:plus`
- Run regression tests
- Anything else you can think of :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ debug/
 .vscode
 **/.DS_Store
 test/unit/fixtures/elm/queries/output/*
+test/unit/fixtures/elm/output/*
 test/unit/fixtures/elm/queries/.start-translator
 coverage
 cache
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "commander": "^6.1.0",
         "core-js": "^3.49.0",
         "cql-exec-fhir": "^2.1.6",
-        "cql-execution": "^3.3.0",
+        "cql-execution": "^3.3.1",
         "fhir-spec-tools": "^0.4.0",
         "handlebars": "^4.7.7",
         "lodash": "^4.17.21",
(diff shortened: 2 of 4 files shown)


cmoesel (approved):
I'm not sure exactly what is required when reviewing a cql-execution bump PR, but I ran `npm run build:elm-fixtures` since you touched that make file and it seems to have worked fine. I also confirmed all the tests pass and the lockfile is not modified after install.

One thing I noticed, which is probably outside the scope of this PR, is that after running `npm run build:elm-fixtures`, there are a bunch of untracked JSON files in `test/unit/fixtures/elm/output/`. Should those be in the `.gitignore` file?

elsaperelli:
> I'm not sure exactly what is required when reviewing a cql-execution bump PR, but I ran `npm run build:elm-fixtures` since you touched that make file and it seems to have worked fine. I also confirmed all the tests pass and the lockfile is not modified after install.
> 
> One thing I noticed, which is probably outside the scope of this PR, is that after running `npm run build:elm-fixtures`, there are a bunch of untracked JSON files in `test/unit/fixtures/elm/output/`. Should those be in the `.gitignore` file?

Thank you so much for the review! That's perfect. Regarding the `.gitignore` comment, that's a good point and I will add that. The reason we didn't have that before is because we rarely touch that part of the code, but definitely good to add so thank you for the suggestion!

lmd59 (approved):
Looks good! 🎉

PR #372, opened 2026-07-29
Treat a measurement period end with no time as inclusive of the period it names

### LINKED ISSUE #371: Date-only measurementPeriodEnd (incl. the Measure.effectivePeriod.end default) is treated as start-of-day, excluding events on the last day of the measurement period
## Environment
- fqm-execution **1.8.5** (`Calculator.calculate`)
- Node 24, plain FHIR R4 patient bundles, `trustMetaProfile: false`

## Summary
A date-only `measurementPeriodEnd` (e.g. `'2026-12-31'`) appears to be parsed as **start-of-day** (`2026-12-31T00:00:00`), so clinical events occurring later on the final day of the measurement period are treated as outside the period. The README documents date-only as a valid format for this option ("Can be formatted as a date, date-time, or partial date"), and — more importantly — the option **defaults to `Measure.effectivePeriod.end`**, which in official MADiE-exported measure bundles is itself date-only (`2026-12-31`). So default-configuration executions of official content can silently exclude day-365 events.

## Reproduction (public content)
Using the official MADiE test content from [cqframework/dqm-content-qicore-2025](https://github.com/cqframework/dqm-content-qicore-2025) @ `ca4b4951`, measure **CMS125FHIRBreastCancerScreen** (v1.0.000), two published test cases exercise a denominator-exclusion Procedure performed at `2026-12-31T23:59:59Z` (the last second of the measurement period):

| Test case | Patient UUID | Expected (MADiE) | Actual w/ `measurementPeriodEnd: '2026-12-31'` | Actual w/ `'2026-12-31T23:59:59.999Z'` |
|---|---|---|---|---|
| BilateralMastProcOnDec31OfMP | `4cf81a94-81fb-4be2-b075-7d8f9ff02a6e` | IPP 1 / DEN 1 / **DENEX 1** / NUM 0 | IPP 1 / DEN 1 / **DENEX 0** / NUM 0 | matches expected |
| UniMastRandLProcDec31OfMP | `857fec09-9c8c-4e4b-a123-85f473b8fc2a` | IPP 1 / DEN 1 / **DENEX 1** / NUM 0 | IPP 1 / DEN 1 / **DENEX 0** / NUM 0 | matches expected |

With the date-only period end these were the only 2 mismatches out of 121 official test cases (CMS122 55/55 + CMS125 64/66); normalizing the end to `T23:59:59.999Z` gives 121/121 agreement with the published expected MeasureReports.

## Expected behavior
Per FHIR/CQL semantics for a day-precision boundary in this context, a date-only measurement-period **end** should be interpreted inclusively (end-of-day), matching how the reference content/tooling evaluates these cases. (Start-of-day is correct for the period **start**.)

## Workaround
Callers can normalize date-only ends to `T23:59:59.999Z` before calling the Calculator — but the `effectivePeriod.end` default path can't be worked around without overriding the option, which is easy to miss.

## Context
Found while validating a TypeScript measure-evaluation stack against the official MADiE test cases following the July 2026 CMS FHIR Connectathon (FHIR Quality Reporting with DEQM track). Happy to provide the full comparison harness output if useful.

Taleef7:
# Summary

Fixes #371.

A `measurementPeriodEnd` given without a time is parsed as midnight at the *start* of the period it names, so anything recorded later falls outside the interval. Nothing errors — the result is just quietly wrong.

FHIR R4 states the rule for the datatype this option defaults to, on `Period.end`:

> The high value includes any matching date/time. i.e. 2012-02-03T10:00:00 is in a period that has an end value of 2012-02-03.

I found this building a regression harness around the published MADiE decks for CMS122 and CMS125. With a date-only end it scored 119/121; with this change, 121/121. Both failures were CMS125 patients whose only qualifying procedure fell on the last day of the measurement period, which is part of why it is easy to miss.

`DEFAULT_MEASUREMENT_PERIOD_END` is itself `'2019-12-31'`, so callers relying on the default have been losing December 31st too.

# New behavior

The README documents all three FHIR `date` precisions as valid input here, and all three were affected, at increasing cost:

| Input | Resolved today | Resolved after | Lost today |
|---|---|---|---|
| `2019-12-31` | `2019-12-31T00:00:00.000` | `2019-12-31T23:59:59.999` | the last day |
| `2019-12` | `2019-12-01T00:00:00.000` | `2019-12-31T23:59:59.999` | the last 31 days |
| `2019` | `2019-01-01T00:00:00.000` | `2019-12-31T23:59:59.999` | the last 364 days |
| `2019-12-31T08:30:00Z` | `2019-12-31T08:30:00.000` | unchanged | — |

`measurementPeriodStart` is untouched. Start-of-period is already correct there.

# Two questions a reviewer will reasonably ask

**Why `.999` rather than something less arbitrary-looking?** Because it is not an approximation of the end of the day — it *is* the end of the day at the only precision the engine has. Millisecond is the finest field `cql-execution`'s `DateTime` carries, and `normalizeMillisecondsField` truncates anything finer *downward* (`.9999` → `999`), so nothing can land between `.999` and the next midnight. `successor(2019-12-31T23:59:59.999)` returns `2020-01-01T00:00:00.000`.

**Why not a half-open interval `[start, 2020-01-01)` instead?** It is equivalent, and it produces the same value anyway — `Interval.end()` calls `toClosed()`, which applies `predecessor()` to an open high bound and returns that same `.999`. I measured all four candidate designs against `cql-execution` 3.3.2:

| Design | event at 12-31 23:59:59 | at 12-31 23:59:59.9999 | at 2020-01-01 00:00 | `end of "Measurement Period"` |
|---|---|---|---|---|
| today, `[start, 12-31T00:00]` | false | false | false | `2019-12-31T00:00:00.000` |
| **this PR**, `[start, 12-31T23:59:59.999]` | true | true | false | `2019-12-31T23:59:59.999` |
| half-open, `[start, 2020-01-01)` | true | true | false | `2019-12-31T23:59:59.999` |
| day precision, `[start, 2019-12-31]` | **null** | **null** | **null** | `2019-12-31` |

The closed form is the one the pretty printer renders back to the user as the period they asked for, which is why I kept it. The fourth row is worth noting for anyone tempted by it: keeping day precision on the boundary is the most CQL-pure option and it returns null for everything, including a date obviously outside the period.

# Code changes

Two code paths carried the defect and both are fixed:

- **`Execution.getCQLIntervalEndpoints`** — the single place both option paths converge. By the time it runs, `Calculator` has already folded `Measure.effectivePeriod.end` into the options, so one change covers the caller-supplied and default paths alike.
- **`DataRequirementHelpers.createIntervalFromEndpoints`** — its end-only branch parses the end directly rather than through the above, so it had the same bug in a place the first fix does not reach. Its duration stays exactly one year: the start is derived from the resolved end, so both endpoints move together and the docblock's promise still holds.

The normalization itself is `parseTimeStringAsUTCConvertingToEndOfPrecision` in `ValueSetHelper`, resolving to the last millisecond of the year, the month, or the day according to how much of the date was given.

Four existing expectations encoded the old reading and are updated: three in `test/unit/DataRequirementHelpers.test.ts`, and one integration snapshot in `test/integration/pretty-various-types` whose printed interval now ends at `12/31/2022 11:59:59 PM`.

**This will move results for any measure bundle whose `effectivePeriod.end` carries no time, which includes the published eCQM content.** The regression script will report differences, and that is the fix working rather than a failure — worth a careful look before merge.

# Related, and deliberately not in this PR

While measuring the above I found that `parseTimeStringAsUTC` silently drops milliseconds and ignores a non-UTC offset, because `moment.defaultFormatUtc` has no fractional-second token and treats its trailing `Z` as a literal. Filed separately as **#376**: it affects the period *start* as well as the end and every other caller of that function, so it has a wider blast radius and wants its own review.

It leaves one wart while both are open, which I would rather name than leave to be discovered: after this PR a date-only end resolves to `.999`, while a caller passing `23:59:59.999Z` explicitly still resolves to `.000`. The default ends up more precise than the explicit value. Nothing here makes that worse — the explicit path was already lossy — but #376 is what closes it.

Also noticed: `parseTimeStringAsUTCConvertingToEndOfYear` has no call sites anywhere in the repository, and its `add(1,'years').subtract(1,'seconds')` idiom lands on `:59.000`, 999 ms short of the year it names. The new resolver's year branch supersedes it. I have left it alone rather than delete a public export in a bug-fix PR — say the word if you would like it removed.

# Testing guidance

`npm run check` passes (484 unit tests, lint, prettier) and `npm run test:integration` passes (46 tests).

`test/unit/Execution.test.ts` is new and covers the behavior directly: each of the three date precisions, short months, February in leap and non-leap years, a leap day, a datetime end left untouched, the default, and the start staying at start-of-period.

To verify end to end, calculate any measure against a patient whose only qualifying event falls on the last day of the period, passing `measurementPeriodEnd` as a bare date. Before this change the patient misses the population; after it, they are in it.

I also ran this against the published CMS122 and CMS125 MADiE decks through my own harness, swapping only the built library underneath it: stock 1.8.5 scores 119/121, this branch scores 121/121, with no other change on my side.


Taleef7:
Rebased onto current `master` and expanded, after going back over this against the running engine rather than the spec. The title and description are updated; summary of what changed since the version you may have glanced at:

**1. Partial dates were still broken.** The README documents this option as accepting "a date, date-time, or partial date", and my first cut only matched `YYYY-MM-DD`. So `2019-12` still resolved to December *1st* and `2019` to January 1st — losing 31 and 364 days respectively, rather than one. The resolver now handles all three precisions.

**2. A second code path had the same defect.** `DataRequirementHelpers.createIntervalFromEndpoints` routes through `getCQLIntervalEndpoints` only when *both* endpoints are given; its end-only branch parses the end directly and was still reading it as start-of-day. Fixed there too, with the one-year duration preserved by deriving the start from the resolved end.

**3. Filed #376 rather than fixing it here.** `parseTimeStringAsUTC` uses `moment.defaultFormatUtc`, which has no fractional-second token and treats its trailing `Z` as a literal — so it silently drops milliseconds *and* ignores a non-UTC offset (`08:30:00+02:00` is read as `08:30:00Z`, a two-hour error). That affects the period start as well as the end and every other caller of that function, so it seemed wrong to bundle into a date-only-end fix. It does mean that until #376 lands, a date-only end resolves to `.999` while an explicitly-passed `23:59:59.999Z` still resolves to `.000` — the default ends up more precise than the explicit value. Flagged in the description rather than left to be found.

**4. Wrote down why `.999`, since it looks arbitrary and is not.** Millisecond is the finest precision `cql-execution`'s `DateTime` carries, and it truncates anything finer downward, so `.999` is the last representable instant of the day: `successor(2019-12-31T23:59:59.999)` is `2020-01-01T00:00:00.000`. I also measured the half-open alternative — `[start, 2020-01-01)` behaves identically and `Interval.end()` returns the same `.999` from it, because `toClosed()` applies `predecessor()`. The comparison table is in the description. Keeping day precision on the boundary, which is the most CQL-pure option, returns null for everything including dates obviously outside the period.

One thing you may want to decide on: `parseTimeStringAsUTCConvertingToEndOfYear` has no call sites anywhere in the repo, and its `add(1,'years').subtract(1,'seconds')` lands 999 ms short of the year it names. The new resolver's year branch supersedes it. I have left it in place rather than delete a public export in a bug-fix PR — happy to remove it if you would prefer.

`npm run check` (484 unit, lint, prettier) and `npm run test:integration` (46) both pass. No CI has run on this PR yet — I think the workflows need maintainer approval for a first-time contributor, if that is something you can kick off.

PR #375, opened 2026-08-05
Normalize effectivePeriod start and end on measure

elsaperelli:
# Summary
This PR normalizes the measurementPeriod start and end dates so that the start date time is 00:00:00.000Z and the end date time is 23:59:59.999Z. 

## New behavior
- The start and end dates for a measurementPeriod are normalized in the `getCQLIntervalEndpoints` function.

## Code changes
- `src/execution/Execution.ts` - added the `normalizeMeasurementPeriodDate` function into this file and call it in `getCQLIntervalEndpoints`. Also makes sure the `DEFAULT_MEASUREMENT_PERIOD_START` and `DEFAULT_MEASUREMENT_PERIOD_END` are normalized if they are a simple Date. Same with the options.measurementPeriodStart and options.measurementPeriodEnd.
- `src/execution/ValueSetHelper.ts` - change that makes sure that the milliseconds get included
- `pretty-various-types.test.ts` - update to the integration tests
- `test/unit/DataRequirementHelpers.test.ts` - updates to 3 tests because they call `getCQLIntervalEndpoints` in `createIntervalFromEndpoints`
 
# Testing guidance
- `npm run test:plus`
- Run the regression scripts. They will fail but check that the failures make sense due to the time changes.
- `npm run cli -- detailed -m dqm-content-qicore-2025/bundles/measure/CMS125FHIRBreastCancerScreen/CMS125FHIRBreastCancerScreen-bundle.json -p ../../../patient-bundle.json --trust-meta-profile true --debug -o` (`patient-bundle.json` is attached on the Jira ticket for this task. You should see that the overall-clause-coverage is now 21.0% (compared to 20.6% on master)


test/unit/DataRequirementHelpers.test.ts
```diff
@@ -335,7 +335,7 @@ describe('DataRequirementHelpers', () => {
       const start = '2019-01-01';
       const end = '2020-01-01';
       const startCql = DateTime.fromJSDate(moment.utc(start, 'YYYYMDDHHmm').toDate(), 0);
```
lmd59, line 344: Maybe silly, but is it worth being more specific with the expectations for the startCql as well to future proof more? Like explicitly set it to the `T00:00:00.000Z` since the start is normalized too (even though that is what it defaults to at the moment)?

elsaperelli: I can if you really feel strongly about this, but `startCql` and `endCql` are both Dates that are just being used for comparison. 

This is what `startCql` looks like:
```
      DateTime {
        year: 2019,
        month: 1,
        day: 1,
        hour: 0,
        minute: 0,
        second: 0,
        millisecond: 0,
        timezoneOffset: 0
      }
```

And this is what `endCql` looks like:

```
      DateTime {
        year: 2020,
        month: 1,
        day: 1,
        hour: 23,
        minute: 59,
        second: 59,
        millisecond: 999,
        timezoneOffset: 0
      }
```

The input to the `extractDataRequirementsMeasurementPeriod` function is `options` and `mp`, so the `startCql` and `endCql` are just there for comparison and if I make the change that I think you are suggesting, `startCql` is the same.

lmd59: Yeah- I see what you're saying. Okay, fine to leave as is.

src/helpers/DataRequirementHelpers.ts
lmd59: I don't have a strong opinion either way, but did we still want to do this empty return if there's no measurement period info?

elsaperelli: If there is no measurement period info it is still calling `getCQLIntervalEndpoints` which has the defaults.

lmd59: Hmm- I'm not sure I understand how `getCQLIntervalEndpoints` gets called if we hit this branch of logic since it returns immediately. I think I'm missing something?

lmd59: I removed effectivePeriod from a test measure and ran dataRequirements and confirmed that it no longer creates dateFilters, so I believe that the defaults are not set for that case.

elsaperelli: Oh I see. Should I just remove lines 307-309?

lmd59: Yeah- I think that would align with our expectations for setting defaults in that case!

elsaperelli: Thanks Lauren! Just updated.

lmd59 (approved):
I think there's some streamlining we could do for measurement period in the future to reduce a bit of redundancy in the logic and centralize our options, effectivePeriod, defaults, and normalization.
For this issue though, I think this addresses it well without doing that larger refactor! Looks good to me! I did some spot checking in regression testing and that looks good to me as well!

hossenlopp (approved):
Agreed with @lmd59 that we should clean this up. But for fixing the issue, this is good for now. Checked in with regression and it was all on test cases that had data on 12/31 that was now being captured. No changes in how it executes with fqm-testify for last day of the year data.

